PR #52, opened 2020-12-02
Feature/notification

jouni:
Open for early review, so that you can see how I’ve chosen to implement these examples that need special attention. 

[Preview deployment](https://deploy-preview-52--vaadin-docs-preview.netlify.app/design-system/components/notification/)

Not all examples are implemented yet, but my plan is to create them similarly as these (except for the “positions” example, which won’t have the “card” preview).


frontend/demo/component/notification/notification-basic.ts
```diff
@@ -0,0 +1,51 @@
...
+  @internalProperty()
+  private notification: NotificationElement | null | undefined;
+
+  render() {
+    return html`
+      <vaadin-button @click="${this.open}">Try it</vaadin-button>
+
+      <!-- tag::snippet[] -->
+      <vaadin-notification
+        .renderer="${(root: HTMLElement) =>
```
tomivirkki, line 20: This will reassign the renderer on each render, which causes problems. Use one of the following approaches to avoid it:
- [Use a separate bound renderer](https://github.com/vaadin/docs/blob/fa4bb855289f71fc9462ad1d8e4fc6ae235125e1/frontend/demo/component/notification/notification-basic.ts)
- [Use an inline renderer with guard directive](https://github.com/vaadin/docs/blob/29b339f4d994e83c9cb17a7e20b5a2e64428d435/frontend/demo/component/notification/notification-basic.ts)

jouni: Which one should we prefer?

tomivirkki: Maybe the inline renderer is nicer as the content template is then part of the main template and not detached from the host element.

jouni: Why does it need to be a bound renderer? I don’t see a problem with the references with an unbound renderer, e.g. `.renderer="${this.renderer}"`, if I define it like this:

```typescript
renderer = (root: HTMLElement) =>
  render(...
```

tomivirkki: In the first example, without binding, the `this` in `this.notificationOpen = false` would have an incorrect reference IIRC.

jouni: The reason I wanted to inline renderer is that all the HTML would be visible in one snippet. That would prefer the use of a guard. But I’m not sure if that’s preferred otherwise?

tomivirkki: If `private renderer = (root: HTMLElement) =>` works without binding, then it's an option also 👍

jouni: Using `guard` (to be replaced with `vaadinRenderer` eventually) for examples where we want to show the full HTML in the snippet, and the other approach for cases where the snippet emphasis should be more on the other attributes/properties.

jouni:
I’m going to merge this PR without the Java examples, so that we get the current content visible at least.

PR #68, opened 2020-12-16
fix: add note on autogeneration

caalador:
Closes vaadin/flow#9598


articles/theming/application-theme.asciidoc
```diff
@@ -37,6 +37,9 @@ themes/ <1>
 <8> Generated configuration file
 <9> configuration for assets etc.
 
+[NOTE]
```
jouni, line 40:
```suggestion
.Create the theme folder manually
[NOTE]
```

jouni: All admonitions should have a title.

jouni (approved):

jouni:
We need to fix the writing style guide checks at some point. But not blocking on those now (as I expect we’ll need to rewrite this section later on anyways).

PR #94, opened 2021-01-07
feat: add doc for typescript auth helpers

haijian-vaadin:
This PR copies the [commit](https://github.com/vaadin/flow-and-components-documentation/commit/cee4d8d38b08bb87af53290d9a3a0a42baa7763e) done in the old doc repo for typescript authentication helpers.

For https://github.com/vaadin/flow-and-components-documentation/issues/1402.


jouni:
@haijian-vaadin, please update from master, to get the Vale check running correctly. Do a rebase preferably, if possible. Otherwise you need to touch each .asciidoc file to trigger the check for them.

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
```diff
@@ -0,0 +1,174 @@
...
+      this.onSuccess(result);
+    }
+
+    return result;
+  }
+
+  onAfterEnter(location: RouterLocation) {
+    this.returnUrl = location.redirectFrom || this.returnUrl;
+  }
+}
```
joheriks, line 93: Since this is pretty much self-contained, would it make sense to have here embedded code isntead of inline to ensure that it compiles now and in the future?

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
```diff
@@ -0,0 +1,174 @@
...
+----
+After the login view is defined, you should also define a route for the login view component, for example in the `index.ts` file. Note, the `path` for the login view component should be the same as what you defined in `SecurityConfig` for `http.formLogin().loginPage()`.
```
joheriks, line 95: "should be the same as what you defined in" could be reworded as "should match the one defined in".

May need a newline after this as the below snippet is not rendered correctly:
rndr

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
```diff
@@ -0,0 +1,174 @@
...
+title: Adding a Custom Login Form with Spring Security
+order: 91
+layout: page
+---
+
+ifdef::env-github[:outfilesuffix: .asciidoc]
+
+= Adding a Custom Login Form with Spring Security
+
+Instead of using the <<fusion-security-spring-login#, default Spring login form>>, you may want to use your own customized form. This article describes how to do that.
```
joheriks, line 11: One sentence per line.

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
```diff
@@ -0,0 +1,174 @@
...
+    component: 'main-layout',
+    children: [
+      ...
+    ]
+  }
+  ...
+}
+----
+The `isUserLoggedIn()` method in the above code examples uses
+a `loggedIn` variable stored in the https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage[localStorage] to check if the user is logged in. The `loggedIn` variable needs to be reset when logging out. Using localStorage allows navigating to sub views without having to check authentication from the backend on every navigation.
```
joheriks, line 156: One sentence per line.

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
```diff
@@ -0,0 +1,83 @@
+---
+title: Appendix: Handle session expiration in TypeScript
```
joheriks, line 2: Is it intended to be an appendix? Seems other sections don't have appendices.

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+layout: page
+---
+
+= Appendix: Handle session expiration in TypeScript
+
+Vaadin provides a built-in <<../advanced/fusion-advanced-client-middleware#, middleware>> `InvalidSessionMiddleWare` that you can use to detect session expiration, for example, to show a login view to the user.
+
+== How to use the InvalidSessionMiddleWare?
+
+The `InvalidSessionMiddleWare` requires a `OnInvalidSessionCallback` as a constructor parameter. The `OnInvalidSessionCallback` is a function that takes no parameters and should return a promise of `LoginResult`. `LoginResult` contains the metadata of a login result, including:
```
joheriks, line 13: One sentence per line.

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
```diff
@@ -0,0 +1,83 @@
...
+    return new Promise(resolve => {
+      this.onSuccess = (result: LoginResult) => {
+        this.remove();
+        resolve(result);
+      }
+      document.body.append(this);
+    });
+  }
+}
+----
```
joheriks, line 83: Could these two be embedded code instead of snippet?

joheriks (changes requested):

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
haijian-vaadin: good catch, done

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
haijian-vaadin: done

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
haijian-vaadin: done

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
haijian-vaadin: done, removed the Appendix prefix

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
haijian-vaadin: done

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
haijian-vaadin: good point moved them to separate src files

articles/fusion/security/fusion-security-custom-spring-login.asciidoc
joheriks: Seems there still two sentences per line in some places. Also the other Vale errors could be good to fix.

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
```diff
@@ -0,0 +1,32 @@
+---
+title: Handle session expiration in TypeScript
```
jouni, line 2:
```suggestion
title: Handling Session Expiration
```
No need to explicitly state “in TypeScript”, as we are writing documentation for Fusion.

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
```diff
@@ -0,0 +1,32 @@
...
+order: 141
+layout: page
+---
+
+= Handle session expiration in TypeScript
```
jouni, line 7:
```suggestion
= Handling Session Expiration
```

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
```diff
@@ -0,0 +1,32 @@
...
+
+Vaadin provides a built-in <<../advanced/fusion-advanced-client-middleware#, middleware>> `InvalidSessionMiddleWare` that you can use to detect session expiration, for example, to show a login view to the user.
```
jouni, line 9:
```suggestion
Use the built-in <<../advanced/fusion-advanced-client-middleware#, middleware>> `InvalidSessionMiddleWare` to detect session expiration, for example, to show a login view to the user.
```

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
```diff
@@ -0,0 +1,32 @@
...
+
+[source, typescript]
+----
+include::{root}/frontend/demo/fusion/authentication/handle-session-expiration/connect-client.ts[group=TypeScript]
+----
+
+[source, typescript]
+----
+include::{root}/frontend/demo/fusion/authentication/handle-session-expiration/login-view.ts[group=TypeScript]
+----
```
jouni, line 32:
```suggestion
.Example using `InvalidSessionMiddleWare`
[.example]
--
[source, typescript]
----
include::{root}/frontend/demo/fusion/authentication/handle-session-expiration/connect-client.ts[]
----

[source, typescript]
----
include::{root}/frontend/demo/fusion/authentication/handle-session-expiration/login-view.ts[]
----
--
```

frontend/demo/fusion/authentication/handle-session-expiration/login-view.ts
```diff
@@ -0,0 +1,53 @@
...
+  @property({ type: Boolean })
+  private error = false;
+
+  private returnUrl = '/';
+
+  private onSuccess = (_: LoginResult) => { Router.go(this.returnUrl) };
+
+  render() {
+    return html`
+      <vaadin-login-overlay opened .error=${this.error} @login="${this.login}">
```
jouni, line 16:
```suggestion
      <vaadin-login-overlay opened .error="${this.error}" @login="${this.login}">
```

Consistency with quotes.

frontend/demo/fusion/authentication/login-view.ts
```diff
@@ -0,0 +1,40 @@
...
+  private error = false;
+
+  // the url to redirect to after a successful login
+  private returnUrl = '/';
+
+  private onSuccess = (_: LoginResult) => { Router.go(this.returnUrl) };
+
+  render() {
+    return html`
+      <vaadin-login-overlay opened .error=${this.error} @login="${this.login}">
```
jouni, line 18:
```suggestion
      <vaadin-login-overlay opened .error="${this.error}" @login="${this.login}">
```

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
+---
+title: PWA offline authentication checks
```
jouni, line 2:
```suggestion
title: Offline Authentication Checks
```
IMO “PWA” is not something we should have in the article titles. It’s a technical umbrella term, and I feel it doesn’t add value.

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+order: 8
+layout: page
+---
+
+= Security aspects of Offline authentication checks
```
jouni, line 7:
```suggestion
= Security Aspects of Offline Authentication Checks
```

Not sure about this, but it feels the page title could start with the same terms as are used in the navigation title.

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+It is possible to do authentication checks offline. The article describes the security aspects of offline authentication checks that are good to take into account for PWAs.
```
jouni, line 9:
```suggestion
Take the following security aspects into account when you do authentication checks while the user is offline
```

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+== Authenticity
+Do you need to protect the local user's data on the device from whoever else may have access to the same device? If that is a requirement, then you will end up adding a `passcode` in addition to the regular online login (can be replaced with a fingerprint/face id auth through the WebAuthn API on the devices that support those). This is what many mobile bank apps do. 
+It is exactly to encrypt any user data that is cached locally for offline use. 
+If you do go that route, most likely you will end up having a helper library that you can call to check if the user is logged in or not.
+
+Arguably, not many apps would need to securely verify the user authenticity offline, and most can simply check if an online login happened in the not-so-distant past.
```
jouni, line 16:
```suggestion
Arguably, not many apps would need to securely verify the user authenticity offline, and most can simply check if an online login happened in the not-so-distant past.

```

Lists do not render as expected unless they are separated from the paragraph before it.

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+ - For that the simplest approach would be set a local storage flag after a user has successfully logged in online (and clear it when the user logs out). During offline navigation, the user auth then would be checked from the local storage. The <<../security/fusion-security-custom-spring-login#, Adding a Custom Login Form with Spring Security>> article describes how to do that. It is worth to classify that this as a UX improvement, not as a security feature, because it is easy to bypass through the browser dev tools.
+ - And if you need to add a more reliable expiration mechanism, you could use signed tokens (for example, JWT) so that the client can strongly verify the timestamp of the last login. Though, it is still not bulletproof because the user may change the date on the device.
```
jouni, line 18:
```suggestion
- For that the simplest approach would be set a local storage flag after a user has successfully logged in online (and clear it when the user logs out).
During offline navigation, the user auth then would be checked from the local storage.
The <<../security/fusion-security-custom-spring-login#, Adding a Custom Login Form with Spring Security>> article describes how to do that.
It is worth to classify that this as a UX improvement, not as a security feature, because it is easy to bypass through the browser dev tools.

- And if you need to add a more reliable expiration mechanism, you could use signed tokens (for example, JWT) so that the client can strongly verify the timestamp of the last login.
Though, it is still not bulletproof because the user may change the date on the device.
```

No need to indent the list dashes. Also, each sentence on their own line.

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+== Authorization
+When it comes to authorization (checking roles), in reality, it is only feasible on the backend. The reason is simple - the backend is the only place that can reliably guard the data the user is not authorized to view/modify. Once some data has reached the client, you have to assume it may be tampered with. So when it comes to, for example, checking whether or not a user has a role to access a certain view, the client-side implementation is mostly about the UX. You would need a way to check if the user has the permissions to navigate to a view, but this way does not need to be a particularly secure one. In case if the user tampers with the permission on their device, they may end up navigating to that view anyway, but they will not see any sensitive data because that would be securely guarded by the backend.
```
jouni, line 21:
```suggestion
When it comes to authorization (checking roles), in reality, it is only feasible on the backend.
The reason is simple – the backend is the only place that can reliably guard the data the user is not authorized to view/modify.
Once some data has reached the client, you have to assume it may be tampered with.
So when it comes to, for example, checking whether or not a user has a role to access a certain view, the client-side implementation is mostly about the UX.
You would need a way to check if the user has the permissions to navigate to a view, but this way does not need to be a particularly secure one.
In case if the user tampers with the permission on their device, they may end up navigating to that view anyway, but they will not see any sensitive data because that would be securely guarded by the backend.
```

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+So, what you may end up doing in case when you want to keep more than a plain `isLoggedIn` flag, is keeping a list of user roles in plain text in the local storage, and checking against that list in offline navigation.
```
jouni, line 23:
```suggestion
What you may end up doing in case when you want to keep more than a plain `isLoggedIn` flag, is keeping a list of user roles in plain text in the local storage, and checking against that list in offline navigation. 
```

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+So, what you may end up doing in case when you want to keep more than a plain `isLoggedIn` flag, is keeping a list of user roles in plain text in the local storage, and checking against that list in offline navigation. 
+If the user changes their roles through the dev tools, they could see the views they are not allowed to see (if the view templates are stored in the browser cache), but they will not see any data there.
+
+Using secure tokens, like JWT, makes the list of roles secure - it is cryptographically signed and the client app may verify that it has not been tampered with. While such tokens are extremely useful when accessing the backend because they allow the backend to be stateless, there are no particular benefits for client-side offline authentication.
```
jouni, line 26:
```suggestion
Using secure tokens, like JWT, makes the list of roles secure – it is cryptographically signed and the client app may verify that it has not been tampered with.
While such tokens are extremely useful when accessing the backend because they allow the backend to be stateless, there are no particular benefits for client-side offline authentication.
```

jouni (changes requested):
It pains me a little bit, that there are two `login-view.ts` files that have some amount of the same code.

I wonder if there’s a way to avoid the duplication? I didn’t look in detail if that would make sense, though – maybe the examples are different enough that the copy-pasted code doesn’t hurt.

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
haijian-vaadin: Good comment, done.
Regarding the 2 login-views. good comment, one focuses on the overlay part, renamed that one into login-overlay.ts, and removed some duplicated code.

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
haijian-vaadin: done

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
haijian-vaadin: done

articles/fusion/security/fusion-security-handle-session-expiration.asciidoc
haijian-vaadin: done

frontend/demo/fusion/authentication/handle-session-expiration/login-view.ts
haijian-vaadin: done

frontend/demo/fusion/authentication/login-view.ts
haijian-vaadin: done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: Good point, done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: done

articles/fusion/pwa/pwa-offline-authentication-check.asciidoc
haijian-vaadin: done

haijian-vaadin:
Oh, finally it's all green, no more Vale errors.

joheriks:
Got this TS error when running `mvn spring-boot:run`:
```
ERROR in [at-loader] ./frontend/demo/fusion/authentication/handle-session-expiration/login-overlay.ts:10:11 
    TS6133: 'onSuccess' is declared but its value is never read.
```

joheriks:
I noticed an error in the build:
```
error /Users/joheriks/vaadin/docs/articles/fusion/security/fusion-security-custom-spring-login.asciidoc, line 9 - level 0 sections can only be used when doctype is book
```
Seems to be because of the line just before:
```
ifdef::env-github[:outfilesuffix: .asciidoc]
```
What is this for, is it needed?

haijian-vaadin:
Confirmed with the doc team, it's not needed, I removed it.

joheriks:
> Confirmed with the doc team, it's not needed, I removed it.

There are some other Vale errors in that file now reported that could be fixed (Vale errors blink because they are cleared between checks and only reported for files touched in the last commit).

joheriks (approved):

jouni (approved):

.github/styles/Vocab/Docs/accept.txt
```diff
@@ -29,6 +29,7 @@ repo
 [sS]crollbars?
 [sS]croller
 [sS]ervlets?
+src
```
jouni, line 33: We should not introduce this change. Where is this causing problems? I believe all occurences of this “word” should either be inside code snippets or `[filename]##` classes, and therefore not part of the linting checks.

jouni (changes requested):

.github/styles/Vocab/Docs/accept.txt
haijian-vaadin: I don't remember, I think I got a Vale error. I can remove this and let's see.

haijian-vaadin: Yes, this one https://github.com/vaadin/docs/pull/94/files#annotation_853322011

jouni (changes requested):

.github/styles/Vocab/Docs/accept.txt
```diff
@@ -13,7 +14,9 @@ Gradle
 [gG]rayscale
 Gretty
 http
+JWT
 Karaf
+LDAP
```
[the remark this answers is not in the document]
haijian-vaadin: thanks for the comment, done

jouni (approved):

PR #95, opened 2021-01-07
Add description for themes hierarchy

tanbt:
Resolves https://github.com/vaadin/flow/issues/9700


articles/theming/overview.asciidoc
```diff
@@ -25,31 +25,37 @@ It is also important to understand <<style-scopes#,Style Scopes>> and <<css-cust
...
 It defines global font styles and text and background colors, among other shared visual aspects.
-+
+
 A theme defines all component styles but can also affect component behaviour.
 For example, the position of the date picker’s overlay on small viewports is determined by the theme.
-+
-Learn more about <<using-themes#,Using Themes>> and <<theme-variants#,Theme Variants>>.
 
-Theming::
+A theme is created in a separate folder under the `frontend/themes/`, `src/main/resources/static` or `src/main/resources/META-INF/resources/themes`.
```
caalador, line 36: I feel we should stay only on frontend/themes and not bring the reusable folders into the overview explanation.

articles/theming/overview.asciidoc
```diff
@@ -25,31 +25,37 @@ It is also important to understand <<style-scopes#,Style Scopes>> and <<css-cust
...
+The `[classname]#@Theme#` annotation is used to configure what theme to be applied in the application.
+
+Themes can be packaged as a JAR file and reused in multiple applications. A theme can also extend and overwrite an existing "parent theme".
```
caalador, line 39: Here we could mention that in this case the theme files should be located in either of the resources folders

articles/theming/overview.asciidoc
```diff
@@ -75,3 +82,31 @@ The guidelines are followed whenever possible.
...
+----
+no theme <1>
+├── built-in themes <2>
+│   ├── Lumo theme (default) <3>
+│   │   ├── @CssImport <4>
+│   │   ├── page.addStyleSheet <5>
+│   │   ├── @StyleSheet <6>
+│   │   └── (parent theme) ─ theme <7>
+│   └── Material theme (deprecated) <8>
+└── custom theme (server-side only)  <9>
```
caalador, line 99: I don't see why is this server-side only?

articles/theming/overview.asciidoc
```diff
@@ -75,3 +82,31 @@ The guidelines are followed whenever possible.
...
+----
+
+<1> Do not use any theme. See <<using-themes#no-theme,No Theme>>.
+<2> Use Lumo or Material built-in theme. See <<#built-in-themes, Built-in Themes>>.
+<3> Lumo theme is the default built-in theme which is applied without any specific configuration. See <<lumo/lumo-overview#,Lumo>>.
+<4> Import an external CSS file using `@CssImport`. See <<importing-style-sheets#, Importing Style Sheets>>.
+<5> Load external CSS file by `addStyleSheet` method. See <<../flow/importing-dependencies/tutorial-ways-of-importin#advanced-configuration-of-css-javascript-and-html-imports,Advanced configuration of CSS>>.
+<6> Import an external CSS file using `@StyleSheet`. See <<importing-style-sheets#, Importing Style Sheets>>.
+<7> A theme is the recommended way to style a Vaadin application. It can extend a parent theme. Themes are based on Lumo built-in theme. See <<application-theme#, Theme>>.
```
caalador, line 108: Why are these under lumo specifically? stylesheet and cssimport don't have to depend on lumo at all.
Parent theme also is only relevant for custom-theme and we shouldn't promote the old way of extending lumo or material

caalador (changes requested):

articles/theming/overview.asciidoc
tanbt: Fixed.

articles/theming/overview.asciidoc
tanbt: Added.

articles/theming/overview.asciidoc
tanbt: I saw https://vaadin.com/docs/v14/themes/creating-a-custom-theme.html mentions `This documentation applies only to server-side (Java) applications.`
Is it correct?

articles/theming/overview.asciidoc
tanbt: Fixed.

articles/theming/overview.asciidoc
caalador: V14 theme is quite different for what the Application theme is as it's available for use with fusion and can be used for client side theme things also.

caalador: Also I don't think the custom-theme should be mentioned anymore as it's not a feature we want to promote @pleku

jouni:
The Vale check is again running as expected. Please fix all errors and hopefully all warnings as well 🤗

articles/theming/overview.asciidoc
tanbt: This hierarchy is meant to structure and connect all the ways of styling an app.
That's why I mentioned custom-theme here, so users won't ask e.g. "why don't just use custom theme instead of application?".

Should custom-theme be marked as `deprecated` or `not recommended`?

caalador: @pleku Now that the new theme is called Custom Theme what should be done about this legacy custom theme? should it get the legacy label on this document?

articles/theming/overview.asciidoc
```diff
@@ -75,3 +82,31 @@ The guidelines are followed whenever possible.
...
+[source, filesystem]
+----
+no theme <1>
+├── built-in themes <2>
+│   ├── Lumo theme (default) <3>
+│   │   ├── @CssImport <4>
+│   │   ├── page.addStyleSheet <5>
+│   │   ├── @StyleSheet <6>
+│   │   └── (parent theme) ─ theme <7>
+│   └── Material theme (deprecated) <8>
```
jouni: I’ve seen this “deprecated” notice twice now. When was this decided? @rolfsmeds, can you elaborate?

articles/theming/overview.asciidoc
jouni: The whole “Theming” sections would benefit from a holistic rewrite. I’ve added that to my backlog, but I can’t promise when I would have the time to start drafting that.

articles/theming/overview.asciidoc
```diff
@@ -75,3 +82,32 @@ The guidelines are followed whenever possible.
...
+<3> Lumo theme is the default built-in theme which is applied without any specific configuration. See <<lumo/lumo-overview#,Lumo>>.
+<4> A theme is the recommended way to style a Vaadin application. A theme can extend another one. Themes are based on Lumo built-in theme. See <<application-theme#, Theme>>.
+<5> Material theme is a deprecated built-in theme. See <<material/material-overview#,Material>>.
+<6> Import an external CSS file using `@CssImport`. See <<importing-style-sheets#, Importing Style Sheets>>.
+<7> Load external CSS file by `addStyleSheet` method. See <<../flow/importing-dependencies/tutorial-ways-of-importin#advanced-configuration-of-css-javascript-and-html-imports,Advanced configuration of CSS>>.
+<8> Import an external CSS file using `@StyleSheet`. See <<importing-style-sheets#, Importing Style Sheets>>.
+<9> Use a custom server-side theme instead of a built-in theme. See <<creating-a-custom-theme#, Create a Custom Theme>>.
+
+NOTE: In case styles of an element are defined in multiple places (by `@CssImport` and `@StyleSheet` and a theme, for example) with the same selector, the styles will be applied following this <<css-loading-order#, CSS Loading Order>>.
```
jouni, line 113:
```suggestion
If multiple style sheets define styles for an element (for example, imported with `@CssImport`, `@StyleSheet`, and by a theme), the styles are applied in a specific order. See <<css-loading-order#, CSS Loading Order>> for details.
```

articles/theming/overview.asciidoc
pleku: There has been some back-and-worth with this, but to my understanding, the way things go now is:
1) Vaadin has two inbuilt themes for the components: Lumo and Material.
2) Lumo is the default theme you get when building apps with Vaadin
3) Starting from v19, we have a new _custom theme feature_ for Vaadin applications which improves theming experience by
- providing a unified way to do all theming related things
- being able to reuse and extend your custom themes in other applications
4) When you want to do any styling of your application, using a custom theme is the way to go.
5) Anything we used to have for theming before v19, like `@StyleSheet` or `@CssImport` does still work and is not deprecated to be removed but it is secondary to creating a custom theme. These are supported and documented, but not "highlighted" anymore due to 4).

If I've misinterpreted this, please correct me.

I think we've done a good effort overall, but when I look at the whole theming documentation chapter at the moment, I don't think it reflects the points 1-5 from above. This is not in particular anyone's fault, it is just what has evolved for number of reasons like: not having clear idea on what to call the new feature (now it is custom theme, used to be application theme), lacking design for the theming docs structure, combining flow, fusion and component theming in to one chapter, unclear ownership for the theming docs ... 

As we're now going out with the new custom theme feature, it is obvious that the theming documentation needs a larger overhaul. But it is an open question that who does it and when ? Should the flow team just get on with it ? @jouni @rolfsmeds

jouni: Thanks, Pekka, for the comprehensive overview of the situation. I agree with all that you said.

I also want to emphasize that this isn’t anyones fault, there’s not even anything to consider as a failure. Well, perhaps communication is the thing that always fails 😄, but this is a good effort to fix that.

I’’m mentally prepared to do this work. But since I’m also caught up in a much larger restructure design of the docs, I would appreciate any help.

rolfsmeds: Fwiw, I've also had _"Rewrite theming documentation with Jouni"_ on my backlog for some time already, so I'd say it's mostly a matter of scheduling some time for us to do that (preferrably before V19 GA), possibly with some help from the Flow team as needed.

tanbt: My big concern is that "custom theme" term is already used by https://vaadin.com/docs/themes/creating-a-custom-theme.html which will likely to appear on Google search.

Now "application theme" is also called "custom theme". Then how to deal with that old "custom theme"? Deprecate, remove, hide or redirect it or something?

pleku: When I look at that  https://vaadin.com/docs/themes/creating-a-custom-theme.html I feel that it is something that nobody should do anymore for any new project. So that leaves couple of options:
a) remove the outdated custom theme doc completely and rename application theme doc -> _Custom theme_  or _Creating a custom theme_
b) "deprecate" the outdated _creating a custom theme_ doc with notices in the doc saying that it does not apply after v19, linking to the new docs. Move it to be the last document page under theming, change the header to "Custom themes before v19" (later on Custom themes before Vaadin 14.5 & 19). The application theme doc would still be renamed to Custom Theme.

I would be fine with going with A) immediately as it is not needed even after reworking things - it is very rarely used feature and someone can probably find it from the old docs site if needed (?). WDYT @jouni @rolfsmeds 

I think would help a lot already now.

jouni: I would replace the current “Creating a Custom Theme” article with a new one (and have it at the same place in the nav as the current “Application Theme” article), that explains how to use the new feature. We are, after all, writing documentation for Vaadin 19, not for Vaadin 14. The previous version of the article is still available in the v14 docs – we don’t need to drag it with us in the latest docs.

tanbt: > replace the current “Creating a Custom Theme” article with a new one

Done. Please check.

jouni:
> Please fix all errors and hopefully all warnings as well

As a clarification: you don’t need to fix the warnings and errors for content that you haven’t modified. But avoid introducing more content that has warnings and/or errors.

articles/theming/overview.asciidoc
jouni: Works as expected 👍

I’m not reading the content with a super critical eye, as the expectation is that parts of it will be rewritten. I believe it serves it’s purpose decently until we have the time for that.

PR #99, opened 2021-01-11
Merge offline feature

haijian-vaadin:
Merge the docs for offline features to master.


jouni:
Gah, now Vale fails on some new error again. Something related to `custom-connection-indicator.ts`.

jouni:
The preview deployment fails on

```
12:51:49 PM: ERROR in [at-loader] ./frontend/demo/fusion/forms/contact-form.ts:12:31
12:51:49 PM:     TS2307: Cannot find module 'promise-file-reader' or its corresponding type declarations.
```

joheriks:
> The preview deployment fails on
> 
> ```
> 12:51:49 PM: ERROR in [at-loader] ./frontend/demo/fusion/forms/contact-form.ts:12:31
> 12:51:49 PM:     TS2307: Cannot find module 'promise-file-reader' or its corresponding type declarations.
> ```

We need to add `promise-file-reader` as a dependency (possibly worked previously due to being a transitive dependency of some other package). I assume it must be a runtime dependency (not dev) for this check to pass?

jouni:
> I assume it must be a runtime dependency (not dev) for this check to pass?

All files under `frontend` will be part of the docs-app build, so they need to compile.

haijian-vaadin:
My bad, I forgot the `promise-file-reader` dependency when resetting the `package.json` file. Now it's all green.

package.json
```diff
@@ -47,7 +47,8 @@
       "@vaadin/vaadin-material-styles": "1.3.2",
       "@vaadin/vaadin-rich-text-editor": "1.3.0",
       "lit-element": "2.3.1",
-      "@vaadin/vaadin-avatar": "1.0.2"
+      "@vaadin/vaadin-avatar": "1.0.2",
+      "promise-file-reader": "^1.0.3"
```
joheriks, line 51: Should be added to `/dependencies` rather than `vaadin/dependencies`?

joheriks (changes requested):

package.json
haijian-vaadin: done

articles/fusion/forms/appendix-form-images.asciidoc
```diff
@@ -0,0 +1,45 @@
+---
+title: Appendix: Image Fields in Client-Side Forms
```
joheriks, line 2: This probably should also have prefix "Appendix" in the sidebar, as it has been removed from other subsections that previously had it (e.g. "Appendix: Binding Data to Vaadin Components").

haijian-vaadin: done

joheriks (approved):

PR #102, opened 2021-01-12
feat: support application theme

tomivirkki:
This PR adds support for custom application themes in the docs app.

To try out the PR (with the Docbuilder app), you can use the built-in project theme or create a [reusable theme](https://vaadin.com/docs-beta/latest/theming/application-theme/#creating-a-reusable-application-theme). NOTE: Remember to run `mvn clean` whenever switching from a reusable theme back to the project theme.

- The built-in project theme (frontend/docs) can be modified while the app is running
  -  Changes to styles.css and document.css get applied on the fly
  -  Changes to component styles require a browser refresh
- Changes to a reusable theme (jar) require a full server restart

To enable a reusable theme:
- add the following to your pom.xml dependencies:
```xml
<dependency>
  <groupId>com.example</groupId>
  <artifactId>reusable</artifactId>
  <version>1.0-SNAPSHOT</version>
</dependency>
```
- Change the theme name in DemoExporter
`@Theme("reusable")`


frontend/demo/component/accordion/accordion-best-practices.ts
```diff
@@ -4,10 +4,18 @@ import { html, LitElement, customElement } from 'lit-element';
...
 
 // tag::snippet[]
 @customElement('accordion-best-practices')
 export class Example extends LitElement {
+  constructor() {
+    super();
+    // Apply custom application theme to the view.
+    // This is only supported if your app uses a custom theme.
+    applyTheme(this.shadowRoot);
+  }
```
tomivirkki, line 17: I kind of have mixed feelings about adding this snippet to every single TS example...

jouni: Yeah, it adds boilerplate. But what alternatives do we have? I can only think of adding a base class that every example extends from.

tomivirkki: Would be nicer if the logic could be wrapped as a decorator. Something like

```ts
@applyTheme
@customElement('checkbox-basic')
export class Example extends LitElement {
```

but apparently, this isn't doable

jouni (approved):
That was a lot of files changed 😅 
I didn’t take enough time to properly understand everything that’s going on in webpack.docbuilder.js, but I did try it out locally and tested that the theme works.

frontend/demo/component/accordion/accordion-basic.ts
```diff
@@ -4,10 +4,18 @@ import { html, LitElement, customElement } from 'lit-element';
...
 import '@vaadin/vaadin-ordered-layout/vaadin-vertical-layout';
 import '@vaadin/vaadin-text-field/vaadin-text-field';
+import { applyTheme } from 'themes/theme-generated.js';
 
 // tag::snippet[]
 @customElement('accordion-basic')
 export class Example extends LitElement {
+  constructor() {
+    super();
+    // Apply custom application theme to the view.
```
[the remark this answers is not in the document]
tomivirkki: Would be nice to only add 1 comment line instead of two. How would this sound?
`// Apply custom theme (only supported if your app uses one)`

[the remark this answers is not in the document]
jouni: Yeah, that’s nice and succinct 👍

jouni (approved):

PR #104, opened 2021-01-13
Deprecate CE install instructions for V19

pekam:
Porting https://github.com/vaadin/collaboration-engine-internal/pull/481


articles/ce/tutorial.asciidoc
```diff
@@ -18,12 +18,18 @@ and add the following collaboration features to a simple form:
...
 
 [[ce.tutorial.install]]
 === Install Collaboration Engine
 
-Collaboration Engine is a dependency that you add to your Vaadin project.
+Starting from Vaadin 19, Collaboration Engine is included in Vaadin and doesn't
+need any extra installation.
+
+If you are using Vaadin 14 (which is the default for a new project),
+you need to add Collaboration Engine as a dependency to your project.
```
jouni, line 32: I would like to change the content here so that it does not explicitly state any Vaadin versions (applies to the previous “Set Up a Vaadin Project” section as well).

Write it in the way that this feature has always worked with the current latest major version, and that it will always keep working with it. That said, we should annotate new features with the versioning badges: https://github.com/vaadin/docs/wiki/Authoring-articles#indicating-new-and-deprecated-features

If there are some additional things a user needs to do when combined with Vaadin 14, include those in the v14 branch of docs.

In this case, when CE is not part of Vaadin 18, which is still supported for a while, we could keep the installation instructions but mark them as deprecated since V19.

jouni (changes requested):

articles/ce/tutorial.asciidoc
pekam: Done. Thanks for the great advice!

jouni (approved):

pekam:
Normally I'd have waited for someone from the CE team to review as well, but after the latest update there's really nothing CE-specific in these changes.

PR #111, opened 2021-01-14
Migrate Listbox text content

alkali47:
Migrate text, remains example implementation


rolfsmeds (changes requested):
"Related Components" heading is missing.

jouni (approved):

PR #114, opened 2021-01-18
Fix: Update json snippets to read clearer

caalador:
Update all custom theme json snippets
to be complete json and add decorator text
for each item to make it clear that it's the
full json for shown feature.

Closes vaadin/flow#9827


caalador:
@jouni What should the string be for Vale to recognize it as not mutliple lines, tried with `.Sample [filename]#theme.json# for declaring assets`, but it still thinks there are multiple lines

jouni:
@caalador, yeah, sorry about that. We can ignore those errors for now, it’s fine, as we know it’s complaining for the wrong reason. I’m not sure if we can configure Vale better than this. I’ll try it locally at least. 

If we can’t make it work accurately, we need to make it more lax, to avoid false positives, even if that means more positives getting ignored. Should probably change this to a warning also, not an error.

jouni:
At least what should work is that you change from `[filename]#theme.json#` to `` `theme.json` ``.

jouni (approved):

taefi (approved):

PR #116, opened 2021-01-19
Fix show-hide Designer new-design dialog for different versions

tanbt:
The two dialogs for both V10+ and V15+ are shown, which is confused.

Only the dialog related to the current version of the docs should be shown.

image


jouni (changes requested):
I feel there is no need to use the `since` annotation. This documentation (the `master` branch) is for Vaadin 15+, and we can assume that we do not need to document changes related to Vaadin 14 and before.

So, author the documentation like it applies to Vaadin 15, and remove everything that no longer applies (i.e. only applies for Vaadin 14 or older). We have the `v14` branch for the documentation that applies for Vaadin 14 and before.

jouni (approved):
What’s up with the pnpmfile.js changes? It looks like nothing has changed, but it still shows up in the changes here in GitHub.

PR #119, opened 2021-01-19
fix: Fix OSGi docs as per PiT feedback

mshabarov:
Related-to: https://github.com/vaadin/osgi/issues/22

### DIFF
--- a/articles/flow/advanced/osgi-basic.asciidoc
+++ b/articles/flow/advanced/osgi-basic.asciidoc
@@ -33,7 +33,7 @@ The easiest way to convert regular Maven-based Vaadin application into a valid O
 ----
     <dependency>
         <groupId>com.vaadin</groupId>
-        <artifactId>vaadin-core</artifactId>
+        <artifactId>vaadin</artifactId>
         <scope>provided</scope>
     </dependency>
 ----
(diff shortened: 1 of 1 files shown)


jouni (approved):

jouni:
Does someone else need to review these changes?

mshabarov:
> Does someone else need to review these changes?
Yes, although the changes are quite simple, I added Denis as a reviewer just in case.

mshabarov:
This PR would most likely need more changes since the changes in the OSGi starter: https://github.com/vaadin/base-starter-flow-osgi/pull/108. Let me handle the comments there first and then decide, what clarification should be reflected here in this doc section.

articles/flow/advanced/osgi-basic.asciidoc
```diff
@@ -120,9 +122,16 @@ It is the developer's responsibility to register a `VaadinServlet` in the servle
 
 [source, Java]
 ----
-@Component(service = Servlet.class, , property =
-  HttpWhiteboardConstants.HTTP_WHITEBOARD_SERVLET_INIT_PARAM_PREFIX +
-  InitParameters.SERVLET_PARAMETER_COMPATIBILITY_MODE + "=false")
+import javax.servlet.Servlet;
```
denis-anisimov, line 125: I personally don't have a strong opinion but instead of having a number of imports I would use `@org.osgi.service.component.annotations.Component(....)`.

denis-anisimov: But this is minor.

denis-anisimov (approved):

articles/flow/advanced/osgi-basic.asciidoc
mshabarov: Yeah, this is also could be an option. I think, since there are couple of other OSGi annotations here, it's useful to show their packages in this code snippet as well.

PR #122, opened 2021-01-19
Test/vale

jouni:
# DO NOT MERGE

Testing Vale configuration.


jouni:
Looks like these are working nicely. I’ll push the workflow changes to master and v14 directly.

PR #123, opened 2021-01-19
Combine linting workflows

jouni:


jouni:
The workflows are running correctly when combined into one, which is expected. Alex linter doesn’t seem to work correctly, though, as it did not annotate the word "color" as expected. Need to check that separately.

PR #127, opened 2021-01-21
CE: Update tutorial to discourage PushMode.MANUAL usage

Peppe:


articles/ce/tutorial.asciidoc
```diff
@@ -61,6 +61,10 @@ public class MainView extends AppLayout {
 }
 ----
 
+Collaboration Engine works with any of the transport modes for push.
+The default push mode, `PushMode.AUTOMATIC` should be used.
+Collaboration Engine does not behave correctly if you explicitely configure push to use `PushMode.MANUAL`, as Collaboraiton Engine does not invoke `UI.push()`, and the user interactions would therefore not be pushed out to the other connected users.
```
pekam, line 66: Fixed couple of typos.
Removed one instance of "Collaboration Engine".
Shortened the end a bit.
```suggestion
If you explicitly configure push to use `PushMode.MANUAL`, Collaboration Engine does not work, as it doesn't manually invoke `UI.push()` to send the user interactions to the other connected users.
```

pekam, line 66: Maybe it would be better to wrap this in a NOTE box, because it's not required to continue the tutorial, but just extra information?

jouni: If you do, remember to add a title for the admonition (note), which gives the gist, so that readers can quickly skip it if they are not interested in reading it all.

Peppe: @jouni can you check if I did it right? I took model from the "Subscription required for production use" IMPORTANT notice a bit above, but even by only copy-pasting the admonition to this location, GitHub preview wasn't able to display the header correctly anymore.

articles/ce/tutorial.asciidoc
```diff
@@ -61,6 +61,12 @@ public class MainView extends AppLayout {
 }
 ----
 
+.The default push mode, `PushMode.AUTOMATIC` should be used.
```
jouni, line 64:
```suggestion
.Manual push mode is not supported
```
IMO this would be easier to understand. Also no punctuation (periods, question/exclamation marks) in headings and titles.

Peppe: Doesn't this turn the message from more positive to negative?

jouni: Yes. I didn’t think that would be a problem. We would explicitly state what the limitation is, not try to glean over it. But if you feel strongly that you want to put a positive spin on it, I would go with `.Default push mode is recommended`

jouni (approved):

pekam (approved):

PR #133, opened 2021-01-25
Update collaboration-avatar-group.asciidoc

Peppe:
Based on Olli's feedback from the PiT:
https://github.com/vaadin/collaboration-engine-internal/issues/484

The topic "avatars" doesn't make much sense here, because you are probably not collaborating on avatars. It could be something like "insuranceClaim" or "recipient data" or whatever.

This changes the topic from "avatars" to "insuranceClaims"
On the same time I also:
- Explained that "insuranceClaims" is an example, and can be anything
- Changed the freeform Note section to a [NOTE]
- Added a screenshot of CollaborationAvatarGroup to show what it looks like


articles/ce/collaboration-avatar-group.asciidoc
```diff
@@ -16,6 +16,10 @@ will be added to the group for all users. When the component becomes detached, o
 of navigating to another view or closing the browser tab, the user's avatar is
 removed from the group.
 
+
+.Example of a `CollaborationAvatarGroup`
+image::images/collaboration-avatar-group-example.png[Example of a `CollaborationAvatarGroup`]
```
jouni, line 21:
```suggestion
image::images/collaboration-avatar-group-example.png[Three avatars in a group, with three more avatars collapsed to an overflow item]
```
As general guidance, the alt text of an image should try to describe what the image contains. The original text is probably sufficient as well, but I wanted to try and improve it a bit, and not repeat the image title.

Side note: why are two avatars in the image using the same color (BS and the sheep)?

articles/ce/collaboration-avatar-group.asciidoc
```diff
@@ -26,11 +30,14 @@ User userEntity = userService.getCurrentUser();
 UserInfo userInfo = new UserInfo(userEntity.getId());
 
 CollaborationAvatarGroup avatarGroup = new CollaborationAvatarGroup(
-        userInfo, "avatars");
+        userInfo, "ensuranceClaims");
 add(avatarGroup);
 ----
 
-NOTE: `User` and `UserService` are examples of classes that might exist in your application for user management, and they are not available in Collaboration Engine.
+.`User`, `UserService` and "ensuranceClaims" are examples
```
jouni, line 37:
```suggestion
.`User`, `UserService` and `"ensuranceClaims"` are examples
```
I would keep this consistent, that all three of them are code snippets

articles/ce/collaboration-avatar-group.asciidoc
```diff
@@ -26,11 +30,14 @@ User userEntity = userService.getCurrentUser();
...
+[NOTE]
+`User` and `UserService` are examples of classes that might exist in your application for user management, and they are not available in Collaboration Engine.
+"ensuranceClaims" is the topic id, which can be any string to identify what the users are collaborating on.
```
jouni, line 40:
```suggestion
The `User` and `UserService` classes are not available in Collaboration Engine. They are examples of classes that might exist in your application for user management.
The topic ID `"ensuranceClaims"` can be any string to identify what the users are collaborating on.
```
I would try to avoid starting a sentence with either quotes or lowercase word/code snippet.

articles/ce/collaboration-avatar-group.asciidoc
Peppe: Not sure exactly how it should behave. In our demo, if you have color green, when editing one entity in the list, then you will have green after switching to another entity as well. If you switch over to another entity where someone already has the green color assigned, then there will now be two persons with green.

articles/ce/collaboration-avatar-group.asciidoc
Peppe: I updated the suggestion to add a line break on row 39, as it now contained two sentences.

jouni (approved):

PR #141, opened 2021-01-26
Add CE example source files and include them in code blocks

heruan:
Add CE example source files and include them in code blocks.

Closes https://github.com/vaadin/collaboration-engine-internal/issues/464


jouni:
Regarding the User entity: we already have a Person domain object in the docs. Could we reuse that instead of introducing another closely related one?

Then again, a “user” is not necessarily the same as a “person” in an application.

In any case, I think we should use the same package in the project, so the new User entity should also be located under `src/main/java/com/vaadin/demo/domain`.

jouni:
This is outside the scope of this PR, but as food for thought…

Eventually, I think we should build a proper demo application within the docs project (instead of these “dummy” classes), and then we can include the source code snippets from the application’s source code.

heruan:
> Regarding the User entity: we already have a Person domain object in the docs. Could we reuse that instead of introducing another closely related one?

I agree, they have a lot of similarities. For the scope of the PR, which is to move source files from the CE repo to this one, I'd just move the `User` entity in the right package as you're suggesting. A new PR might then merge the two entities and services (also the JSON files needs to be adapted).

> Eventually, I think we should build a proper demo application (or multiple) within the docs project (instead of these “dummy” classes), and then we can include the source code snippets from the application’s source code.

Yeah, that would be great!

heruan:
Vale linting fails with errors about many rewording/rephrasing matters, but this PR scope is just adding source files with code snippets and replace inline code blocks by including them.

articles/ce/topic-api.asciidoc
```diff
@@ -108,20 +85,11 @@ This is also done in the activation callback so that the subscription is opened
...
-        fieldValues.subscribe(event -> {
-            if ("isFriday".equals(event.getKey())) {
-                checkbox.setValue(
-                        Boolean.TRUE.equals(event.getValue(Boolean.class)));
-            }
-        });
-
-        return registration;
-    }
+include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[render,tags=open-topic;subscribe;!*,indent=0,group=Java]
```
pekam, line 90: These multi-include settings produce some weird outcomes, e.g.:
```java
CollaborationEngine.getInstance().openTopicConnection(this, "tutorial",
        localUser, topic -> {
            fieldValues.subscribe(event -> {
                if ("isFriday".equals(event.getKey())) {
                    checkbox.setValue(Boolean.TRUE
                            .equals(event.getValue(Boolean.class)));
                }
            });
        });
```
Having the entire `openTopicConnection` structure here gives the impression that this is the entire content of the callback, but `fieldValues` is not defined.

Maybe it would be better to remove the `openTopicConnection` wrapping from this and other snippets (except the first one which actually describes it)? Users can click the expand-button if they're unsure where the code is actually located.

src/main/java/com/vaadin/demo/ce/AvatarGroupDocumentation.java
```diff
@@ -0,0 +1,65 @@
...
+    private CollaborationAvatarGroup avatarGroup;
+    private UserService userService;
+
+    public AvatarGroupDocumentation() {
+        // tag::avatar-group-new[]
+        User userEntity = userService.getCurrentUser();
+        UserInfo userInfo = new UserInfo(userEntity.getId());
+
+        CollaborationAvatarGroup avatarGroup = new CollaborationAvatarGroup(
+                userInfo, "avatars");
```
pekam, line 27: This was changed by Peppe in the docs repo earlier, but got now overridden by bringing in these sources from the CE repo:
```suggestion
                userInfo, "ensuranceClaims");
```

articles/ce/collaboration-binder.asciidoc
```diff
@@ -136,16 +123,12 @@ When a `Converter` is used, you must provide the field's value type in `forField
...
-Checkbox married = new Checkbox();
-binder.forField(married, Boolean.class)
-        .withConverter(
-                fieldValue -> fieldValue ? MaritalStatus.MARRIED
-                        : MaritalStatus.SINGLE,
-                MaritalStatus.MARRIED::equals)
-        .bind("maritalStatus");
+include::{root}/src/main/java/com/vaadin/demo/ce/BinderDocumentation.java[render,tags=binder-converter,indent=0,group=Java]
 ----
+
```
pekam, line 131: You missed one :)

I can't add a comment there because it's outside the diff, but it's the last code snippet in this binder document.

src/main/java/com/vaadin/demo/ce/TopicView.java
```diff
@@ -0,0 +1,56 @@
...
+                                fieldValues.put("isFriday",
+                                        valueChangeEvent.getValue());
+                            });
+                    return registration;
+                    // end::registration[]
+                });
+        // end::open-topic[]
+    }
+}
+// end::topic-view[]
```
pekam, line 56: You could configure your IDE to ensure that there's a new line at the end of all files.

articles/ce/topic-api.asciidoc
heruan: True. What about having the example grow step-by-step? Seems even more clear to me to see how the different snippets build-up (committed this way, tell if you prefer otherwise).

pekam: Looks good to me! :+1:

pekam (approved):

pekam:
@heruan you need to resolve a conflict in this one

jouni:
> Vale linting fails with errors about many rewording/rephrasing matters

Yeah, it’s an issue with the vale-action at the moment, that it also checks lines that are not modified in the PR.

Let me know when the PR is ready, and I can merge it (admins are allowed to merge even if required checks are failing).

heruan:
> @heruan you need to resolve a conflict in this one

@pekam yeah, the conflict comes from my other PR for the correct location of `@Push` in V15+ (while the current status of this PR would be okay for V14).

heruan:
As discussed I've resolved the conflict with an inlined code block to avoid messing up with the existing Spring application class.

jouni:
Ready to merge?

heruan:
Yes @jouni, thanks!

PR #144, opened 2021-01-26
Vale: test `onlyAnnotateModifiedLines`

jouni:
Do not merge. Testing `onlyAnnotateModifiedLines: true`


jouni:
Test concluded, that it works as expected. The behavior might change in the future: https://github.com/errata-ai/vale-action/issues/32

PR #149, opened 2021-01-27
docs: add TS examples for Login

DiegoCardoso:


jouni:
Oh, man, it’s almost impossible to view the examples 😓 They seem to grab focus automatically, and the browser scrolls the example into the view. It starts happening after the first Modal Overlay example.

frontend/demo/component/login/login-modal-overlay-header-customisation-styles.ts
```diff
@@ -0,0 +1,26 @@
...
+      background-size: cover;
+      background-position: center;
+      background-image: url(${unsafeCSS(img)})
+    }
+
+    :host([theme='header-customised']) ::slotted(.title) {
+      color: var(--lumo-primary-contrast-color) !important;
+      font-weight: 600 !important;
+      margin-top: var(--lumo-space-m) !important;
+      margin-bottom: var(--lumo-space-xs) !important;
```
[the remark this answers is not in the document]
jouni: Yeah, that would be Antlers (the vaadin.com design system) stylesheet that’s styling raw HTML elements. It’s annoying.

@tomivirkki, would it be possible somehow to render the example-frame template without the docs site style sheets?

articles/ds/components/login/index.asciidoc
```diff
@@ -189,13 +212,13 @@ Login’s titles, descriptions, labels, and messages are all customizable using
...
-----
-include::{root}/src/main/java/com/vaadin/demo/component/login/LoginOverlayInternationalization.java[render,tags=snippet,indent=0,group=Java]
-----
+// [source, java]
+// ----
+// include::{root}/src/main/java/com/vaadin/demo/component/login/LoginOverlayInternationalization.java[render,frame,tags=snippet,indent=0,group=Java]
+// ----
 --
 
 === Header
```
jouni, line 224:
```suggestion
pass:[<!-- vale Vale.Terms = NO -->]

=== Header
```

Should then also add `pass:[<!-- vale Vale.Terms = YES -->]` after the table.

frontend/demo/component/login/login-modal-overlay-header-customisation-styles.ts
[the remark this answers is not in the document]
tomivirkki: I think so, any API suggestions? `[render, frame-unstyled, ...]`?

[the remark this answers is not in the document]
tomivirkki: Didn't find a clean way to do this :/ A conditional dynamic import of the global.css file would result in a layout shift.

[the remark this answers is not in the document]
jouni: It's possible to define different page templates for different routes/urls, right? Can those templates define their own stylesheets?

https://github.com/vaadin/docs-app/blob/master/gatsby-node.js#L328

articles/ds/components/login/index.asciidoc
DiegoCardoso: Done.

frontend/demo/component/login/login-modal-overlay-header-customisation-styles.ts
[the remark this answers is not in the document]
tomivirkki: Example frame already uses its own page template https://github.com/vaadin/docs-app/blob/master/gatsby-node.js#L359

Here's the related issue https://github.com/gatsbyjs/gatsby/issues/14333

frontend/demo/component/login/login-modal-overlay.ts
```diff
@@ -12,10 +13,25 @@ export class Example extends LitElement {
...
 
+  @internalProperty()
+  private loginOpened = false;
+
+  openLoginOverlay() {
+    this.loginOpened = true;
+  }
+  performLogin() {
+    this.loginOpened = false;
+  }
```
tomivirkki, line 24: nit: with functions this small, an inline function would be nicer: `@click=${this.loginOpened = true}` etc

tomivirkki (approved):

jouni:
What do we do with the focus issue? Try the page in Safari and you’ll see what I mean – it’s unusable IMO.

tomivirkki:
> What do we do with the focus issue? Try the page in Safari and you’ll see what I mean – it’s unusable IMO.

Oh. Is this because of the example visibility toggling?

DiegoCardoso:
Why is this happening?

jouni:
I tested if we could mitigate the issue using `include::[render, persistent]`, which keeps the rendered examples in the DOM even after they have gone outside the visible viewport. 

It helps only after all the examples have been scrolled through (during the initial scroll they still cause the page to jump), And for some reason, if I add it to all examples, some of them stop rendering completely 🤷

jouni:
> Is this because of the example visibility toggling?

DiegoCardoso:
The source of all evil was that as most of the `LoginOverlay` examples opened by default, an open change listener is triggered on `vaadin-overlay`, which at some point calls `_cycleTab(0,0)`, which in turn, focus the first element of the overlay. On Safari, this was causing the viewport jump to show the element.

To remedy this, we decided to try a similar solution present at the Notification example, which is to create a `LoginOverlay` mock-up and use that instead of the actual component.

articles/ds/components/login/index.asciidoc
```diff
@@ -27,6 +27,37 @@ In the meantime, you can view the previous examples for the web component and th
...
+  margin-top: 0;
+  border-top-left-radius: 0;
+  border-bottom-right-radius: 0;
+}
+.example.merge + .example .render {
+  text-align: center;
+  padding-top: 0;
+  padding-bottom: var(--docs-space-m);
+}
+</style>
```
jouni, line 58: We should at least consider extracting this into a shared include, as this is now being used in multiple component pages.

DiegoCardoso: How should that be done?

jouni: Copy the block, everything within and including the `++++` lines, and put it in a new asciidoc file. I suggest `articles/ds/components/_shared.asciidoc`. We could rename the `_shared-field-features.asciidoc`, and use tags to pick up the correct pieces from there.

Then use `include::../_shared.asciidoc[tag=merge-examples]` to pick up the style block into the Login (and other component) page.

DiegoCardoso: Done:

- rename `_shared-field-features.asciidoc` to `_shared.asciidoc`
- surround previous content with custom tag
- fix includes and add new custom tag
- add shared styles for merge example within a new tag
- refactor Login and Notification demo pages to include shared styles

frontend/demo/component/login/login-modal-overlay.ts
DiegoCardoso: Done.

jouni (approved):

PR #150, opened 2021-01-27
Frontend package installation with pnpm instructions

joheriks:
Part of vaadin/flow#9719.

PR #155, opened 2021-02-01
Frontend package installation with pnpm in V14

joheriks:
This is a backport of #150 to V14, with added explanation due to the V14.4 vs. V14.5 pnpm idiosyncrasies.


haijian-vaadin (approved):

haijian-vaadin (approved):

jouni:
I suggest you add `pass:[<!-- vale Vaadin.Versions = NO -->]` in the beginning of the document, to avoid the checks for explicit Vaadin version numbers (as I assume it’s impossible to avoid mentioning those here).

Also, if you want/can, we could copy the Vale checks from `latest` to here, so that they are aligned (I think there are some minor differences currently). That basically means copying the `.github/styles/Vaadin` and `.github/styles/Vocab` directories from `latest` to `v14`.

jouni:
I won’t be reviewing the content in more detail (don’t have the time). Let us know when it’s ready to merge, and either me or @magi42 can merge it.

jouni:
> Also, if you want/can, we could copy the Vale checks from latest to here

I did this just now, after all.

haijian-vaadin:
> I suggest you add pass:[] in the beginning of the document, to avoid the checks for explicit Vaadin version numbers (as I assume it’s impossible to avoid mentioning those here).
I added the `pass:[<!-- vale Vaadin.Versions = NO -->]` to the 2 documents, but seems it still gets the `Don't mention a specific version` error.

jouni:
@haijian-vaadin, pick up this commit from `latest`: https://github.com/vaadin/docs/commit/05aea9e71d2d6276003e2b6dd67d2f8ba784714f

That should allow `pass:[<!-- vale Vaadin.Versions = NO -->]` to work

joheriks:
> @haijian-vaadin, pick this commit from `latest`: [05aea9e](https://github.com/vaadin/docs/commit/05aea9e71d2d6276003e2b6dd67d2f8ba784714f)
> 
> That should allow `pass:[<!-- vale Vaadin.Versions = NO -->]` to work

I picked the commit, but it seems the `Vaadin.Versions` check still failed.

jouni:
Then there’s this one line which doesn’t trigger the error: https://github.com/vaadin/docs/pull/155/files#diff-3f45b4772c0eab28a0ee068018cf1bd50150dcfae08a0ac9ffd87b43339faf89R18

Strange indeed.

PR #158, opened 2021-02-02
feat: Add lit template documentation

caalador:
Add lit template documentation and
note in for polymer that lit is recommended

part of vaadin/flow#8862


articles/flow/templates/polymer-templates/styling-polymer-templates.asciidoc
```diff
@@ -6,6 +6,8 @@ layout: page
 
 = Styling Polymer Templates
 
+*Usage of Lit based templates is recommended even though polymer based templates are still be available in the next LTS.*
```
jouni, line 9:
```suggestion
.Polymer templates are deprecated
[CAUTION]
Polymer templates are still available, but <<../basic#,Lit templates>> are the recommended way of creating HTML based templates.
```

jouni (commented):
Suggestion for the notice about preferring Lit templates.

Would like to see many of the Vale errors and warnings fixed, at least for the new content (Lit templates). Polymer template docs are not as important, as they are deprecated.

articles/flow/templates/polymer-templates/styling-polymer-templates.asciidoc
caalador: Except this is for 14.5+ so Polymer templates are not deprecated for this version. That's why it's a recommendation only.

caalador:
As this is for 14 and the lit documentation is from master I feel they should be fixed in master and the fixes backported.

articles/flow/templates/polymer-templates/styling-polymer-templates.asciidoc
jouni: Right, I thought that might be the case. Then I wondered why they wouldn’t be deprecated, and thought I’d try and see the response 😄 

I’ll update the suggestion.

jouni:
> As this is for 14 and the lit documentation is from master I feel they should be fixed in master and the fixes backported.

Right, that’s true.

caalador:
What is the replacement for `../production/tutorial-production-mode-basic` as it was just removed and I don't know where the existing links should point to now.

pleku:
> What is the replacement for ../production/tutorial-production-mode-basic as it was just removed and I don't know where the existing links should point to now.

https://github.com/vaadin/docs/blob/latest/articles/guide/production/index.asciidoc

I can push it

jouni (approved):

PR #160, opened 2021-02-02
docs: add App Layout samples

DiegoCardoso:


frontend/demo/component/app-layout/app-layout-bottom-navbar.ts
```diff
@@ -16,40 +16,42 @@ export class Example extends LitElement {
...
+    return css`
+      h1 {
+        margin: 0 var(--lumo-space-s);
+        font-size: var(--lumo-font-size-l);
+      }
+
+      vaadin-tab {
+        flex: 1;
+      }
+
```
jouni, line 29:
```suggestion
      /* hidden-full-source-line: the bottom navbar is forced on in the example */
```

frontend/demo/component/app-layout/app-layout-navbar.ts
```diff
@@ -16,47 +16,38 @@ export class Example extends LitElement {
...
+  static get styles() {
+    return css`
+      h1 {
+        margin: 0 var(--lumo-space-s);
+        font-size: var(--lumo-font-size-l);
+      }
+
+      vaadin-tabs {
+        box-shadow: none;
+      }
```
jouni, line 28: Hmm, this is suspicious. Why is it needed? Isn’t `theme="minimal"` enough? Or perhaps that is too much, as it also removes the selected tab indicator?

DiegoCardoso: Yeah, this is with `theme="minimal":

Screenshot 2021-02-03 at 12 05 00

And this without removing `box-shadow`:
Screenshot 2021-02-03 at 12 04 38

At the mobile example, the box shadow can also be seen, although it looks ok since it takes the whole width of the page.

tomivirkki (approved):

PR #162, opened 2021-02-04
chore: add documentation for deploying project to Karaf

denis-anisimov:


articles/flow/advanced/osgi-karaf.asciidoc
```diff
@@ -0,0 +1,107 @@
+---
+title: Using Vaadin With Karaf
+layout: page
+---
+
+[[osgi.karaf]]
+= Using Vaadin With Karaf
+
+The Karaf `http` feature along with `war` feature provides HTTP Whiteboard implementation 
+which is de facto standard in Karaf.
```
jouni, line 10:
```suggestion
which is in practice the standard in Karaf.
```

jouni: Alternatively (I’m not sure which is more correct):

```suggestion
which is standard practice in Karaf.
```

pleku (approved):

pleku:
@jouni can you please merge this as the failed vale checks are not even part of this PR ?

jouni:
I'd like for @magi42 (or someone else) to first do a language check for the new content.

pleku:
> I'd like for @magi42 (or someone else) to first do a language check for the new content.

I would think that the content can be published (PR merged) and then the editor can do the changes there - no need to keep the original author waiting and the added value hostage here.

Legioth:
I think we should have the same policy for reviewing documentation as for reviewing code.

It's not like someone with ownership for implementation would be eager to merge a PR with potential omissions either. The main reason is the same: merging incomplete contributions leads to situations where it's the maintainer's responsibility to ensure any necessary follow-up will be taken care of.

PR #165, opened 2021-02-05
fix: use a range for lit-html version to avoid duplicates

tomivirkki:
Vaadin generates a package.json file that depends on a fixed version of lit-html (1.2.1). Starting up the project runs `npm install` and results in 2 different lit-html versions installed.

Screenshot 2021-02-05 at 10 26 15

The conflicting versions cause problems in examples that use lit-html to render content. Currently, they use `render` imported from `lit-html` and `html` tag imported from `lit-element` which uses a different version of `lit-html` internally.

```js
import { render } from 'lit-html';
import { html, ... } from 'lit-element';

...

render(html`<div>Financial report generated</div>`, root);
```

One solution would be to also import `html` (with a different name) from `lit-html`:

```js
import { render, html as htmlTag } from 'lit-html';
import { html } from 'lit-element';
```

...but that gets messy when `html` is used for the root level (and possibly interpolated content) and `htmlTag` for `render`'ed content.

This PR changes the explicit `lit-html` dependency from `1.2.1` to `^1.2.1` to ensure that only one version of `lit-html` is installed.

Screenshot 2021-02-05 at 11 42 01

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
       "workbox-precaching": "5.1.4",
       "webpack-manifest-plugin": "2.2.0"
     },
-    "hash": "73eb1061976ee848763dd2beb832f412be0fc3ae27032efafcc827ad83e65f96"
+    "hash": "c2c87cfcbb2e59c02637afc3e76b0708116455970390548c8677976a6d40d606"
   },
   "dependencies": {
     "@polymer/iron-icon": "3.0.1",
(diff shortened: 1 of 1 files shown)


tomivirkki:
No longer relevant since https://github.com/vaadin/flow/commit/4df7c9c4028c98dfb062953b52d600b83c6f16bc

PR #167, opened 2021-02-05
Fix Vale errors in CE docs

pekam:
Note: this fixes only errors, not the warnings.

Fix https://github.com/vaadin/collaboration-engine-internal/issues/511

### DIFF
--- a/articles/ce/overview.asciidoc
+++ b/articles/ce/overview.asciidoc
@@ -61,7 +61,7 @@ There are two high-level APIs that enable the specific use case of editing a for
 
 `CollaborationBinder` enhances the regular Vaadin `Binder` to share the current value of each field with all other users configured to use the same topic.
 
-There is also a `CollaborationAvatarGroup` component that shows the avatar of each user who is present in a given topic, for example, all the users currently looking at or editing the same item.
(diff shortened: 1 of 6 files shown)


jouni (approved):

articles/ce/overview.asciidoc
```diff
@@ -84,7 +84,7 @@ Collaboration Engine is production-ready and stable, however the following featu
 * Missing support for complex data structures with nested arrays and maps.
 * Topic data is not persisted between server restarts.
   Applications can manually persist topic data and repopulate after a restart if necessary.
-* There is no support for collaboration between multiple nodes or pods in a cluster of application servers.
+* Collaboration between multiple nodes or pods in a cluster of application servers is not supported.
   A future version will enable running Collaboration Engine as a standalone server or a cloud deployment that can be used from multiple application servers.
-* There is only support for Vaadin views implemented with Flow.
+* Collaboration Engine can be used only in Vaadin views implemented with Flow.
```
Peppe, line 89:
```suggestion
* Collaboration Engine can only be used in Vaadin views implemented with Flow.
```
The order of words seems better this way.

Peppe (changes requested):

articles/ce/overview.asciidoc
pekam: Yeah, would've been slightly better, but I was too hasty to merge.

PR #171, opened 2021-02-08
Add Dialog documentation and TS examples

DiegoCardoso:


jouni:
Error:

```
5:00:39 PM: ERROR in [at-loader] ./frontend/demo/component/dialog/dialog-resizable.ts:28:11
5:00:39 PM:     TS2564: Property 'people' has no initializer and is not definitely assigned in the constructor.
```

Btw, in case you are unaware, you can check the build logs yourself by clicking the "Details" link for the deploy/netlify check.

DiegoCardoso:
> Btw, in case you are unaware, you can check the build logs yourself by clicking the "Details" link for the deploy/netlify check.

Yeah, I am aware of it. It's just that it fails silently and I forgot to return to check it. 😅

jouni:
I think we should combine the `draggable` and `drag-handles` examples together. The difference is so minor, I think it would be nice to get the information in the same example. Just add a code comment for the line that adds `class="draggable"`.

Another change I would make is to enable dragging for the resizable example. A dialog that you can resize but not reposition is a really odd use case :)

DiegoCardoso:
Changed to use only one example for draggable dialog and made the resizable dialog draggable as well.

jouni:
Looks like the grid inside the resizable dialog isn’t shrinking below a certain height.

Screen Shot 2021-02-18 at 16 50 57

DiegoCardoso:
> Looks like the grid inside the resizable dialog isn’t shrinking below a certain height (you get double scrollbars).

Now it should work. I had added a `min-height` to the Grid, but it's not needed and it's causing the double scrollbars effect.

articles/ds/components/dialog/index.asciidoc
```diff
@@ -38,3 +23,150 @@ The usage examples for this component are currently in progress. In the meantime
 - https://vaadin.com/components/vaadin-dialog/html-examples[Web Component Examples]
 - https://vaadin.com/components/vaadin-dialog/java-examples[Java Component Examples]
 ====
+
+// Custom style additions for the examples, that combine two renders together
+++++
+<style>
+.example.merge,
```
tomivirkki, line 30: Same as https://github.com/vaadin/docs/pull/149#discussion_r577803795 Should make this a shared include instead of repeating

DiegoCardoso: The shared styles is included on #149 PR. After it gets merged, I'll update this one to use it as well.

DiegoCardoso: Refactor to use the shared styles.

frontend/demo/component/dialog/dialog-draggable.ts
```diff
@@ -0,0 +1,72 @@
...
+        .renderer="${guard([], () => (root: HTMLElement) => {
+          render(
+            html`
+              <vaadin-vertical-layout
+                theme="spacing"
+                style="width: 300px; max-width: 100%; align-items: stretch;"
+              >
+                <vaadin-horizontal-layout
+                  class="draggable"
+                  style="border-bottom: 1px solid var(--lumo-contrast-20pct); cursor: move; padding: var(--lumo-space-m) var(--lumo-space-l); margin: calc(var(--lumo-space-s) * -1) calc(var(--lumo-space-l) * -1) 0"
```
tomivirkki, line 44: Remember to replace the negative margin hack with `theme="no-padding"` as soon as we have it available in V20 alpha

tomivirkki (approved):

PR #173, opened 2021-02-09
feat: add custom ds init script

tomivirkki:
Relates to https://github.com/vaadin/docs-app/issues/175

### DIFF
--- a/scripts/docbuilder-init.js
+++ b/scripts/docbuilder-init.js
@@ -3,7 +3,6 @@ const fs = require('fs');
 const path = require('path');
 
 const execShellCommand = cmd => {
-  console.log(cmd);
   return new Promise((resolve, _) => {
     execSync(cmd, (_, stdout) => resolve(stdout));
   });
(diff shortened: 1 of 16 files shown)


scripts/docbuilder-init.js
```diff
@@ -0,0 +1,63 @@
...
+
+  for (const match of matches) {
+    dsFoundationSharedFileContent = dsFoundationSharedFileContent.replace(
+      match[0],
+      `link:http://vaadin.com/docs-beta/latest/theming/${match[1]}[${match[2]}]`
+    );
+  }
+  execShellCommand(
+    `echo '${dsFoundationSharedFileContent}' > articles/ds/foundation/_shared.asciidoc`
+  );
```
jouni, line 58: This is likely to be unnecessary soonish. The Theming section is going to be rewritten and split under Flow, Fusion, and Design System sections.

jouni: I suppose we can keep it until then?

tomivirkki: Sure, just need to remember to remove it afterward.

jouni (approved):

PR #175, opened 2021-02-10
CE: add config auto-detection and setDataDir

heruan:
Update `ce-next` with latest changes and depend on CE snapshot.

Closes https://github.com/vaadin/collaboration-engine-internal/issues/512

### DIFF
--- a/articles/ce/going-to-production.asciidoc
+++ b/articles/ce/going-to-production.asciidoc
@@ -53,20 +53,42 @@ You can set the path as a system property in a `VaadinServiceInitListener`, whic
 
 ==== Spring Applications
 
-If you have a Spring application, you can register a `VaadinServiceInitListener` by implementing the interface and annotating the class as a `@SpringComponent`.
-In the `serviceInit` method, you can then set `vaadin.ce.dataDir` as a system property.
+If you have a Spring application, you can provide a bean of type `CollaborationEngineConfiguration` from your main Spring class by annotating a method with the `@Bean` annotation and returning a configuration instance with the data directory path set using the `setDataDir` method.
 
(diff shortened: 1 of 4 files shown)


articles/ce/going-to-production.asciidoc
```diff
@@ -53,20 +53,42 @@ You can set the path as a system property in a `VaadinServiceInitListener`, whic
...
-include::{root}/src/main/java/com/vaadin/demo/ce/MyVaadinInitListener.java[tags=init-listener;!*,indent=0]
+@SpringBootApplication
+public class Application {
+
+    public static void main(String... args) {
+        SpringApplication.run(Application.class, args);
+    }
+
+    @Bean
+    public CollaborationEngineConfiguration ceConfigBean() {
```
pekam, line 69: Can't we `include` this snippet?

If it's because of the `@SpringBootApplication`, there are two options:
- Just having a snippet for the `@Bean` method and mention in the text that this can be e.g. in the `Application` class.
- Keep the annotation inline, but `include` the rest.

heruan: I've included just the `@Bean` method snippet, it's already mentioned that it could be in the main Spring class.

pekam (approved):
Approved. I didn't check if everything renders correctly, because there's no preview. But I think we should anyway double-check these changes once we include them to `latest`/`v14`.

Feel free to squash and merge, but use some more meaningful commit message than "Update ce-next branch", e.g "CE: add config auto-detection and setDataDir" or something like that.

heruan:
> I didn't check if everything renders correctly, because there's no preview.

I checked the render locally, the snippets are fine although they're different from the doc site.

> Feel free to squash and merge, but use some more meaningful commit message

👍

PR #177, opened 2021-02-11
Remove the --shamefully-hoist requirement for V19

joheriks:
V19 generates the .npmrc file with this flag.


Haprog (approved):

haijian-vaadin (approved):

jouni:
I would have preferred to avoid the admonition (or at least have a title for it), and to avoid the explicit Vaadin version number mention. But those are very minor issues in the end.

PR #179, opened 2021-02-11
Added download link

AtteNymalm:


samie (approved):

jouni:
Unfortunately GitHub actions are not run on pull requests made from repo forks. So the the Vale check is going to stay pending. This is going to be an issue for external contributors, so we probably need to relax the requirement eventually. We’ve kept if required for now, so that we can catch the most commonly occurring issues and tweak the checks accordingly.

I can merge this.

PR #180, opened 2021-02-11
fix: stress that endpoint methods should be public

vlukashov:
This is based on direct user feedback. Ppl do not immediately realise that non-public endpoint methods are not exposed to TypeScript.

Related: https://github.com/vaadin/flow/issues/7681


jouni:
One more Vale error on line 154: https://github.com/vaadin/docs/pull/180/files#diff-e35d6932940f0313ca12b5b5698563fb103ccbcb3733f33182b700c0cffbf425R154

Then you can have a green check 😃

jouni (approved):
In case you want to reduce some of the “one sentence per line” warnings, use monospace formatting for the filenames, for example, `` .`City.ts` ``

PR #182, opened 2021-02-12
feat: add docs on getting the browser url to Java

pleku:
Part of vaadin/flow#1897


articles/flow/browser/accessing-data-from-the-browser.asciidoc
```diff
@@ -0,0 +1,23 @@
+---
+title: Accessing Browser Details
+order: 3
+layout: page
+---
+
+# Accessing Browser Details
```
jouni, line 7:
```suggestion
= Accessing Browser Details
```

articles/flow/browser/accessing-data-from-the-browser.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+
+Sometimes you need to access certain details from the browser, like: screen dimensions, device pixel ratio.
+The `Page` object provides API to interact with the browser and retrieve data to the server side.
+
+## Getting the Window Location URL
```
jouni, line 12:
```suggestion
== Getting the Window Location URL
```

articles/flow/browser/accessing-data-from-the-browser.asciidoc
```diff
@@ -0,0 +1,23 @@
+---
+title: Accessing Browser Details
+order: 3
+layout: page
+---
+
+# Accessing Browser Details
+
+Sometimes you need to access certain details from the browser, like: screen dimensions, device pixel ratio.
+The `Page` object provides API to interact with the browser and retrieve data to the server side.
```
jouni, line 10:
```suggestion
The `Page` object provides API to interact with the browser and retrieve data, for example, screen dimensions or device pixel ratio, to the server side. 
```

articles/flow/browser/accessing-data-from-the-browser.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+
+## Getting the Window Location URL
+To fetch the user browser current URL, you can use the method `Page.fetchCurrentURL`.
```
jouni, line 13:
```suggestion
Use the [methodname]`Page.fetchCurrentURL` method to fetch the current URL from the browser.
```

articles/flow/browser/accessing-data-from-the-browser.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+
+For example:
+
+.Getting the browser URL
+[source,java]
+----
+include::src/main/java/com/vaadin/flow/browser/BrowserDetail.java[tags=get-current-url, indent=0, group=Java]
```
jouni, line 20:
```suggestion
include::src/main/java/com/vaadin/flow/browser/BrowserDetail.java[tags=get-current-url, indent=0]
```
Only use `group` if you have the same example implemented for both Flow and Fusion.

articles/flow/browser/src/main/java/com/vaadin/flow/browser/BrowserDetail.java
```diff
@@ -0,0 +1,37 @@
+/*
```
jouni, line 1: This is a source file from Flow, right (judging from the package name)? If so, then it should not be added to the docs project. Write the source code inline inside Asciidoc, as this is an exceptional case.

If this is a class that could be in a user project, then put it under `src` where you’d put it in a real project. All source files should always go either under `frontend` or `source`, never under `articles`.

jouni (changes requested):

articles/flow/browser/src/main/java/com/vaadin/flow/browser/BrowserDetail.java
caalador: This is just for the document. It's the same as was set into the theming tutorial which we based the file position and style on.

jouni: The “Theming” section contains “legacy” structure, with a nested `src` folder it it. I’m currently working on restructuring the whole section, and removing the source files under it.

jouni: The same should be done for `/articles/flow/src´ (delete it completely). I haven’t yet removed it because I haven’t checked if some of those files are used in `include::[]` directives somewhere in the Flow docs. If they aren’t, we could remove the `src` folder immediately.

So, I suggest we remove this file, and either create an example under `/src` (in the project root), or copy the snippet inside Asciidoc. The latter will add some amount of debt, which should be addressed later, so the former approach is recommended.

pleku: Done

jouni (approved):

PR #188, opened 2021-02-16
Add crud TS examples

yuriy-fix:


jouni:
Something strange is happening when I browse the page on an iPhone. The editor dialog keeps opening automatically during scrolling. Probably related to the example being attached/detached when it enters/exits the viewport.

jouni:
Also, the grid columns are rendered somehow broken:

image

jouni:
Not seeing that kind of problems at https://cdn.vaadin.com/vaadin-crud/1.3.0/demo/#crud-basic-demos

frontend/demo/component/crud/crud-open-editor.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  async firstUpdated() {
+    this.items = await getPeople();
+  }
+
+  render() {
+    return html`
+      <!-- tag::snippet[] -->
+      <vaadin-crud
+        edit-on-click
```
tomivirkki, line 29: The example description says `you can have it open using double click:`, but this opens it with a single click

tomivirkki (changes requested):

tomivirkki:
> The editor dialog keeps opening automatically during scrolling

This happens because at least one of the examples has the editor pre-opened. On mobile, the editor occupies the full screen.

frontend/demo/component/crud/crud-open-editor.ts
yuriy-fix: Have updated that one to use double click. Got confused as `edit-on-click` is also removing the edit column, but is opening on single click 😅

yuriy-fix:
crud-example
Have tried it out, it seems to be stuck in that position until scrolling is completed (not always). Probably something related to the scrolling in the docs app. Did it go to normal after some time for you, @jouni?

jouni:
Yeah, it's not persistent all the time.

yuriy-fix:
Editor is not visible by default anymore, it removes extra hacks to set edited items, combo-box values and trick with editor positioning.

frontend/demo/component/crud/crud-toolbar.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  async firstUpdated() {
+    this.items = await getPeople();
+  }
+
+  render() {
+    return html`
+      <!-- tag::snippet[] -->
+      <vaadin-crud include="firstName, lastName" .items=${this.items}>
+        <div slot="toolbar" style="flex: 1;">Total: <b>${this.items.length}</b> employees</div>
```
tomivirkki, line 30: I'd imagine the count here is expected to change if you delete / add items.

frontend/demo/component/crud/crud-toolbar.ts
```diff
@@ -0,0 +1,39 @@
...
+        <a slot="toolbar" new-button href="javascript:"
```
tomivirkki, line 31: Why is an anchor used here (and not a tertiary button)?

frontend/demo/component/crud/crud-open-editor.ts
```diff
@@ -0,0 +1,50 @@
...
+      <!-- end::snippet[] -->
+    `;
+  }
+
+  onDblClick(e: MouseEvent) {
+    const item = (this.grid?.getEventContext(e) as GridEventContext).item;
+    if (item) {
+      this.grid?.dispatchEvent(new CustomEvent('edit', { detail: { item: item } }));
+    }
+  }
```
tomivirkki, line 49: This is too hacky. Crud has `editedItem` property, that you should use instead
```suggestion
  onDblClick(e: MouseEvent) {
    this.editedItem = ((e.target as GridElement).getEventContext(e) as GridEventContext).item;
  }
```

frontend/demo/component/crud/crud-editor-actions.ts
```diff
@@ -0,0 +1,43 @@
...
+  private items: Person[] = [];
+
+  async firstUpdated() {
+    this.items = await getPeople();
+    // tag::snippet[]
+    const deleteBtn = this.shadowRoot
+      ?.querySelector('vaadin-crud')
+      ?.shadowRoot?.querySelector('vaadin-dialog-layout')
+      ?.querySelectorAll('vaadin-button')[2];
+    deleteBtn?.setAttribute('theme', 'primary error');
```
tomivirkki, line 28: This can't be the way we want to promote styling the editor buttons. How would you do this in Java?

tomivirkki (changes requested):

frontend/demo/component/crud/crud-editor-actions.ts
jouni: Oh, wow, yeah, this looks terrible 😆 Wasn’t there a way to slot in new buttons or something like that? Or maybe I remember that from Confirm Dialog. 

If there’s no nicer way of doing this, I think we should not customize the buttons then.

frontend/demo/component/crud/crud-toolbar.ts
yuriy-fix: Updated the logic to reflect the current number of items.
Using `size` property in order not to update the `innerHTML` as need to set **bold**, still not sure if it's the best way 🤔

frontend/demo/component/crud/crud-toolbar.ts
yuriy-fix: Was following existing example: https://vaadin.com/components/vaadin-crud/html-examples/crud-custom-demos#toolbar.
Have updated it, using `vaadin-button` now.

frontend/demo/component/crud/crud-editor-actions.ts
yuriy-fix: Yeah, that's quite difficult to achieve. I have added `TODO commit` before: https://github.com/vaadin/docs/pull/188/commits/33afbfe6d54f33cc29b8a2e0eaaeb435a5660201.

There is [relevant issue](https://github.com/vaadin/vaadin-crud/issues/228) created some time ago, but I am not sure if we can implement that one or just skip this example. 

It looks like important stuff to show, probably can convert it to cookbook example instead.

frontend/demo/component/crud/crud-open-editor.ts
yuriy-fix: There is a problem with editing the same item. Unfortunately `editor-opened-changed` event cannot be used in this case either as it's fired before `editedItem` is changed on closing the editor.

Have created corresponding issue and attached a PR right away: https://github.com/vaadin/vaadin-crud/issues/241.
This is an enhancement, other option is to make sure `editedItem` is changed before emitting `editor-opened-changed`. Not sure which one is preferable.

tomivirkki: It's a shame that `editedItem` doesn't notify. Making it notify could almost be considered a bugfix as it's more like a missing event.

The following logic should work until we have the proper notify event available:
```js
@editor-opened-changed=${(e: CustomEvent) => {
  if (e.detail.value) {
    // The editor was opened
    this.editedItem = (e.currentTarget as CrudElement<Person>).editedItem;
  } else {
    this.editedItem = null;
  }
}}
```

or a shorter version:
```js
const editedItem = (e.currentTarget as CrudElement<Person>).editedItem;
this.editedItem = e.detail.value ? editedItem : null;
```

frontend/demo/component/crud/crud-toolbar.ts
tomivirkki: If you want to avoid adding another property to the class you could bind the length of the already existing `items` property instead:
`<b>${this.items.length}</b>`
and then request an update on size change:
`@size-changed=${() => this.requestUpdate()}`

frontend/demo/component/crud/crud-toolbar.ts
tomivirkki: 👍  Why is the `style="margin: 0;"` override needed?

frontend/demo/component/crud/crud-open-editor.ts
yuriy-fix: The `editedItem` notifies now, example updated.

frontend/demo/component/crud/crud-toolbar.ts
yuriy-fix: It is getting default indent from button styles, it's not that big, so not overriding anymore.

frontend/demo/component/crud/crud-toolbar.ts
yuriy-fix: Good point! Updated.

frontend/demo/component/crud/crud-open-editor.ts
tomivirkki: Seems to have an error

Screenshot 2021-04-12 at 16 10 18

frontend/demo/component/crud/crud-open-editor.ts
```diff
@@ -0,0 +1,55 @@
...
+
+import { html, LitElement, customElement, internalProperty } from 'lit-element';
+import '@vaadin/vaadin-crud/vaadin-crud';
+import { GridElement, GridEventContext } from '@vaadin/vaadin-grid';
+import { CrudEditedItemChanged } from '@vaadin/vaadin-crud';
+import { getPeople } from 'Frontend/demo/domain/DataService';
+import Person from 'Frontend/generated/com/vaadin/demo/domain/Person';
+import { applyTheme } from 'Frontend/generated/theme';
+
+@customElement('crud-open-editor')
```
tomivirkki, line 12: There's something odd about this example. Seems that you can only edit once. After that, none of the items can be edited:

Kapture 2021-04-12 at 16 20 26

frontend/demo/component/crud/crud-open-editor.ts
yuriy-fix: Fixed the error.

frontend/demo/component/crud/crud-open-editor.ts
yuriy-fix: Good catch! The issue is created. Can we consider it as a non-blocker?
Will be fixed automatically be fixed in the example with bumping the version after the issue is fixed in the component.

tomivirkki: Yup, if it's a component bug and can't be worked around in this example then it's not a blocker

frontend/demo/component/crud/crud-open-editor.ts
```diff
@@ -0,0 +1,55 @@
...
+    return html`
+      <!-- tag::snippet[] -->
+      <vaadin-crud
+        include="firstName, lastName, email, profession"
+        .items=${this.items}
+        .editedItem=${this.editedItem as any}
+        @edited-item-changed=${(e: CrudEditedItemChanged<Person>) =>
+          (this.editedItem = e.detail.value)}
+      >
+        <vaadin-grid slot="grid" @dblclick="${this.onDblClick}">
```
vursen, line 40: Is it possible to configure `eslint` or `prettier` so that it will format quotes around an expression in some consistent way?

Seems that it could look a bit better in terms of documentation.

tomivirkki (approved):

frontend/demo/component/crud/crud-open-editor.ts
jouni: Yeah, would be great to have a lint rule for this. This was recently discussed in Slack: https://vaadin.slack.com/archives/C0168FAKA8H/p1617025824119800

PR #190, opened 2021-02-16
Update overview.asciidoc

kaviwe:
@Peppe


Peppe (changes requested):

jouni:
The preview build failed during Maven dependency download. We can ignore that, these changes only affect the text, so the build should be unaffected.

Peppe (approved):

PR #206, opened 2021-02-19
docs: add Grid Pro text and TS examples

DiegoCardoso:


src/main/resources/data/people.json
```diff
@@ -1,3602 +1,4002 @@
...
+    "email": "rdarthe1@netvibes.com",
     "id": "2",
-    "subscriber": true,
-    "membership": "Regular",
+    "birthday": "1977-07-12",
+    "subscriber": false,
+    "membership": "Premium",
     "pictureUrl": "",
-    "profession": "Oncologist"
+    "profession": "Software Engineer III"
```
tomivirkki, line 40: The existing data shouldn't be changed this much since it affects existing examples also.

Wouldn't feel safe picking a "Software Engineer" as my doctor 😅  (from ComboBox examples)

Screenshot 2021-02-22 at 10 41 35

tomivirkki (changes requested):

frontend/demo/component/gridpro/grid-pro-styling-editable-cells.ts
```diff
@@ -0,0 +1,40 @@
...
+  private items: Person[] = [];
+
+  async firstUpdated() {
+    this.items = await getPeople();
+  }
+
+  render() {
+    return html`
+      <!-- tag::snippet[] -->
+      <vaadin-grid-pro class="editableCustomEffect" .items=${this.items}>
```
tomivirkki, line 31: By convention, we don't use camel case for class names. Use "editable-custom-effect" instad.

tomivirkki (changes requested):

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -9,21 +9,22 @@ section-nav: incomplete, commercial
 ---
 
 = Grid Pro
+:experimental:
```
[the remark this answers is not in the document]
DiegoCardoso: It is needed due to the use of the keyboard macro.

frontend/demo/component/gridpro/grid-pro-styling-editable-cells.ts
DiegoCardoso: Changed it to kebab-case.

src/main/resources/data/people.json
DiegoCardoso: Added only the `birthday` property to the existing people dataset instead.

tomivirkki (approved):

frontend/demo/component/gridpro/grid-pro-editors.ts
```diff
@@ -0,0 +1,64 @@
...
+          }}
+          .editModeRenderer=${(
+            root: HTMLElement,
+            _column?: GridColumnElement,
+            model?: GridItemModel
+          ) => {
+            root.innerHTML = '';
+            const datePicker = document.createElement('vaadin-date-picker');
+            datePicker.value = (model?.item as Person).birthday;
+            root.appendChild(datePicker);
```
tomivirkki, line 57: Doesn't `lit-html`'s `render` work here?

frontend/demo/component/gridpro/grid-pro-edit-column.ts
```diff
@@ -0,0 +1,49 @@
...
+  render() {
+    return html`
+      <!-- tag::snippet[] -->
+      <vaadin-grid-pro .items=${this.items} enter-next-row>
+        <vaadin-grid-column
+          header="Name (read-only)"
+          .renderer=${(root: HTMLElement, _column?: GridColumnElement, model?: GridItemModel) => {
+            const item = model?.item as Person;
+            root.textContent = `${item.firstName} ${item.lastName}`;
+          }}
```
tomivirkki, line 39: The column and model parameters shouldn't need to be optional. The optional `model`'s `item` is cast to `Person` and accessed later, so optional doesn't work here.
```suggestion
          .renderer=${(root: HTMLElement, _column: GridColumnElement, model: GridItemModel) => {
            const { firstName, lastName } = model.item as Person;
            render(
              html`
                ${firstName} ${lastName}
              `,
              root
            );
          }}
```

The renderers in `grid-pro-editors.ts` have the same issue.

tomivirkki: Well, interestingly `GridBodyRenderer` defines them as optional...not sure why.

frontend/demo/component/gridpro/grid-pro-editors.ts
DiegoCardoso: I tried using it, but the second render won't work. My understanding is that the `render` helper adds two comments markers before and after the element to be attached to mark its position. But, when the cell leaves the editing mode, it removes all the content to place the plain text. The next time the cell enters edit mode, `render` is called, but the comment markers won't be attached anymore and it will not attach the field.

I guess it might've work if GridPro also uses `render` internally for the same purpose.

tomivirkki: Reported a grid pro issue here: https://github.com/vaadin/vaadin-grid-pro/issues/163

frontend/demo/component/gridpro/grid-pro-edit-column.ts
tomivirkki: This is still an issue:
```js
const item = model?.item as Person; // Casting a possible undefined as a Person
item.firstName; // Accessing a property of the possibly undefined object
```

articles/ds/components/grid-pro/index.asciidoc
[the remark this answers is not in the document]
jouni: Yeah. We could set that attribute in docs-app already, but right now it’s needed in each article separately.

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -35,7 +36,14 @@ Grid Pro is a commercial extension of the free Vaadin Grid component. Grid Pro s
 // include::../../../../src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags=snippet]
 // ----
 
-// --
+--
+
+.Features shared with Grid
+[NOTE]
+====
+Grid Pro is an extension of the <<../grid#,Grid>> component and thus all of Grid’s features
```
jouni, line 44:
```suggestion
Grid Pro is an extension of the <<../grid#,Grid>> component and all Grid’s features
```

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -46,3 +54,248 @@ The usage examples for this component are currently in progress. In the meantime
...
+// include::../../../../src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags=snippet]
+// ----
+
+--
+
+=== Enter Next Row
+
+Pressing kbd:[Enter] and kbd:[Shift+Enter] saves the changes and exists edit mode by default.
+
+kbd:[Enter] and kbd:[Shift+Enter] can be made to additionally focus the editable cell in the next and previous row, respectively, by using the Enter Next Row mode.
```
jouni, line 127:
```suggestion
kbd:[Enter] and kbd:[Shift+Enter] can be made to focus the editable cell in the next and previous row, respectively, by using the Enter Next Row mode.
```

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -46,3 +54,248 @@ The usage examples for this component are currently in progress. In the meantime
...
+// [source, java]
+// ----
+// include::../../../../src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags=snippet]
+// ----
+
+--
+
+=== Recommended Built-in Editors
+
+Grid Pro features three recommended built-in editors: Text Field, Checkbox and Select.
```
jouni, line 175:
```suggestion
Grid Pro features three recommended built-in editors: Text Field, Checkbox, and Select.
```

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -46,3 +54,248 @@ The usage examples for this component are currently in progress. In the meantime
...
+// ----
+// include::../../../../src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags=snippet]
+// ----
+
+--
+
+== Styling Editable Cells
+
+Unless all columns are editable, apply styling to visually distinguish between editable and non-editable cells using the `editable-cell` part name selector.
+Additionally or alternatively, you can suffix read-only columns with “(read-only)” and/or editable columns with “(editable)”.
```
jouni, line 240:
```suggestion
For example, you can suffix read-only columns with “(read-only)” and/or editable columns with “(editable)”.
```

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -46,3 +54,248 @@ The usage examples for this component are currently in progress. In the meantime
...
+// [source, java]
+// ----
+// include::../../../../src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags=snippet]
+// ----
+
+--
+
+== Best Practises
+
+=== Inline vs Non-inline Editing
```
jouni, line 269:
```suggestion
=== Inline vs Non-Inline Editing
```

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -46,3 +54,248 @@ The usage examples for this component are currently in progress. In the meantime
...
+
+Inline editing is recommended when:
+
+- the user typically needs to make a lot of small changes to different items
+- quick editing is important
+
+Non-inline editing is preferable
+
+- there’s a lot of columns/fields
+- users typically need to edit only one particular item at a time
```
jouni, line 279:
```suggestion
- users typically need to edit only one item at a time
```

articles/ds/components/grid-pro/index.asciidoc
```diff
@@ -46,3 +54,248 @@ The usage examples for this component are currently in progress. In the meantime
...
+- adding new items is common (as you might want to have edit and create modes work the same way, and creating new items with inline editing is not recommended with Grid Pro).
+- any of the editors need to be bigger than a simple field, such as a Text Area or multiselect field of any kind
```
jouni, line 281:
```suggestion
- any of the editors need to be bigger than a simple field, such as a Text Area or multi-select field of any kind
```

frontend/demo/component/gridpro/grid-pro-edit-column.ts
DiegoCardoso: Added checks for undefined.

PR #212, opened 2021-02-22
Add context-menu TS examples and docs

yuriy-fix:


frontend/demo/component/contextmenu/context-menu-checkable.ts
```diff
@@ -0,0 +1,43 @@
...
+    { text: 'Tamaki Ryushi', checkable: true }
+  ];
+
+  @internalProperty()
+  private selectedItem?: { text: string } = this.items[1];
+  // end::snippet[]
+
+  render() {
+    return html`
+      <!-- tag::snippethtml[] -->
```
DiegoCardoso, line 32: I feel that separating the snippet like this makes the code shows a bit confusing:
image

frontend/demo/component/contextmenu/context-menu-presentation.ts
```diff
@@ -0,0 +1,139 @@
...
+        { component: 'hr' },
+        { component: this.createItem('vaadin:trash', 'Delete') }
+      ];
+    }
+  }
+  // end::snippet[]
+
+  render() {
+    return html`
+      <!-- tag::snippethtml[] -->
```
DiegoCardoso, line 63: Same issue as reported on the other comment.

frontend/demo/component/contextmenu/context-menu-presentation.ts
```diff
@@ -0,0 +1,139 @@
...
+    const itemsArray = this.createItemsArray(this.gridItems.slice(0, 6));
+
+    const contextMenu = this.shadowRoot?.querySelector('vaadin-context-menu');
+    if (contextMenu) {
+      contextMenu.items = [
+        { component: this.createItem('vaadin:file-search', 'Open') },
+        {
+          component: this.createItem('vaadin:user-check', 'Assign'),
+          children: [
+            { component: this.createSearchField() },
```
DiegoCardoso, line 39: The search field behavior feels a bit odd. Once you start typing, focus jumps to the item that matches the value (the Context Menu's default behavior).

DiegoCardoso (changes requested):

frontend/demo/component/contextmenu/context-menu-checkable.ts
jouni: One option is to use two separate include::[] lines for the tags, and add ... between them. Here’s an example: 

```asciidoc
[source, html]
----
include::{root}/frontend/demo/component/combobox/combo-box-presentation.ts[render,tag=combobox,indent=0,group=TypeScript]

...

include::{root}/frontend/demo/component/combobox/combo-box-presentation.ts[render,tag=renderer,indent=0,group=TypeScript]
----
```

Result: https://vaadin.com/docs-beta/latest/ds/components/combo-box/#custom-item-presentation

frontend/demo/component/contextmenu/context-menu-presentation.ts
yuriy-fix: Yeah, I **have removed the search field**. Firstly, tried to workaround it with stopping the propagation of the event, but it turns out to be more complex on mobile, so it's definitely not a part of best practices.

frontend/demo/component/contextmenu/context-menu-keyboard-shortcuts.ts
```diff
@@ -0,0 +1,76 @@
...
+        </vaadin-grid>
+      </vaadin-context-menu>
+      <!-- end::snippethtml[] -->
+    `;
+  }
+
+  activeItemChanged(e: CustomEvent) {
+    // Used for demonstration purposes.
+    const item = e.detail.value;
+    this.grid && (this.grid.selectedItems = item ? [item] : []);
```
tomivirkki, line 62: Instead of storing the grid element as a property and assigning its `selectedItems` programmatically, make the `selectedItems` an internal property and data-bind it to the grid in the template (just like you do with `gridItems`).

frontend/demo/component/contextmenu/context-menu-keyboard-shortcuts.ts
```diff
@@ -0,0 +1,76 @@
...
+  onKeyDown(e: KeyboardEvent) {
+    if ((e.key === 'Space' || e.key === ' ' || e.key === 'Enter') && e.target) {
+      const { x, y } = (e.composedPath()[0] as HTMLElement).getBoundingClientRect();
+      e.target.dispatchEvent(
+        new CustomEvent('open-context-menu', { detail: { x, y, e }, bubbles: true })
+      );
+
+      e.preventDefault();
+      e.stopPropagation();
+    }
```
tomivirkki, line 74: This goes really deep into the component internal details. Too much so to be in the official product documentation.

tomivirkki (changes requested):

frontend/demo/component/contextmenu/context-menu-dividers.ts
```diff
@@ -0,0 +1,56 @@
...
+@customElement('context-menu-basic')
+export class Example extends LitElement {
+  constructor() {
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot);
+  }
+
+  // tag::snippet[]
+  @internalProperty()
```
tomivirkki, line 22: Only use `@internalProperty` where it's meaningful

frontend/demo/component/contextmenu/context-menu-checkable.ts
```diff
@@ -0,0 +1,43 @@
...
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot);
+  }
+
+  // tag::snippet[]
+  @internalProperty()
+  private items = [
+    { text: 'Abigail Lewis', checkable: true },
+    { text: 'Allison Torres', checkable: true, checked: true },
```
tomivirkki, line 20: The `checked` property never changes here. `checkable` is not part of context-menu API as the documentation suggests. It was used in the demo logic of an [old example](https://cdn.vaadin.com/vaadin-context-menu/4.5.0/demo/#context-menu-items-demos) but you'd need to add similar app logic to have it do anything.

articles/ds/components/context-menu/index.asciidoc
```diff
@@ -11,30 +11,241 @@ section-nav: incomplete
...
+include::{root}/frontend/demo/component/contextmenu/context-menu-hierarchical.ts[render,tag=snippethtml,indent=0,group=TypeScript]
+----
+
+// [source, java]
+// ----
+// include::../../../../src/main/java/com/vaadin/demo/component/contextmenu/ContextMenuHierarchical.java[render,tags=snippet,indent=0,group=Java]
+// ----
+--
+
+=== Custom Items
```
tomivirkki, line 114:
```suggestion
== Custom Items
```

frontend/demo/component/contextmenu/context-menu-presentation.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  @internalProperty()
+  private gridItems: Person[] = [];
+
+  // tag::snippet[]
+  async firstUpdated() {
+    this.gridItems = await getPeople();
+    const itemsArray = this.createItemsArray(this.gridItems.slice(0, 6));
+
+    const contextMenu = this.shadowRoot?.querySelector('vaadin-context-menu');
```
tomivirkki, line 31: No need to querySelector, use
```js
@internalProperty()
private items?: ContextMenuItem[];
```

and bind it to the context menu in the template, just like with the other examples.

yuriy-fix: Updated, thanks for the hint!

frontend/demo/component/contextmenu/context-menu-checkable.ts
yuriy-fix: My bad, have updated that one.

frontend/demo/component/contextmenu/context-menu-dividers.ts
yuriy-fix: Removed.

frontend/demo/component/contextmenu/context-menu-keyboard-shortcuts.ts
yuriy-fix: True that, removed the example.

frontend/demo/component/contextmenu/context-menu-keyboard-shortcuts.ts
yuriy-fix: Example removed.

frontend/demo/component/contextmenu/context-menu-presentation.ts
yuriy-fix: Updated.

frontend/demo/component/contextmenu/context-menu-checkable.ts
yuriy-fix: Updated.

frontend/demo/component/contextmenu/context-menu-checkable.ts
```diff
@@ -0,0 +1,48 @@
...
+  private items = [
+    { text: 'Abigail Lewis' },
+    { text: 'Allison Torres', checked: true },
+    { text: 'Anna Myers' },
+    { text: 'Lauren Wright' },
+    { text: 'Tamaki Ryushi' }
+  ];
+
+  @internalProperty()
+  private selectedItem?: ContextMenuItem = this.items[1];
```
tomivirkki, line 31: Avoid magic numbers

```suggestion
  private selectedItem = this.items.find(item => item.checked);
```

tomivirkki: Even cleaner approach would be not to maintain the redundant/duplicate `checked` property in the items but map the context-menu specific `checked` property to context-menu at the point where the property is bound...something like:

```js
.items=${this.items.map(item => {
  return { ...item, checked: item === this.selectedItem };
})}
```

frontend/demo/component/contextmenu/context-menu-dividers.ts
tomivirkki: Didn't mean only this instance...

frontend/demo/component/contextmenu/context-menu-left-click.ts
```diff
@@ -0,0 +1,48 @@
...
+  @internalProperty()
+  private gridItems: Person[] = [];
+
+  async firstUpdated() {
+    this.gridItems = await getPeople();
+  }
+
+  render() {
+    return html`
+      <hint-badge message="Open the Context Meny by clicking a Grid row."></hint-badge>
```
tomivirkki, line 35: Typo "Meny"
```suggestion
      <hint-badge message="Open the Context Menu by clicking a Grid row."></hint-badge>
```

frontend/demo/component/contextmenu/context-menu-left-click.ts
```diff
@@ -0,0 +1,48 @@
...
+      <!-- tag::snippethtml[] -->
+      <vaadin-context-menu open-on="click" .items=${this.items}>
```
tomivirkki, line 37: The example description says "Open the Context Menu by clicking a Grid row.". If you define the context menu like this, it allows you to open it anywhere, for example over the header. Same issue in some other examples also.

frontend/demo/component/contextmenu/context-menu-best-practices.ts
```diff
@@ -0,0 +1,73 @@
...
+    `;
+  }
+
+  menuBarRenderer(root: HTMLElement) {
+    if (root.firstElementChild) {
+      return;
+    }
+
+    const menuBar = document.createElement('vaadin-menu-bar');
+    menuBar &&
```
tomivirkki, line 59: This cannot be falsy

frontend/demo/component/contextmenu/context-menu-checkable.ts
yuriy-fix: Updated.

frontend/demo/component/contextmenu/context-menu-left-click.ts
yuriy-fix: Updated with workaround.
Haven't applied it for `left-click` example as it won't be working properly over there. (we cannot stop propagation of the `click` event)

frontend/demo/component/contextmenu/context-menu-best-practices.ts
yuriy-fix: Removed redundant check.

frontend/demo/component/contextmenu/context-menu-checkable.ts
tomivirkki: With the latter/cleaner approach, the `checked: true` shouldn't exist in the `items` (it's only mapped at the point where it'd bound). In that case `this.items[1];` makes sense (because the index isn't defined in 2 places).

tomivirkki: > Nit: have to compare text prop, because checked won't be in sync.

This will have them in sync and it guarantees that `selectedItem` is always one of the instances in `items`:
`this.selectedItem = this.items.find(item => item.text === e.detail.value.text);`

frontend/demo/component/contextmenu/context-menu-left-click.ts
tomivirkki: > we cannot stop propagation of the click event

Why not?

yuriy-fix: Closing of the Context Menu and text selection won't work on header and footer rows in that case 🤔

frontend/demo/component/contextmenu/context-menu-checkable.ts
yuriy-fix: Added `this.items[1]` back and updated listener.

frontend/demo/component/contextmenu/context-menu-left-click.ts
tomivirkki: Text selection seems to work (it wouldn't work if you have the context menu opening on click).

Context-menu not canceling on header click doesn't sound too tragic, but if you will, you can improve the logic by not stopping propagation if the context-menu is already open (probably an overkill).

frontend/demo/component/contextmenu/context-menu-best-practices.ts
```diff
@@ -0,0 +1,79 @@
...
+    `;
+  }
+
+  menuBarRenderer(root: HTMLElement) {
+    if (root.firstElementChild) {
+      return;
+    }
+
+    const menuBar = document.createElement('vaadin-menu-bar');
+    menuBar.items = [{ component: this.makeIcon('vaadin:ellipsis-dots-v'), children: this.items }];
```
jouni, line 60:
```suggestion
    menuBar.items = [{ component: this.makeIcon('vaadin:ellipsis-dots-v'), children: this.items }];
    menuBar.theme = 'icon';
```

jouni (approved):
Apart from the weird grid scroll issue (see video, desktop Safari 14.0.3, Responsive Design Mode), and the global reset styles leaking to `<hr>` elements (dividers in the menus), looks good.


https://user-images.githubusercontent.com/66382/110529154-90500800-8121-11eb-8447-038a2db7037b.mov

frontend/demo/component/contextmenu/context-menu-left-click.ts
yuriy-fix: Have updated the example.

tomivirkki:
Screenshot 2021-03-16 at 15 58 52

This dropdown was really unexpected. What I was looking after with context-awareness in the examples is that over some items the "Edit" and "Delete" could be disabled if they're considered read-only or such. Let's not add another example for this as it wasn't in the original designs.

frontend/demo/component/contextmenu/context-menu-best-practices.ts
```diff
@@ -0,0 +1,80 @@
...
+  private gridItems = [
+    { filename: 'Annual Report.pdf', size: '23 MB' },
+    { filename: 'Financials.xlsx', size: '42 MB' }
+  ];
+
+  render() {
+    return html`
+      <hint-badge
+        message="Open the Context Menu by right-clicking (desktop) or long-pressing (mobile) a Grid row. Alternatively, use the Menu Bar in the last column."
+      ></hint-badge>
```
tomivirkki, line 37: Why wouldn't the context-menu examples use admonition blocks for hints like this, like the other examples?

frontend/demo/component/contextmenu/context-menu-left-click.ts
tomivirkki: The workaround lgtm. It might take someone browsing the documentation a while to understand what the workaround is for. There could be a comment (in every workaround) explaining why it's there.

articles/ds/components/context-menu/index.asciidoc
```diff
@@ -25,6 +25,9 @@ In the meantime, you can view the previous examples for the Java/Flow component.
 - https://vaadin.com/components/vaadin-context-menu/java-examples[Java Component Examples]
 ====
 
+[IMPORTANT]
+Open the Context Menu by right-clicking (desktop) or long-pressing (mobile) a Grid row.
+
```
jouni, line 30: Not sure, but you might be able to use `include::[]` from within the same article. Worth trying at least, so we don’t copy-paste the same thing multiple times.

```asciidoc
// tag::example-instructions[]
[IMPORTANT]
Open the Context Menu by right-clicking (desktop) or long-pressing (mobile) a Grid row.
// end::example-instructions[]
```

```asciidoc
include::index.asciidoc[tag=example-instructions]
```

yuriy-fix: Seems to be working. Thank you! 🙇

frontend/demo/component/contextmenu/context-menu-best-practices.ts
yuriy-fix: Replaced `hint-badge` with admonition.

frontend/demo/component/contextmenu/context-menu-left-click.ts
yuriy-fix: Added comment and removed `context-aware` example based on the comment below.

tomivirkki:
A couple of nits about this example:
- You can open two context menus at one
- The buttons with the icons look huge.
- Why not use "tertiary" theme as [suggested](https://docs.google.com/document/d/1o9Qxz8QojsxwoEZymj2MDgGT4gSb3sPleegknckVUTg/edit?disco=AAAALJdKM8c) in the design.

Screenshot 2021-03-17 at 15 55 44

frontend/demo/component/contextmenu/context-menu-best-practices.ts
```diff
@@ -0,0 +1,77 @@
...
+    return html`
+      <!-- tag::snippethtml[] -->
+      <vaadin-context-menu .items=${this.items}>
+        <vaadin-grid .items=${this.gridItems} @vaadin-contextmenu=${this.onContextMenu}>
+          <vaadin-grid-column label="Filename" path="filename"></vaadin-grid-column>
+          <vaadin-grid-column label="Size" path="size"></vaadin-grid-column>
+          <vaadin-grid-column
+            width="80px"
+            flex-grow="0"
+            .renderer="${this.menuBarRenderer.bind(this)}"
```
DiegoCardoso, line 42: Instead of forcing the binding here, you can make the `menuBarRenderer` become a class property

(`menuBarRenderer = (root: HTMLElement) => {...}`)

yuriy-fix: Thanks for the hint! Updated 🙇

DiegoCardoso (approved):

PR #221, opened 2021-02-25
feat: DS publisher themability

tomivirkki:
Related to https://github.com/vaadin/docs-app/issues/199

This PR relocates the vcom-specific files from `docs-app` project here. The PR adds the files as required by `docs-app` themability feature.

To test this PR, use `dspublisher-themability` branch of `docs-app`

- `dspublisher/theme/global.css` includes global styles to be applied on the whole app.
- `dspublisher/theme/header.ts` exports a Web Component that gets included as the app header.


jouni:
We need to replicate the same files in all the different version branches. That’s probably unavoidable. It adds some amount of risk that we only update the latest branch and forget others.

dspublisher/theme/header.ts
```diff
@@ -0,0 +1,26 @@
...
+      window.location.hostname == 'preview.vaadin.com'
+        ? 'https://preview.vaadin.com/vaadincom/haas-service/haas-loader.js'
+        : 'https://vaadin.com/vaadincom/haas-service/haas-loader.js';
+    return script;
+  }
+
+  render() {
+    return html`
+      <div className="vcom-header" id="haas-container"></div>
+      ${this.haasImportScript()}
```
jouni, line 23: This should include the feedback banner element here as well: https://github.com/vaadin/docs-app/compare/dspublisher-themability#diff-81b4a456256cc32691afd44943ddbfb35508f2049222afb47aa0e1b158ae79d6R190-R208

jouni:
> We need to replicate the same files in all the different version branches

I wonder if a git submodule could alleviate this issue, like we discussed today…

jouni:
Tested locally, the works as expected.

dspublisher/theme/header.ts
tomivirkki: Done. One drawback is that the feedback banner's height isn't static so it can't be defined in `global.css`. Now that it's imported dynamically, there is an additional layout shift once the `<dspublisher-header>` gets defined and the banner height gets applied. Good thing is that this is a temporary thing.

jouni: Ah, didn’t consider that. Yeah, should be a short-lived regression.

jouni:
Regarding the feedback banner height, we could fix the height, perhaps using two breakpoints (currently it has 3–4 layout variations).

jouni (approved):

PR #227, opened 2021-03-03
chore: add OSGi services tutorial

### LINKED ISSUE #194: Document the usecases for using OSGi declarative services features with Vaadin
It's already possible to use every OSGi feature inside Vaadin application.
Sometimes it requires an additional boilerplate code and there is no way to use navigation target as a service via DS.
But it's quite easy to use DS anyway programmatically from any Vaadin view/component.

This should be documented with explanation why we don't provide this feature out of the box and all the ways that are possible already to do the same things.

This is the source of this documentation : 
https://github.com/vaadin/flow/issues/9107

denis-anisimov:
fixes #194


articles/flow/integrations/osgi-advanced.asciidoc
```diff
@@ -0,0 +1,310 @@
+---
+title: Advanced Using Vaadin with OSGi
```
jouni, line 2: I feel we should add this as a secondary tab under the main OSGi article. @magi42, would you agree?

Or, it could be a sub-article under the main one (visible in the sidebar navigation, nested under the current OSGi article).

jouni (changes requested):

articles/flow/integrations/osgi-advanced.asciidoc
jouni: In any case, the title needs work.

magi42: Yes, either as a tab or in a section. Should Karaf also be there?
Tabs are good for similar items, and maybe "Basic" and "Advanced" are OK tabs as well.
Yes the title really needs work as all other articles under Integrations are very short.

pleku: This is done - I used the use case as the title as it is what the whole section is about - I think it is easier for people to discover services usage docs if it is directly in the title.

jouni:
There’s a broken cross reference, that’s why the preview build isn’t showing up. I’m figuring out with Netlify how to get the deploy to fail as well if the build fails, so that the PR check would not show that everything is okay.

    1:24:12 PM: error Broken cross reference ../routing/overview in flow/integrations/osgi/osgi-basic.asciidoc

jouni:
Another broken cross reference:

    error Broken cross reference osgi-basic in flow/integrations/osgi-karaf.asciidoc

It would be best if you run the development build yourself, then you can see all the broken references immediately (the preview build fails on the first one it encounters).

jouni:
@denis-anisimov and @pleku, what’s the status of this PR?

pleku:
@jouni now the broken links are fixed, this should IMO be merged.

magi42:
Wait, I'm doing a review.

magi42:
PLEASE check my changes. I removed one example, aiming to simplify the text. I thought that the MainView actually needs the check for the availability of the service (it's always available while the service may not be), so need to give that example right away. Somehow I feel that the text could be shortened to half the size, but I didn't go into that.

articles/flow/integrations/overview.asciidoc
```diff
@@ -24,6 +24,6 @@ In addition to running Vaadin a application in a web page, you can embed it in a
 OSGi is an open source specification that describes a modular system and a dynamic service platform for Java.
 Application and component bundles can be remotely installed, started, stopped, updated, and uninstalled without requiring a reboot.
 
-<<osgi-karaf#, Karaf>>::
+<<osgi/osgi-karaf#, Karaf>>::
```
jouni, line 27: The naming of the articles makes me wonder if these two articles ("OSGi" and "Apache Karaf") should be nested in a hierarchy. Can Karaf be seen as an extension/customization of OSGi?

jouni: But the naming of the articles could be simplified. No need to keep the `osgi-` prefix in the file names.

jouni (commented):
Up to @magi42 when it’s good to go.

jouni (approved):

articles/flow/integrations/overview.asciidoc
```diff
@@ -20,10 +20,10 @@ It is part of the Java EE platform, but similar in purpose to autowiring in Spri
...
 In addition to running Vaadin a application in a web page, you can embed it in a web page, which could be part of some other application.
 
-<<osgi-basic#, OSGi>>::
-OSGi is a specification that describes a modular system and a service platform for Java.
+<<osgi/osgi-basic#, OSGi>>::
+OSGi is an open source specification that describes a modular system and a dynamic service platform for Java.
 Application and component bundles can be remotely installed, started, stopped, updated, and uninstalled without requiring a reboot.
 
-<<osgi-karaf#, Karaf>>::
+<<osgi/osgi-karaf#, Karaf>>::
```
pleku, line 27:
```suggestion
<<osgi/karaf#, Karaf>>::
```

articles/flow/integrations/overview.asciidoc
```diff
@@ -20,10 +20,10 @@ It is part of the Java EE platform, but similar in purpose to autowiring in Spri
 <<embedding/overview#, Embedding Applications>>::
 In addition to running Vaadin a application in a web page, you can embed it in a web page, which could be part of some other application.
 
-<<osgi-basic#, OSGi>>::
-OSGi is a specification that describes a modular system and a service platform for Java.
+<<osgi/osgi-basic#, OSGi>>::
```
pleku, line 23:
```suggestion
<<osgi/basic#, OSGi>>::
```

pleku (changes requested):
I think there are two broken links after the rename of files.

magi42 (approved):

pleku (approved):

pleku (approved):

jouni:
Ah, thanks for fixing the cross refs! Apologies for being completely thoughtless about that 🤦

PR #229, opened 2021-03-04
Described how to use new `VaadinSpringSecurity` helper

joheriks:
Part of vaadin/spring#678.


haijian-vaadin (changes requested):
Should we also show how to use the other 2 helper methods? `getDefaultHttpSecurityPermitMatcher` and `getDefaultWebSecurityIgnoreMatcher`. For example, a user might want to set up a custom matcher based on the default one?

joheriks:
> Should we also show how to use the other 2 helper methods? `getDefaultHttpSecurityPermitMatcher` and `getDefaultWebSecurityIgnoreMatcher`. For example, a user might want to set up a custom matcher based on the default one?

Can do. Just wondering where would be the best place to add it, `spring-login.asciidoc`?

haijian-vaadin:
> > Should we also show how to use the other 2 helper methods? `getDefaultHttpSecurityPermitMatcher` and `getDefaultWebSecurityIgnoreMatcher`. For example, a user might want to set up a custom matcher based on the default one?
> 
> Can do. Just wondering where would be the best place to add it, `spring-login.asciidoc`?

`spring-login.asciidoc` sounds a good place.

joheriks:
> > > Should we also show how to use the other 2 helper methods? `getDefaultHttpSecurityPermitMatcher` and `getDefaultWebSecurityIgnoreMatcher`. For example, a user might want to set up a custom matcher based on the default one?
> > 
> > 
> > Can do. Just wondering where would be the best place to add it, `spring-login.asciidoc`?
> 
> `spring-login.asciidoc` sounds a good place.

Done.

haijian-vaadin (approved):

PR #241, opened 2021-03-08
chore: remove unnecessary infobox

haijian-vaadin:
The code example above the infobox was updated so that the infobox became unnecessary.


jouni (approved):
I can give a 👍 , but I don’t really anything about the actual content 😄

jouni:
Also, no need to update from base branch, just merge when you are ready.

platosha (approved):

PR #242, opened 2021-03-08
fix the code examaple in cache client-side data article

haijian-vaadin:
The `keysToClearOnLogout` is in-memory only. So if you start the app while offline, then log out, no keys would get cleared out. 
Fixes the example by aligning it with what's done in the tutorial.


articles/fusion/pwa/cache-client-side-data.asciidoc
```diff
@@ -74,16 +74,26 @@ export async function cacheable<T>(fn: () => Promise<T>, key: string, defaultVal
...
+    cache[key] = result;
+    localStorage.setItem(CACHE_NAME, JSON.stringify(cache));
   } catch {
     // if failed to retrieve the data from backend, try localStorage.
-    const cached = localStorage.getItem(key);
+    const cache = getCache();
+    const cached = cache[key];
     // use the cached data if available, otherwise the default value.
-    result = cached ? JSON.parse(cached) : defaultValue;
+    result = cached ? cached : defaultValue;
```
joheriks, line 85: Haven't tested it, but is there a risk that if the cached value is "falsish" (e.g, `0`, `false`, `""`, ...) that it would be overridden by `defaultValue`?

haijian-vaadin: very good catch, updated the code to check if the value is undefined.

PR #248, opened 2021-03-09
Feature/upload

alkali47:


rolfsmeds (approved):

jouni:
Should we merge this even if all the examples are just stubs?

Haprog:
Rebased onto `latest` while squashing some of the commits to clean up. Let's see if Vale is happier now without the merge commits (that synchronized with `latest`).

Haprog:
Vale errors are fixed now. There are still some Vale warnings that might be considered to be changed. I'm working on the TS demos now.

Haprog:
I converted this PR to a draft so it won't get merged by accident before the demos are done.

Haprog:
All TS demos before the "Internationalization (i18n)" header have now been implemented (pushed in case someone wants to give early feedback). I'll probably implement the rest today after lunch.

I added a upload-demo-helpers.ts which contains code for mocking the upload requests (XHR). This is based on what we had in place for vaadin-upload tests, but I converted it to TS. The small 3rd party lib it's using from npm essentially hasn't been updated in 11 years so I though it's safer to just include its contents directly (and I converted it to TS) as it's only one file.

Haprog:
All TS demos have now been implemented as specified [here](https://docs.google.com/document/d/1oNHnaJd9ojR--5FocAtOa5Jx-p_I3aiGQy2fq2-ejOc/edit?usp=sharing).

This PR should probably still stay as a draft until the Java demos have been implemented too?

jouni:
> This PR should probably still stay as a draft until the Java demos have been implemented too?

Actually no. We should get the TS examples out there asap, and then continue adding the Java versions. No need to keep this work waiting.

articles/ds/components/upload/index.asciidoc
```diff
@@ -11,15 +11,355 @@ section-nav: incomplete
...
+[source, typescript]
+----
+include::{root}/frontend/demo/component/upload/upload-file-size.ts[render,tags=snippet,indent=0,group=TypeScript]
+----
+
+[source, java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/upload/UploadFileSize.java[render,tags=snippet,indent=0,group=Java]
+----
+--
```
tomivirkki, line 165: All the Java examples should be commented out.

frontend/demo/component/upload/upload-auto-upload-disabled.ts
```diff
@@ -0,0 +1,53 @@
...
+      <div style="text-align: center; margin-top: var(--lumo-space-l)">
+        <vaadin-button @click="${this.setFakeStatus}">Reset demo</vaadin-button>
+      </div>
+      <!-- tag::snippet[] -->
+    `;
+  }
+  // end::snippet[]
+
+  setFakeStatus() {
+    const upload = this.shadowRoot!.querySelector('vaadin-upload');
```
tomivirkki, line 44: Try to minimize use of `shadowRoot.querySelector` (and direct DOM access in general) and instead use Lit template data-binding. For examples, look at the other demos. In this particular case, you could have a class property `files` that you bind to the component with `<vaadin-upload .files=${this.files}`.

Haprog: This fake/mock status setup was deliberately set like this as I was considering hiding these with `hidden-full-source-line` but what do you think?

Haprog: The intention of the demos is not to show how you can mock the initial status of the upload, but it's used for demo purposes in some of the demos.

tomivirkki: Here's an example of the simplified upload-basic demo https://gist.github.com/tomivirkki/0c8d3ebfcec0323d9e5b627b5cc8e5bd

frontend/demo/component/upload/upload-drag-and-drop.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor() {
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot);
+  }
+
+  render() {
+    return html`
+      <div class="flex">
```
tomivirkki, line 36: Why not `<vaadin-horizontal-layout>`?

frontend/demo/component/upload/upload-file-size.ts
```diff
@@ -0,0 +1,48 @@
...
+        max-files="1"
+        .maxFileSize="${maxFileSizeInBytes}"
+        @file-reject="${this.fileRejectHandler}"
+      ></vaadin-upload>
+    `;
+  }
+  // end::snippet[]
+
+  fileRejectHandler(event: UploadFileReject) {
+    window.alert(event.detail.file.name + ' error: ' + event.detail.error);
```
tomivirkki, line 46: Use Vaadin notification instead of a `window.alert`. Examples can be found in the vaadin-messages demos.

Haprog: Ok, makes sense. This alert was basically copied from the old demos.

frontend/demo/component/upload/upload-drag-and-drop.ts
Haprog: Didn't even cross my mind to use that one when using Fusion. Is it used in other TS demos? I can try it and see if I can get away with less custom CSS.

frontend/demo/component/upload/upload-auto-upload-disabled.ts
Haprog: Thanks. That does look a bit nicer without `querySelector` and `firstUpdated` though I'd like to avoid showing binding to `files` property in demos (especially since it would be shown in the non-expanded source already) since that's not something you would normally do when using `vaadin-upload`.

Also regarding `@click="${() => (this.files = createFakeFiles())}"` I think it would be better to prefer method references instead of anonymous functions created in `render`. Though maybe we can make an exception for this reset demo button use case?

frontend/demo/component/upload/upload-all-files.ts
```diff
@@ -0,0 +1,67 @@
...
+      <vaadin-upload no-auto></vaadin-upload>
+      <p>
+        <vaadin-button theme="primary" @click="${this.uploadFiles}">
+          Upload All Files
+        </vaadin-button>
+      </p>
+      <!-- end::snippet[] -->
+      <div style="text-align: center; margin-top: var(--lumo-space-l)">
+        <vaadin-button @click="${this.setFakeStatus}">Reset demo</vaadin-button>
+      </div>
```
tomivirkki, line 42: These "Reset demo" -buttons are not in the example designs and none of the other demos have such buttons either. They bring unnecessary logic to the example source code. If someone wants to reset a demo, they can refresh the browser anytime.

articles/ds/components/upload/index.asciidoc
```diff
@@ -11,15 +11,355 @@ section-nav: incomplete
...
+
+== Upload Restrictions
+
+You can set three types of restrictions: file format, file count and file size.
+
+Exceptions that arise from the user violating any of the imposed restrictions aren't shown in the UI by default.
+Use a <<#listeners, File Rejected listener>> to catch these exceptions and, for example, a <<../notification#, Notification>> to inform the user about the problem at hand and any potential solutions.
+
+However, the user should be informed upfront about any file upload restrictions. Maximum number of files allowed, file size and format limitations should all be communicated clearly to avoid exceptions whenever possible.
```
yuriy-fix, line 103: nit: new line before `Maximum number ...`

articles/ds/components/upload/index.asciidoc
```diff
@@ -11,15 +11,355 @@ section-nav: incomplete
...
+
+=== File Format
+
+Upload can be configured to only accept files of specific formats. The acceptable file formats are set using MIME type patterns or file extensions, for example `"video/*"`, `"image/tiff"` or `".pdf"` and `"audio/mp3"`.
```
yuriy-fix, line 107: nit: new line before `The acceptable ...`.
P.S. Vale should notify about those locally with `Vaadin.OneSentencePerLine`.

frontend/demo/component/upload/upload-all-files.ts
```diff
@@ -0,0 +1,67 @@
+import '../../init'; // hidden-full-source-line
+import { createFakeUploadFiles } from './upload-demo-helpers'; // hidden-full-source-line
+// hidden-full-source-line
```
yuriy-fix, line 3: nit: extra comment.

frontend/demo/component/upload/upload-all-files.ts
Haprog: Ok. I can remove them. I just thought this would be a nice addition and wanted to try how it would feel.

The page and demo sources will look a bit cleaner after removing those.

src/main/java/com/vaadin/demo/component/upload/UploadAllFiles.java
```diff
@@ -0,0 +1,20 @@
+package com.vaadin.demo.component.upload;
```
yuriy-fix, line 1: I guess there is no need to add Java files within this PR as those won't be used. 
We are not adding those in other (ts examples) PRs either.

frontend/demo/component/upload/upload-auto-upload-disabled.ts
tomivirkki: > that's not something you would normally do when using vaadin-upload

This is a fair point.

> I think it would be better to prefer method references instead of anonymous functions

Inline functions would be bad in some cases, like for `renderer` functions, since reassigning would cause a component re-render also. We've used anonymous inline functions in some very simple cases, where the only change on render is the listener function reassignment. When optimization is needed, there's the [`guard` directive](https://github.com/vaadin/docs/blob/1d9dd17548a31a091a73d11192d5ed43ac0ceb07/frontend/demo/component/dialog/dialog-basic.ts#L33) and sure it's an option to have the function out of the template also. Both bring some more verbosity to the example.

frontend/demo/component/upload/upload-drag-and-drop.ts
```diff
@@ -0,0 +1,50 @@
...
+        <!-- tag::snippet[] -->
+        <div>
+          <vaadin-upload ?nodrop="${false}"></vaadin-upload>
```
yuriy-fix, line 39: Can we simplify it and use: `<vaadin-upload></vaadin-upload>` and `<vaadin-upload nodrop></vaadin-upload>` as it has `reflectToAttribute`? 🤔

articles/ds/components/upload/index.asciidoc
jouni: > P.S. Vale should notify about those locally with Vaadin.OneSentencePerLine.

Yeah, would be nice. If you have the time to tweak it, I’d be grateful 🙏

articles/ds/components/upload/index.asciidoc
Haprog: Ok, thanks. I didn't even look at this since those were added before I started working on this, but I can take a look at fixing the one sentence per line issues.

frontend/demo/component/upload/upload-all-files.ts
Haprog: @yuriy-fix This is actually intentional so that the code snippet doesn't begin with unnecessary empty line.

frontend/demo/component/upload/upload-drag-and-drop.ts
Haprog: Good point. I'm not sure why I did it this way. I'll try the simpler approach.

frontend/demo/component/upload/upload-all-files.ts
Haprog: Actually we could just remove this whole line. Other component demos seem to have an empty line here though which manifests in an empty line at the beginning of the code snippets (when expanded).

frontend/demo/component/upload/upload-drag-and-drop.ts
Haprog: Now I remember why I did it this way.
https://cdn.vaadin.com/vaadin-upload/4.4.1/#/elements/vaadin-upload#property-nodrop

The API docs for `nodrop` says:
> Define whether the element supports dropping files on it for uploading. By default it's enabled in desktop and disabled in touch devices because mobile devices do not support drag events in general. Setting it false means that drop is enabled even in touch-devices, and true disables drop in all devices.

And basically this same text is actually on the new Upload docs page right above the demos. But now that I actually tried it on my phone, it doesn't seem to work as documented. At least I don't see the "Drop files here" target on my phone even with `<vaadin-upload ?nodrop="${false}"></vaadin-upload>`.

So it seems to me that either the component is broken or the docs for this are wrong.

Haprog: Actually now I realize my mistake. Of course it doesn't work with `?nodrop="${false}"` syntax. It needs to be `.nodrop="${false}"` to explicitly set the property to false. With that it works on mobile too.

We can simplify this to:
```html
<vaadin-upload .nodrop="${false}"></vaadin-upload>
<vaadin-upload nodrop></vaadin-upload>
```

Haprog: But it seems to be a bit bad API as you can't set `nodrop` to false via attribute. Though maybe that is ok since it's probably not a common use case anyway?

frontend/demo/component/upload/upload-drag-and-drop.ts
Haprog: I tried using `<vaadin-horizontal-layout>` for this but it doesn't really help at all since it's not meant to wrap, and the spacing theme variants for it are implemented naively with `margin-left` only and it seems to behave weirdly in a way that the items do not align nicely when it wraps if I add `flex-wrap: wrap`. This layout is supposed to wrap on small screens like mobile when there's not enough horizontal space. And I need some custom CSS here in any case also for aligning the the paragraphs to the bottom (using `justify-content: space-between` here) when the flex isn't wrapping.

To have the same responsive layout with `<vaadin-horizontal-layout>` I still need all the same custom CSS except for `display: flex` so it doesn't seem to make sense to use `<vaadin-horizontal-layout>` here.

tomivirkki: If you need it to be responsive, would a `<vaadin-form-layout>` cut it? We've used form layout in [some examples](https://vaadin.com/docs/latest/ds/components/input-fields/#focus) for that purpose.

frontend/demo/component/upload/upload-drag-and-drop.ts
yuriy-fix: I have gone through that one as well and understood the intention to show how to disable it for mobile, but I guess if someone is checking TypeScript examples they will easily realize how to bind boolean prop. 
I guess I don't have strong preferences over here, so we can leave it as is.

frontend/demo/component/upload/upload-all-files.ts
jouni: Why do we leave that empty line there? Why not have all imports together?

articles/ds/components/upload/index.asciidoc
Haprog: Done. See new PR https://github.com/vaadin/docs/pull/300

frontend/demo/component/upload/upload-file-size.ts
Haprog: Done. See new PR https://github.com/vaadin/docs/pull/300

frontend/demo/component/upload/upload-all-files.ts
Haprog: Done. See new PR https://github.com/vaadin/docs/pull/300

articles/ds/components/upload/index.asciidoc
Haprog: Done. See new PR https://github.com/vaadin/docs/pull/300

articles/ds/components/upload/index.asciidoc
Haprog: Done (content fix). See new PR https://github.com/vaadin/docs/pull/300

frontend/demo/component/upload/upload-all-files.ts
Haprog: Done. See new PR https://github.com/vaadin/docs/pull/300

src/main/java/com/vaadin/demo/component/upload/UploadAllFiles.java
Haprog: Done (removed java files). See new PR https://github.com/vaadin/docs/pull/300

frontend/demo/component/upload/upload-drag-and-drop.ts
Haprog: Done (simplified slightly). See new PR https://github.com/vaadin/docs/pull/300

frontend/demo/component/upload/upload-drag-and-drop.ts
Haprog: Hmm.. I'll try if I can make that work here with less code compared to the current custom CSS layout (while keeping the layout and alignment nice on both mobile and desktop).

Haprog: Updated (see https://github.com/vaadin/docs/pull/300). I was able to simplify the layout code for this demo (and error-messages demo) by using `vaadin-form-layout`. Thanks for the tip!

The layout is not exactly the same as it used to be but I'm happy with it.

PR #251, opened 2021-03-09
refactor: move paging to getPeople

DiegoCardoso:


frontend/demo/component/tree-grid/tree-grid-basic.ts
```diff
@@ -22,11 +22,19 @@ export class Example extends LitElement {
   async dataProvider(params: GridDataProviderParams, callback: GridDataProviderCallback) {
```
tomivirkki, line 22: The final data provider should look something like:
```js
async dataProvider(params: GridDataProviderParams, callback: GridDataProviderCallback) {
  // The requested data range and the full length of its
  // hierarhcy level is fetched from the data service
  const { people, hierarchyLevelSize } = await getPeople({
    count: params.pageSize,
    startIndex: params.page * params.pageSize,
    managerId: (params.parentItem as Person)?.id
  });
  callback(people, hierarchyLevelSize);
}
```

...where the `hierarhcyLevelSize` is the length of `people` right before it's sliced in the data service.

DiegoCardoso: Updated to return `{ people, hierarchyLevelSize }` and update all usages of `getPeople` to reflect the new API.

tomivirkki: The data providers still didn't look as above, they had duplicate logic for slicing the data result even though the slicing was moved to the data service. Did some cleanup [here](https://github.com/vaadin/docs/pull/251/commits/c95468d8cef6a51e574344d66e20ab080237358c)

tomivirkki (approved):

PR #261, opened 2021-03-12
Add Java variants to Messages examples.

Peppe:
- Remove Message example
- Add Java variations of each example
- Better example data


Peppe:
I'm having problems developing this locally, so I thought of doing a draft PR and verifying that it works similarly on Netlify as locally.

src/main/java/com/vaadin/demo/component/messages/MessagesBasic.java
```diff
@@ -0,0 +1,59 @@
...
+                "Linsey Listy", person.getPictureUrl());
+        message2.setUserAbbreviation("LL");
+        message2.setUserColorIndex(2);
+        list.setItems(Arrays.asList(message1, message2));
+
+        VerticalLayout chatLayout = new VerticalLayout(list, input);
+        chatLayout.setHeight("500px");
+        chatLayout.setWidth("400px");
+        chatLayout.expand(list);
+        // end::snippet[]
```
yuriy-fix, line 54: Missing `add(chatLayout)` is a part of the problem not displaying the example

Peppe: Done.

articles/ds/components/messages/index.asciidoc
```diff
@@ -58,8 +51,22 @@ userColorIndex: number;
 userImg: string;
 ----
 
-Here is an example of an initialized `items` array with two messages:
+Here is an example of an initialized `items` array with two messages, in Java and TypeScript:
```
pekam, line 54: I think that we shouldn't have source code as part of the text like this. We have the code examples for that purpose, and with those you only see them in your preferred language.

Looking at the other component demos, they mostly speak on a more general level about the component in the text part, and then let the code examples show the API.

src/main/java/com/vaadin/demo/component/messages/MessagesBasic.java
```diff
@@ -0,0 +1,62 @@
...
+public class MessagesBasic extends Div {
+
+    public MessagesBasic() {
+        // tag::snippet[]
+        MessageList list = new MessageList();
+        MessageInput input = new MessageInput();
+        input.addSubmitListener(submitEvent -> {
+            MessageListItem newMessage = new MessageListItem(
+                    submitEvent.getValue(), Instant.now(), "Milla Sting");
+            newMessage.setUserAbbreviation("MS");
```
pekam, line 30: I don't think we need to use `setUserAbbreviation` just for the sake of covering all the APIs. It's just extra line of code which doesn't have any effect.

articles/ds/components/messages/index.asciidoc
```diff
@@ -13,23 +13,16 @@ page-links:
 Messages is a collection of components to display messages and to build chats, comment sections and activity logs.
```
pekam, line 13: A bit outside the scope of this PR, but I would avoid saying that you can build activity logs until you actually can.
```suggestion
Messages is a collection of components to display messages and to build chats and comment sections.
```

pekam (changes requested):

articles/ds/components/messages/index.asciidoc
Peppe: Is your suggestion to delete the the code snippets? 

I added it after in DX tests, people was looking on how to define `items`, and didn't know what all was possible to define in these objects (especially in JS). But yeah, the practice is not there in other components.

src/main/java/com/vaadin/demo/component/messages/MessagesBasic.java
Peppe: Okay. We should take it away from the TS version as well to have them match.

frontend/demo/component/messages/MessageListItem.ts
```diff
@@ -1,8 +1,8 @@
 export default interface MessageListItem {
```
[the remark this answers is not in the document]
Peppe: No we don't need it anymore, and it isn't used. Deleted.

articles/ds/components/messages/index.asciidoc
pekam: Maybe the TS code example should be changed from:
```html
<vaadin-message-list .items=${this.items}></vaadin-message-list>
```
to:
```html
<vaadin-message-list .items=${[
  {
    text: "foo"
  }..,
]}></vaadin-message-list>
```

src/main/java/com/vaadin/demo/component/messages/MessagesBasic.java
Peppe: Done.

jouni:
Build failure:

```
11:55:34 AM: [ERROR] ERROR in /opt/build/repo/frontend/demo/component/messages/message-basic.ts
11:55:34 AM: [ERROR] [tsl] ERROR in /opt/build/repo/frontend/demo/component/messages/message-basic.ts(8,29)
11:55:34 AM: [ERROR]       TS2307: Cannot find module './MessageListItem' or its corresponding type declarations.
11:55:34 AM: [ERROR]
```

articles/ds/components/messages/index.asciidoc
Peppe: I can see the point and I changed it. 

However, I'm not that much of a fan myself as inline defining messages is not what I would do in any real-world application and thus I don't consider this best practice. But I'm fine with it.

Done.

src/main/java/com/vaadin/demo/component/messages/MessagesBasic.java
```diff
@@ -0,0 +1,60 @@
...
```
pekam, line 30:
```suggestion
```

src/main/java/com/vaadin/demo/component/messages/MessagesBasic.java
```diff
@@ -0,0 +1,60 @@
...
+                "Nature does not hurry, yet everything gets accomplished.",
+                LocalDateTime.now().minusDays(1).toInstant(ZoneOffset.UTC),
+                "Matt Mambo");
+        message1.setUserColorIndex(1);
+        MessageListItem message2 = new MessageListItem(
+                "Using your talent, hobby or profession in a way that makes you contribute with something good to this world is truly the way to go.",
+                LocalDateTime.now().minusMinutes(55).toInstant(ZoneOffset.UTC),
+                "Linsey Listy", person.getPictureUrl());
+        message2.setUserColorIndex(2);
+        list.setItems(Arrays.asList(message1, message2));
```
pekam, line 48:
```suggestion
        list.setItems(message1, message2);
```

articles/ds/components/messages/index.asciidoc
pekam: Yeah, following best practices is of course a good idea, but it is just more important to easily see from the example what the property looks like.

And in any case, we're hard-coding the array in the code, so it's kinda impossible to really follow best practices. :sweat_smile:

pekam (approved):
These message demos are still standing out from other components as they go into API details in the descriptions, but changing those more is outside the scope of this PR.

PR #264, opened 2021-03-12
Update tabs-content.ts

markhm:
Fixed inconsistency


jouni:
Thanks for the fix, @markhm!

PR #269, opened 2021-03-16
feat: Add documentation about custom theme

caalador:
Add the custom theme feature documentation
to v14 branch.

Part of vaadin/flow#9983

### DIFF
--- a/articles/flow/styling/creating-a-custom-theme-class.asciidoc
+++ b/articles/flow/styling/creating-a-custom-theme-class.asciidoc
@@ -1,12 +1,14 @@
 ---
-title: Creating a Custom Theme
+title: Creating a Custom Theme Class
 order: 80
 layout: page
 ---
 
-= Creating a Custom Theme
+= Creating a Custom Theme Class
 ifdef::web[]
 endif::web[]
+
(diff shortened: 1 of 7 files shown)


jouni:
I think this is outdated already, as the article titles are different than in latest.

caalador:
So there has been changes to 14? master? after yesterday around 15?

jouni:
Please check the cross references. At least the following is broken:

```
error Broken cross reference ../../flow/integrations/embedding/overview in flow/styling/custom-theme.asciidoc
```

jouni:
One more thing: please delete the old `creating-a-custom-theme.asciidoc` article.

caalador:
Well it's sort of valid for 14 and in 14.6 you can have your own `Theme extends Lumo` in combination with the new custom theme feature.
@pleku what do you think about the `extends AbstractTheme` documentation.

jouni:
Ah, right, sorry about. I got confused about this, that it was the previous new V19 article. 

But at least the article title (and file name) should be updated. We shouldn’t have two “Creating a Custom Theme” articles.

pleku:
I think there needs to be note in the old article that "it doesn't apply after 14.6 since, while it still works, the new custom theme feature that should be used"

jouni (approved):

PR #270, opened 2021-03-16
chore: upgrade prettier to version 2

Haprog:
Upgrading to prettier 2 makes it possible to use new TypeScript 3.8+ features like type-only imports and exports.
See https://prettier.io/blog/2020/03/21/2.0.0.html#typescript.

Use the new defaults `"trailingComma": "es5"`, `"arrowParens": "always"` and `"embeddedLanguageFormatting": "auto"` and apply them with the `--fix` option.
See https://prettier.io/blog/2020/03/21/2.0.0.html#breaking-changes and https://prettier.io/docs/en/options.html#embedded-language-formatting


Haprog:
Rebased onto latest.

Haprog:
Rebased onto latest (again).

tomivirkki:
> Turn off the newly available option embeddedLanguageFormatting to prevent unwanted transforms to html tagged template literals for now.

This would also prevent wanted transformations. Ideally, all the examples would have identically formatted HTML template content and things like indentation errors get automatically cleaned up.

Haprog:
@tomivirkki Ok. If you think it's better to start using that new option I can change it and apply --fix. It seemed that this would keep one whitespace at the beginning and end of html tagged template literals, which I think could probably be trimmed out too (manually).

tomivirkki (approved):

PR #271, opened 2021-03-16
Fusion security intro

### LINKED ISSUE #268: Fusion Security Document: Security in Vaadin Applications
part of https://github.com/vaadin/docs/issues/110

Have a Security in Vaadin Applications article in the Securtiy section, which gives an introduction to the Vaadin Fusion security architecture and how it works in practice. (Where and how does it determine if a user has access? Where does the auth info come from?) Further, release and security practices in developing and releasing Vaadin products are described (How would I replace the default Spring Security setup with something else? Which classes can I provide my own implementations for, and how?)

joheriks:
Fixes #268.


articles/fusion/security/intro.asciidoc
```diff
@@ -0,0 +1,71 @@
+---
+title: Security in Fusion applications
```
jouni, line 2:
```suggestion
title: Security in Fusion Applications
```

joheriks: Done.

haijian-vaadin (commented):
Should we mention **access control for Vaadin views must be based on the actual Vaadin views, and not only on the URL** and link to [this blog](https://vaadin.com/blog/the-dangers-of-using-the-wrong-abstraction-for-vaadin-access-control)?
Or maybe could do it in the authentication article, what do you think?

joheriks:
> Should we mention **access control for Vaadin views must be based on the actual Vaadin views, and not only on the URL** and link to [this blog](https://vaadin.com/blog/the-dangers-of-using-the-wrong-abstraction-for-vaadin-access-control)?
> Or maybe could do it in the authentication article, what do you think?

A summary of the information in the blog post (or a link) would be good to have in the docs. But I feel is place should be in the Flow security documentation (since it is about authorizing the views on the server side using `BeforeEnterListener` and so on).

haijian-vaadin (approved):
ok, then let's merge this PR. The blog uses Java views as an example, the principle is the same for Fusion views as well. I will mention that in the authentication doc.

PR #280, opened 2021-03-18
docs: add session of how to use Charts 9 in V14

DiegoCardoso:
Part of https://github.com/vaadin/vaadin-charts/issues/562

### DIFF
--- a/articles/ds/charts/java-api/charts-migrating-from-earlier-version.asciidoc
+++ b/articles/ds/charts/java-api/charts-migrating-from-earlier-version.asciidoc
@@ -1,0 +1,115 @@
+---
+title: Updating Vaadin Charts 9+ to V14
+order: 8
+layout: page
+---
+
+[[charts.migratingfromearlierversions]]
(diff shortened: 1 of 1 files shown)


articles/ds/charts/java-api/charts-migrating-from-earlier-version.asciidoc
```diff
@@ -0,0 +1,115 @@
+---
+title: Updating Vaadin Charts 9+ to V14
```
rolfsmeds, line 2: I would propose "Using Charts 9+ in V14" as a title, to avoid giving the impression that Charts 9 is upgraded to Charts 14.

rolfsmeds (changes requested):
LGTM except the title

articles/ds/charts/java-api/charts-migrating-from-earlier-version.asciidoc
DiegoCardoso: Changed the title and added a bit more of text on the beginning of the page.

PR #287, opened 2021-03-19
merge adding login form articles

haijian-vaadin:
Merge the 2 Spring login form into 1, just show how to use a custom login page, since not many people would use the default Spring login page.
Updated the reference links accordingly.
Related to https://github.com/vaadin/flow/issues/10246 and https://github.com/vaadin/flow/issues/10221


articles/fusion/security/spring-login.asciidoc
```diff
@@ -178,39 +179,101 @@ public static void configure(WebSecurity web) throws Exception {
...
+  {
+    path: '/login',
+    component: 'login-view'
+  },
+  // more routes
+}
+----
+
+→ *Step 3* - Access control in the UI.
+The access control for Vaadin views must be based on the actual Vaadin views, and not only on the URL filtering.
```
platosha, line 206: `Vaadin views` ... `actual Vaadin views` sounds repetitive.
 
Also confusing. Does that refer to URL filtering on the server provided by Spring Security?

This description refers to views — somehow after reading it I thought something needs to be additionally implemented in `my-view` to prevent access, but the code example here is about route actions.

haijian-vaadin: updated, any further comments?

platosha:
@haijian-vaadin I noticed that the `fusion/security/custom-spring-login.asciidoc` article still exists, both in as file and in preview deploy. Should it be deleted in this PR? I guess it’s no longer needed now that its content is moved.

haijian-vaadin:
> @haijian-vaadin I noticed that the `fusion/security/custom-spring-login.asciidoc` article still exists, both in as file and in preview deploy. Should it be deleted in this PR? I guess it’s no longer needed now that its content is moved.
Oh, that's a very good catch, thanks. I deleted all the references, forgot the delete the article itself. Now fixed

platosha (approved):

PR #288, opened 2021-03-19
Add TS examples for badge component

Lodin:


jouni:
I wonder, if the data types should actually be defined in Java (like the previous Person and Card objects, so that the same things can be used for the eventual Java examples?

frontend/demo/component/badge/badge-color.ts
```diff
@@ -14,9 +25,18 @@ export class Example extends LitElement {
 
   render() {
     return html`
-      <!-- tag::snippet[] -->
-      <span theme="badge pill contrast">Badge</span>
-      <!-- end::snippet[] -->
+      <section class="container">
```
yuriy-fix, line 28: We could use `vaadin-horizontal-layout` over here instead of `div.container`

frontend/demo/component/badge/badge-counter.ts
```diff
@@ -1,11 +1,18 @@
...
 
-import { html, LitElement, customElement } from 'lit-element';
-import '@vaadin/vaadin-checkbox/vaadin-checkbox';
+import '@vaadin/vaadin-list-box';
+import { html, LitElement, customElement, css } from 'lit-element';
 import { applyTheme } from 'Frontend/generated/theme';
 
 @customElement('badge-counter')
 export class Example extends LitElement {
+  static styles = css`
```
yuriy-fix, line 9: Nit: could probably be better to use inline styles instead of this one to align with other examples.

frontend/demo/component/badge/badge-highlight.ts
```diff
@@ -1,22 +1,109 @@
...
+      box-sizing: border-box;
+      width: calc(var(--lumo-space-xl) * 15);
+    }
+
+    .container vaadin-grid {
+      height: calc(var(--lumo-space-xl) * 5.125);
+    }
+  `;
+
+  @internalProperty() private _items: readonly Report[] = [];
```
yuriy-fix, line 83: Nit: no need to use underscore. In addition let's add new line:
```
@internalProperty()
private items: Person[] = [];
```
Example: https://github.com/vaadin/docs/blob/39cafb82efdba084028c88f38aabe65d75ab1fbd/frontend/demo/component/gridpro/grid-pro-basic.ts#L20

frontend/demo/component/badge/badge-highlight.ts
```diff
@@ -1,22 +1,109 @@
...
   render() {
     return html`
-      <!-- tag::snippet[] -->
-      <span theme="badge pill contrast">Badge</span>
-      <!-- end::snippet[] -->
+      <section class="container">
+        <!-- tag::snippet[] -->
+        <vaadin-grid .items=${this._items}>
+          <vaadin-grid-column path="report" header="Report"></vaadin-grid-column>
+          <vaadin-grid-column header="Due Date" .renderer=${renderDate}></vaadin-grid-column>
```
yuriy-fix, line 101: Example of setting renderer inline in other examples: https://github.com/vaadin/docs/blob/39cafb82efdba084028c88f38aabe65d75ab1fbd/frontend/demo/component/gridpro/grid-pro-editors.ts#L45
Probably should be aligned 🤔

src/main/java/com/vaadin/demo/domain/Profession.java
```diff
@@ -0,0 +1,45 @@
+package com.vaadin.demo.domain;
```
yuriy-fix, line 1: I guess there is no need in creating specifically `Profession` class as we can get those from `Person` (`people` data with smth like `${[...new Set(this.people.map(i => i.profession))]}`)

yuriy-fix (changes requested):
Badges styles are not actually applied in the demos, so for me in the demos there are just `span`s displayed 😅 
Screen Shot 2021-03-24 at 20 06 36
I have tried a small trick to use `badge` styles in the examples. You can take a look onto https://github.com/vaadin/docs/blob/39cafb82efdba084028c88f38aabe65d75ab1fbd/frontend/demo/component/tabs/tabs-badges.ts#L11. Probably we can come up with smth better for this one.

Lodin:
> Badges styles are not actually applied in the demos, so for me in the demos there are just spans displayed 😅

@yuriy-fix, that's strange... They are displayed correctly in the [netlify build](https://deploy-preview-288--vaadin-docs-preview.netlify.app/ds/components/badge).

frontend/demo/component/badge/badge-counter.ts
Lodin: Is it only about `badge-counter`, or about all the styles? I see that `styles` approach is widely used across examples, e.g. in `app-layout`, in `confirm-dialog`, or in `upload`

frontend/demo/component/badge/badge-color.ts
Lodin: Well, this change is ok for me though the grid looks like a better solution because it requires only one container for both lines.

frontend/demo/component/badge/badge-interactive.ts
```diff
@@ -1,22 +1,112 @@
...
+    const { people } = await getPeople();
+    this.items = [...new Set(people.map(({ profession }) => profession))];
+  }
+
   render() {
     return html`
-      <!-- tag::snippet[] -->
-      <span theme="badge pill contrast">Badge</span>
-      <!-- end::snippet[] -->
+      <section class="container">
```
yuriy-fix, line 59: Nit: can we use `ordered-layout`s instead? 😅

frontend/demo/component/badge/badge-icons.ts
```diff
@@ -14,9 +27,34 @@ export class Example extends LitElement {
 
   render() {
     return html`
-      <!-- tag::snippet[] -->
-      <span theme="badge pill contrast">Badge</span>
-      <!-- end::snippet[] -->
+      <section class="container">
```
yuriy-fix, line 30: Same as with the comment for `colors` example.

frontend/demo/component/badge/badge-icons-only-table.ts
```diff
@@ -1,22 +1,103 @@
...
+  async firstUpdated() {
+    this.items = await getUserPermissions();
+  }
+
   render() {
     return html`
-      <!-- tag::snippet[] -->
-      <span theme="badge pill contrast">Badge</span>
-      <!-- end::snippet[] -->
+      <section class="container">
```
yuriy-fix, line 79: Nit: do we really need `container` over here?

frontend/demo/component/badge/badge-icons-only-table.ts
```diff
@@ -1,22 +1,103 @@
...
-import '@vaadin/vaadin-checkbox/vaadin-checkbox';
+import '@vaadin/vaadin-grid';
+import '@vaadin/vaadin-icons/vaadin-icons';
+import '@vaadin/vaadin-lumo-styles/icons';
+import type { GridColumnElement, GridItemModel } from '@vaadin/vaadin-grid';
 import { applyTheme } from 'Frontend/generated/theme';
+import { css, customElement, html, internalProperty, LitElement } from 'lit-element';
+import { render } from 'lit-html';
+
+const renderBoolean = (
```
yuriy-fix, line 13: Nit: probably we can also move these renders to align with the example above.

yuriy-fix (changes requested):
You are right, seems to be working properly now.

frontend/demo/component/badge/badge-icons-only-table.ts
Lodin: For this render, it would require repeating the function three times. That's why I left it moved out.

frontend/demo/component/badge/badge-interactive.ts
Lodin: Just checked: unfortunately, no.
image

It is necessary to have a tag list that wraps when the line is over but `vaadin-horizontal-layout` does not have that functionality. So it would be better to use a container with custom styles here.

frontend/demo/component/badge/badge-icons-only-table.ts
Lodin: I solved it by wrapping the `render()` internals with `// tag::snippet[]` and putting the `renderBoolean` function into it.

frontend/demo/component/badge/badge-icons-only-table.ts
```diff
@@ -1,22 +1,89 @@
+import { getUserPermissions } from 'Frontend/demo/domain/DataService'; // hidden-full-source-line
 import 'Frontend/demo/init'; // hidden-full-source-line
+import UserPermissions from 'Frontend/generated/com/vaadin/demo/domain/UserPermissions'; // hidden-full-source-line
 
-import { html, LitElement, customElement } from 'lit-element';
-import '@vaadin/vaadin-checkbox/vaadin-checkbox';
+import '@vaadin/vaadin-grid';
+import '@vaadin/vaadin-icons/vaadin-icons';
+import '@vaadin/vaadin-lumo-styles/icons';
+import '@vaadin/vaadin-ordered-layout';
```
yuriy-fix, line 8: Nit: Extra import.

yuriy-fix (commented):
One nit. 
If possible, let's rebase it on top of `latest`, so Vale check won't be broken.

frontend/demo/component/badge/badge-icons-only-table.ts
```diff
@@ -1,22 +1,88 @@
+import { getUserPermissions } from 'Frontend/demo/domain/DataService'; // hidden-full-source-line
 import 'Frontend/demo/init'; // hidden-full-source-line
+import UserPermissions from 'Frontend/generated/com/vaadin/demo/domain/UserPermissions'; // hidden-full-source-line
 
-import { html, LitElement, customElement } from 'lit-element';
-import '@vaadin/vaadin-checkbox/vaadin-checkbox';
+import '@vaadin/vaadin-grid';
+import '@vaadin/vaadin-icons/vaadin-icons';
+import '@vaadin/vaadin-lumo-styles/icons';
```
Haprog, line 7: Unnecessary import of Lumo icons as only Vaadin icons are used here.
```suggestion
import '@vaadin/vaadin-icons/vaadin-icons';
```

frontend/demo/component/badge/badge-highlight.ts
```diff
@@ -1,22 +1,99 @@
...
+            let theme: string;
+
+            switch (status) {
+              case ReportStatus.COMPLETED:
+                icon = 'lumo:checkmark';
+                title = 'Completed';
+                theme = 'success';
+                break;
+              case ReportStatus.IN_PROGRESS:
+                icon = 'lumo:cog';
```
Haprog, line 68: I see there doesn't seem to be such an icon in Vaadin icons or Lumo icons than what is in the design doc for this, but I think `vaadin:progressbar` might be more suitable for this. Also this file is missing the import for Lumo icons (and Vaadin icons if you update this to use one).

I don't have a strong opinion about the icon used here, but at least the missing import needs to be added.

yuriy-fix (approved):

Haprog:
Please wait a bit before merging. I have some comments on the way.

frontend/demo/component/badge/badge-interactive.ts
```diff
@@ -1,22 +1,108 @@
-import 'Frontend/demo/init'; // hidden-full-source-line
+import { getPeople } from 'Frontend/demo/domain/DataService'; // hidden-full-source-line
+import 'Frontend/demo/init';
 
-import { html, LitElement, customElement } from 'lit-element';
-import '@vaadin/vaadin-checkbox/vaadin-checkbox';
+import '@vaadin/vaadin-button';
+import '@vaadin/vaadin-combo-box';
+import '@vaadin/vaadin-icons/vaadin-icons';
+import '@vaadin/vaadin-lumo-styles/icons';
```
Haprog, line 7: Unnecessary import of Vaadin icons since only Lumo icons is used here.
```suggestion
import '@vaadin/vaadin-lumo-styles/icons';
```

frontend/demo/component/badge/badge-interactive.ts
```diff
@@ -1,22 +1,108 @@
...
+          label="Profession"
+          .items="${this.items}"
+          @change="${this.onChange}"
+        ></vaadin-combo-box>
+        <div class="professions">
+          ${repeat(
+            this.selectedProfessions,
+            (profession) => profession,
+            (profession) => html`
+              <span theme="badge contrast">
```
Haprog, line 68: Should be using pill style as in design doc.
```suggestion
              <span theme="badge pill contrast">
```

frontend/demo/component/badge/badge-counter.ts
```diff
@@ -15,7 +25,28 @@ export class Example extends LitElement {
...
+          <div class="item">
+            <span>Messages</span>
+            <span theme="badge contrast pill">2</span>
+          </div>
+        </vaadin-item>
+        <hr />
+        <vaadin-item>
+          <div class="item">
+            <span>Settings</span>
+            <span theme="badge error primary pill">1</span>
```
Haprog, line 46: This badge should have some `aria-label` indicating that there's some error or something needs attention as mentioned in the text that is shown under this example.

> Consider the last item in the example above. For sighted users, it’s clear something requires their attention. To make sure it’s obvious for people using screen readers, set the badge’s `aria-label` attribute.

Haprog (changes requested):
Also the theme names in the "Theme name" column (table in Color section) should be all lowercase. Now they're capitalized.

Haprog:
Btw at the top of the Badge page there is a note:
> Badge styles need to be imported separately. The code examples show how to import them for Flow (Java) and/or Fusion (TypeScript).

But it's actually not showed in any of the TS examples how to import them. Technically it comes through `applyTheme(this.shadowRoot);` for TS but it also requires adding the `lumoImports` in `theme.json` which is currently not documented anywhere (`lumoImports` property is not documented in custom theme docs either, or at least wasn't last I checked).

But imo this should not block this PR. After this is merged we should create another issue for documenting how to get badge imported as it requires writing some new content.

frontend/demo/component/badge/badge-highlight.ts
Lodin: I would probable leave lumo icon here because vaadin icon requires additional styles to make it smaller while lumo already has the proper size.

Haprog:
This part in `index.asciidoc`
```
page-links:
  - https://github.com/vaadin/vaadin-lumo-styles/blob/v{moduleNpmVersion:vaadin-lumo-styles}/badge.html[Source]
```
needs to be updated to:
```
page-links:
  - https://github.com/vaadin/web-components/blob/v{moduleNpmVersion:vaadin-lumo-styles}/packages/vaadin-lumo-styles/badge.js[Source]
```

Lodin:
@Haprog, I guess I can add a `theme.json` file to each example. That would follow the @jouni's recommendation in [Board PR](https://github.com/vaadin/docs/pull/310) that we should have any source used for this example creation displayed.

Haprog:
> I guess I can add a `theme.json` file to each example. That would follow the @jouni's recommendation in [Board PR](https://github.com/vaadin/docs/pull/310) that we should have any source used for the example creation displayed.

I think it would be a bit confusing without extra explanation since the preview snippet also doesn't show the path to where `theme.json` needs to be and it's not directly used in the example sources. There should be some new section about this imo as it's about theme configuration. But you can ignore it for this PR.

frontend/demo/component/badge/badge-highlight.ts
Haprog: Ok, makes sense. Just add the missing import then.

frontend/demo/component/badge/badge-counter.ts
Lodin: I have used a following approach: 
```html
<span aria-label="Settings require attention!" theme="badge error primary pill">
  1
</span>
```
Is it ok?

Haprog: Seems ok to me.

Haprog (approved):

frontend/demo/component/badge/badge-interactive.ts
```diff
@@ -1,22 +1,107 @@
...
-      <!-- tag::snippet[] -->
-      <span theme="badge pill contrast">Badge</span>
-      <!-- end::snippet[] -->
+      <vaadin-vertical-layout theme="spacing">
+        <vaadin-combo-box
+          label="Profession"
+          .items="${this.items}"
+          @change="${this.onChange}"
+        ></vaadin-combo-box>
+        <div class="professions">
```
tomivirkki, line 62: Unclosed tag.

Screenshot 2021-04-01 at 16 26 17

frontend/demo/component/badge/badge-icons-only-table.ts
```diff
@@ -1,22 +1,86 @@
...
     applyTheme(this.shadowRoot);
   }
 
+  async firstUpdated() {
+    this.items = await getUserPermissions();
+  }
+
   render() {
+    // tag::snippet[]
+    const renderBoolean = (
```
tomivirkki, line 28: Redefining the renderer function on every render will effectively make the bound `renderer` properties reassign on every render also. Move this function as a class function instead and assign as `.renderer="${this.renderBoolean}"`.

frontend/demo/component/badge/badge-interactive.ts
Haprog: I wonder if there's some tooling we could use to automatically lint markup in html tagged template literals, and fail the build (and not rely on manual checking of IDE warnings). I've used `htmlhint` before, but only with plain HTML files. I'm not sure if it can be made to work in template literals.

frontend/demo/component/badge/badge-icons-only-table.ts
Lodin: Yuriy suggested inlining helper functions, and I also aware that re-creating listeners is not a problem for `lit-html` (comparing to React, for example). Wouldn't it be more confusing for a user to have different approaches in the example?

PR #292, opened 2021-03-22
Flow tutorials

marcushellberg:
Move over and clean up 2 main Flow tutorials from tutorials section:

* Quick Start tutorial
* CRM tutorial (=In-depth course)

These same 2 should be available for latest as well, how is that handled in docs?


jouni:
> These same 2 should be available for latest as well

Based on this, I changed the target branch from `latest` → `v14` (a lot less conflicts now 😄 )

> how is that handled in docs?

You add the same content in the `latest` branch as well. There’s no way to reuse content across docs versions at the moment.

articles/flow/tutorials/in-depth-course/components-and-layouts.adoc
```diff
@@ -0,0 +1,139 @@
...
+Here are the most important Vaadin Flow concepts you need to know to get started. 
+
+=== What Is Vaadin Flow? 
+
+Vaadin is a Java framework for building web applications. 
+It has a component-based programming model that allows you to build user interfaces.
+
+=== Vaadin UI Components
+
+The Vaadin platform includes a <<{articles}/ds/components#,large library of UI components>> that you can use as the building blocks of your application.
```
jouni, line 23:
```suggestion
Vaadin includes a <<{articles}/ds/components#,large library of UI components>> that you can use as the building blocks of your application.
```

jouni:
@marcushellberg I’ll merge this as is. Would be great if you can work on some of the Vale issues, but I’ll let you decide if it’s worth it before copying the content over to the `latest` branch.

PR #300, opened 2021-03-23
Update Upload docs/demos based on review

Haprog:
Handle post-merge review issues from https://github.com/vaadin/docs/pull/248.

Refactor some of the file list and request mocking setup to have cleaner demos and hide the irrelevant parts from the code snippet previews.


frontend/demo/component/upload/upload-all-files.ts
```diff
@@ -24,44 +27,27 @@ export class Example extends LitElement {
...
 
   // tag::snippet[]
   render() {
     return html`
-      <vaadin-upload no-auto></vaadin-upload>
+      <vaadin-upload
+        no-auto
+        .__dummy1=${'' /* end::snippet[] */}
+        .files=${this.files /* hidden-full-source-line */}
+        .__dummy2=${'' /* tag::snippet[] */}
```
Haprog, line 39: This is a way I found to be able to hide attributes/bindings from both the unexpanded and expanded code snippet previews. `hidden-full-source-line` only hides it from the expanded preview. Wrapping with `end::snippet[]`/`tag::snippet[]` was needed but sadly there was no cleaner syntax that would work since no normal comments seem to be allowed within a start tag's `<` and `>`.

pom.xml
```diff
@@ -158,6 +158,11 @@
             <artifactId>mailapi</artifactId>
             <version>1.4.3</version>
         </dependency>
+        <dependency>
+            <groupId>org.vaadin.artur</groupId>
+            <artifactId>a-vaadin-helper</artifactId>
+            <version>1.5.0</version>
+        </dependency>
```
Haprog, line 165: This was added to be able to use the notification helper:
```ts
import { showErrorNotification } from '@vaadin/flow-frontend/a-notification';
```

This is also included in start.vaadin.com projects. Imo this type of helper should be part of the notification component in the future. However we could also just copy the implementation into this project if we want to avoid the dependency. The source is here: https://github.com/Artur-/a-vaadin-helper/blob/master/src/main/resources/META-INF/resources/frontend/a-notification.ts

frontend/demo/component/upload/upload-all-files.ts
jouni: It's possible to exclude nested tagged regions, as an alternative. For example `tags=outer;!inner"`

https://docs.asciidoctor.org/asciidoc/latest/directives/include-tagged-regions/#tag-filtering

jouni: But it seems like we should have a clear way of having code thst can be omitted from both collapsed and expanded code. The current hide-full-source-line could be hidden in the collapsed version as well, IMO.

Haprog: Yeah I was surprised that `hide-full-source-line` only hid it from one. I can't think of any case where it would be needed to show something in collapsed but hide it from expanded. If we can make `hide-full-source-line` hide it from both, then we could get rid of these two dummy lines here.

Haprog:
I can rebase this on latest after this has been reviewed and approved. Currently latest is one commit ahead (which only adds Avatar Java demos and does not conflict).

frontend/demo/component/upload/upload-all-files.ts
jouni: https://github.com/vaadin/docs-app/pull/253

Haprog: With @jouni's above PR to docs-app this can be simplified to:
```html
      <!-- hidden-source-line --><!-- prettier-ignore -->
      <vaadin-upload
        no-auto
        .files=${this.files /* hidden-source-line */}
      ></vaadin-upload>
```
Now the prettier-ignore line for the `<vaadin-upload>` tag is needed to prevent prettier from forcing the tag to a single line since is has less code now. At least this seems cleaner than two dummy property bindings.

Haprog: Actually would be nice if `prettier-ignore` lines would also be automatically hidden so the first line here could be just `<!-- prettier-ignore -->` instead of `<!-- hidden-source-line --><!-- prettier-ignore -->`

Haprog: At least I don't think we want to show any `prettier-ignore` comments in the snippets intentionally.

Haprog: This has now been simplified (in all the files where this syntax was used) since above docs-app PR was merged. This now requires the latest version of `docs-app`.

yuriy-fix (changes requested):

frontend/demo/component/upload/upload-all-files.ts
```diff
@@ -13,6 +14,8 @@ export class Example extends LitElement {
     // Apply custom theme (only supported if your app uses one)
     applyTheme(this.shadowRoot);
   }
+  @internalProperty() // hidden-full-source-line
```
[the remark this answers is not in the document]
Haprog: Good point. I guess `@internalProperty` would only be needed in case we would need to update the value (and have it trigger a render)? It's probably not needed for any of the private properties that are basically constant.

frontend/demo/component/upload/upload-demo-mock-files.ts
```diff
@@ -0,0 +1,53 @@
+import { createFakeUploadFiles } from './upload-demo-helpers';
```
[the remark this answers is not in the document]
Haprog: Would that be only used for the Java demo or can it somehow be used for TS demos too? If it's only for the Java demo I guess it should be implemented with the Java demos se it could be outside the scope of this PR.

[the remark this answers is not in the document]
Haprog: The class should probably be called `UploadFile` so it matches the TS interface from `vaadin-upload`.

frontend/demo/component/upload/upload-basic.ts
```diff
@@ -13,32 +13,17 @@ export class Example extends LitElement {
...
+  @internalProperty() // hidden-full-source-line
+  private files = createFakeFilesUploadBasic(); // hidden-full-source-line
 
   render() {
     return html`
       <!-- tag::snippet[] -->
-      <vaadin-upload></vaadin-upload>
+      <!-- hidden-source-line --><!-- prettier-ignore -->
+      <vaadin-upload
+        .files=${this.files /* hidden-source-line */}
```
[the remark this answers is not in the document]
Haprog: In my original version this wasn't a problem when I was setting `files` in `firstUpdated` with `querySelector`. Otherwise I like being able to just bind it here, but would be nice to have a syntax for trimming all the whitespace around a `hidden-source-line` so this could become just `<vaadin-upload></vaadin-upload>` in the preview.

I think I'll experiment with implementing such trimming in `docs-app`.

frontend/demo/component/upload/upload-all-files.ts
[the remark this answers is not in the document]
Haprog: Update: removed unnecessary use of `@internalProperty`.

pom.xml
Haprog: Update: removed the dependency and copied the implementation to a local file `demo/notification-helper.ts` (made minor changes to satisfy prettier and fixed some TS types in it).

frontend/demo/component/upload/upload-basic.ts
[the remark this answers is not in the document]
Haprog: Here's a PR to docs-app that can fix this: https://github.com/vaadin/docs-app/pull/257.

If that gets merged, we could use `hidden-source-line-trim` here and the result in the preview would be a clean `<vaadin-upload></vaadin-upload>`.

frontend/demo/component/upload/upload-all-files.ts
```diff
@@ -24,44 +25,26 @@ export class Example extends LitElement {
...
   }
 
   // tag::snippet[]
   render() {
     return html`
-      <vaadin-upload no-auto></vaadin-upload>
+      <!-- hidden-source-line --><!-- prettier-ignore -->
+      <vaadin-upload
+        no-auto
+        .files=${this.files /* hidden-source-line */}
```
tomivirkki, line 36: You probably don't need the `files` property in the class if you use:
`.files=${createFakeFilesUploadAllFiles()}`

Haprog: Good idea. Will update these.

Haprog: This change also makes the line just long enough that we don't need the `prettier-ignore` line either.

frontend/demo/component/upload/upload-basic.ts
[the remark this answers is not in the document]
Haprog: Update: Now this is using the new feature `hidden-source-line-trim` and the result in the preview is a clean `<vaadin-upload></vaadin-upload>`.

frontend/demo/component/upload/upload-demo-mock-files.ts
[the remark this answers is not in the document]
Haprog: @tomivirkki @yuriy-fix Should I do something about this in the scope of this PR?

frontend/demo/component/upload/upload-all-files.ts
Haprog: Updated.

frontend/demo/component/upload/upload-demo-mock-files.ts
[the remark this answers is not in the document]
tomivirkki: If this is a big change, it could be revisited when the Java-examples are created?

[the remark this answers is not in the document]
Haprog: I'm not sure what exactly this would require or what @yuriy-fix had in mind.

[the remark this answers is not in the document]
yuriy-fix: I guess we can implement it later when working on Java examples. Not blocking this PR.

yuriy-fix (approved):

PR #302, opened 2021-03-23
Update the quick start tutorial so all code samples can be copy-pastes without syntax errors.

gilberto-torrezan:
The problems were detected when running tests with Studio.


jouni (approved):

jouni:
The callouts don't work as expected, and I’m not sure we should merge this PR as it doesn't fix the original issue.

jouni:
It seems they are all fixable. I’ll push an update soon.

jouni:
The two remaining unresolved callout problems are likely because of this issue: https://github.com/vaadin/docs-app/pull/231

They work when testing locally in development mode, but fail in the production build after React hydration.

PR #310, opened 2021-03-25
Add Board component article + TS examples

Lodin:


articles/ds/components/board/index.asciidoc
```diff
@@ -17,27 +17,122 @@ include::{root}/articles/_commercial-banner.asciidoc[]
...
+|===
+| Breakpoint | Viewport width | Max number of columns
+
+| large
+| ≥ 960 pixels
+| 4
+
+| medium
+| 600 - 959 pixels
+| 2 or 3*
```
jouni, line 111: Use a footnote instead of the asterisk: https://docs.asciidoctor.org/asciidoc/latest/macros/footnote/

frontend/demo/component/board/board-breakpoints.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        vaadin-board-row.medium > div {
+          font-size: var(--lumo-font-size-m);
+          padding: var(--lumo-space-m);
+        }
+
+        vaadin-board-row.small > div {
+          font-size: var(--lumo-font-size-l);
+          padding: var(--lumo-space-s);
+        }
```
tomivirkki, line 36: These seem relevant to the example and should show in the initial code snippet (without having to expand to show the example full source)

frontend/demo/component/board/board-breakpoints.ts
```diff
@@ -0,0 +1,64 @@
...
+      <vaadin-split-layout>
+        <vaadin-board style="width: 80%">
+          <vaadin-board-row>
+            <div class="cell">Cell 1</div>
+            <div class="cell">Cell 2</div>
+            <div class="cell">Cell 3</div>
+            <div class="cell">Cell 4</div>
+          </vaadin-board-row>
+        </vaadin-board>
+        <board-resize-info style="20%"></board-resize-info>
```
tomivirkki, line 59: The `style="20%"` isn't valid

articles/ds/components/board/index.asciidoc
```diff
@@ -17,27 +17,205 @@ include::{root}/articles/_commercial-banner.asciidoc[opts=optional]
...
+
+[source,typescript]
+----
+include::{root}/frontend/demo/component/board/shared-styles/board-border.ts[group=TypeScript]
+----
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/board/BoardBasic.java[render,tags=snippet,indent=0,group=Java]
+----
```
tomivirkki, line 51: All the Java examples should be commented out for now

frontend/demo/component/board/board-breakpoints.ts
Lodin: Oh, that's the typo, thanks

frontend/demo/component/board/board-breakpoints.ts
Lodin: Hmm, is there any way in the asciidoc to put two parts of the code into the initial snippet? Or should I put the whole class into the snippet?

tomivirkki: There are examples of this in other demos

Lodin:
I have a concern that the last example that shows the ability to style elements basing on the current breakpoint is not very demonstrative. The font and padding change is not obvious and might be missed by the user. Probably, I need to make it more impactful by using smaller/bigger font/padding? Or change another style, like cell color?

frontend/demo/component/board/board-basic.ts
```diff
@@ -1,50 +1,72 @@
...
+          <board-card type="+">
+            Conversion rate
+            <span slot="current">18%</span>
+            <span slot="difference">3.9%</span>
+          </board-card>
+          <board-card type="±">
+            Custom metric
+            <span slot="current">-123.45</span>
+            <span slot="difference">0.0%</span>
+          </board-card>
```
tomivirkki, line 63: Using slots for "current" and "differnece" adds a lot of verbosity to the examples. If you used properties instead, this would look as:
```html
<board-card current="745" change="33.7" text="Current users"></board-card>
<board-card current="54.6" change="112.45" text="View events"></board-card>
<board-card current="18" change="3.9" text="Custom metric"></board-card>
<board-card current="-123.45" change="0" text="Custom metric"></board-card>
```

frontend/demo/component/board/utils/board-card.ts
```diff
@@ -0,0 +1,65 @@
...
+
+      .icon iron-icon {
+        --iron-icon-width: var(--lumo-font-size-xs);
+        --iron-icon-height: var(--lumo-font-size-xs);
+      }
+    `;
+  }
+
+  @property()
+  public type = '+';
```
tomivirkki, line 31: The `type` property is redundant information. If the Web Component was changed to use properties (as suggested above) instead of slots, you can determine the type from the sign of the `change` property.

frontend/demo/component/board/board-breakpoints.ts
```diff
@@ -0,0 +1,64 @@
...
+          padding: var(--lumo-space-s);
+        }
+      `,
+    ];
+  }
+
+  constructor() {
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot!);
```
tomivirkki, line 44: For some reason, some examples have added non-null assertions.

Screenshot 2021-04-01 at 15 16 21

tomivirkki (changes requested):

frontend/demo/component/board/board-basic.ts
tomivirkki: "board-card" isn't a very descriptive name for such a custom element. It should be something more specific to the domain where "current", "change" and "text" are relevant.

frontend/demo/component/board/utils/board-chart.ts
```diff
@@ -0,0 +1,67 @@
...
+      .title {
+        font-size: var(--lumo-font-size-l);
+        font-weight: 700;
+        margin-block-end: var(--lumo-space-m);
+      }
+    `;
+  }
+
+  @internalProperty()
+  private events: readonly ViewEvent[] = [];
```
tomivirkki, line 42: What is the readonly modifier here for? I haven't seen it used in other examples.

jouni:
> I have a concern that the last example that shows the ability to style elements basing on the current breakpoint is not very demonstrative. The font and padding change is not obvious and might be missed by the user. Probably, I need to make it more impactful by using smaller/bigger font/padding? Or change another style, like cell color?

Yeah, I think I didn’t notice the difference either when I was trying out the examples. Changing the background color would probably be the most obvious thing to notice.

frontend/demo/component/board/board-basic.ts
Lodin: > "board-card" isn't a very descriptive name for such a custom element

Would the `board-indicator` be better then?

> Using slots for "current" and "differnece" adds a lot of verbosity to the examples

Well, I would say that this is a case when slots shine (especially comparing to React). However, I agree, examples should have as less code as possible.

tomivirkki: > Would the board-indicator be better then?

The name shouldn't have "board" in it as the component isn't board-specific.

> Well, I would say that this is a case when slots shine

The primary purpose of slots is to support presenting any arbitrary (HTML) content in a specific place inside the host Web Component's shadow root. Using a slot would require you to take care that, for example, a slotted `<vaadin-text-area>` works nicely inside the slot of the card. Properties on the other hand are more suitable for passing single values of known type. In this case, you just want to pass one numeric value to the card component.

frontend/demo/component/board/board-basic.ts
```diff
@@ -1,50 +1,68 @@
...
+      <!-- tag::snippet[] -->
       <vaadin-board>
         <vaadin-board-row>
-          <div class="cell">Cell 1</div>
-          <div class="cell">Cell 2</div>
-          <div class="cell">Cell 3</div>
-          <div class="cell">Cell 4</div>
+          <example-indicator
+            current="745"
+            change="+33.7%"
```
tomivirkki, line 46: Why did you decide to use string type for `change` (instead of number)?

frontend/demo/component/board/utils/example-indicator.ts
```diff
@@ -0,0 +1,74 @@
...
+  @property()
+  public current = '0';
+
+  @property()
+  public change = 0;
+
+  constructor() {
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot!);
```
tomivirkki, line 41: https://github.com/vaadin/docs/pull/311#pullrequestreview-628734775

tomivirkki:
The PR adds controls for trying out the component responsiveness using a split layout but the appearance is very different from what we already have in form-layout.

Board:

Screenshot 2021-04-07 at 14 56 47

Form-layout:

Screenshot 2021-04-07 at 15 04 51

Shouldn't these be aligned somehow?

tomivirkki:
Bumped into some issues on mobile devices. Please make sure the examples look and feel as you'd expect on mobile devices also.

tomivirkki:
When working on the examples, please follow patterns already used in the existing examples. This also applies to the internal structure.

Screenshot 2021-04-07 at 15 40 49

vs

Screenshot 2021-04-07 at 15 38 45

frontend/demo/component/board/utils/example-indicator.ts
Lodin: Just checked: if I remove `applyTheme`, component loses the badge styling.

image

Haprog: Would be great if we could import just the needed styles (Lumo badge) here, but I think we don't yet have any very nice way to do that into LitElement since the Lumo styles are Polymer style modules.

Easiest way I can think of is extending the experimental [ThemableElement](https://github.com/vaadin/component-mixins/tree/master/packages/themable-element) and then using [registerStyles()](https://github.com/vaadin/vaadin-themable-mixin#styling-in-javascript) with the `include` option like:
```ts
import { ThemableElement } from '@vaadin/themable-element/themable-element.js';
import { registerStyles } from '@vaadin/vaadin-themable-mixin/register-styles.js';

export class ExampleIndicator extends ThemableElement {
  static is = 'example-indicator';
  /*...*/
}
registerStyles('example-indicator', undefined, {include: ['lumo-badge']});
```
I didn't try this but I think it should work. What do you think @tomivirkki, would it be ok to do this here instead of `applyTheme()`?

Imo main issue is that `ThemableElement` is not really documented or findable, but maybe this is ok for this kind of demo?

Haprog: Slightly off-topic but here's some recent discussion about ThemableElement in our Slack https://vaadin.slack.com/archives/C3TGRP4HY/p1617098815150800 (for reference)

Lodin: Hmm, the idea looks interesting but I don't think we should introduce undocumented approaches in documentation examples.

articles/ds/components/board/index.asciidoc
```diff
@@ -17,27 +17,185 @@ include::{root}/articles/_commercial-banner.asciidoc[opts=optional]
...
+| small
+| < 600 pixels
+| 1
+|===
+
+The breakpoints can be customized by overriding the CSS custom properties named `--vaadin-board-width-small` and `--vaadin-board-width-medium`.
+
+=== Breakpoint-Specific Styling
+
+You can apply different styles for each breakpoint, for example to change the font size and space between components.
```
tomivirkki, line 187: Related to https://github.com/vaadin/docs/pull/310#issuecomment-814860291

This looks better now, but the admonition (in the asciidoc) is still needed

Screenshot 2021-04-12 at 9 46 49

Lodin: Ah, I missed that, thanks!

frontend/demo/component/board/utils/example-indicator.ts
tomivirkki: > Just checked: if I remove applyTheme component loses the badge styling

Ah, good point. Then it's good as it is.

articles/ds/components/board/index.asciidoc
Lodin: Fixed

tomivirkki (approved):
Approved. It still has some style flaws on mobile though, like this one:

Screenshot 2021-04-12 at 15 29 30

Lodin:
Thanks, @tomivirkki, I forgot to update imports for the Column Span section. Fixed now.

PR #311, opened 2021-03-25
feat: Split Layout TS demos

Haprog:
Implement Split Layout TS demos. Visual dummy content is based on Rolf's [CodePen prototype](https://codepen.io/rsmeds/pen/rNWPyYG)


Haprog:
I'm not sure happy about how the dummy content looks here. It's based on Rolf's proto, but I'm not sure how it could be made better without spending a lot of time on it. Maybe some designer could tweak them to be nicer?

I think currently the dummy content is a bit distracting. Especially since it uses the same or similar colors as the splitter. Maybe it would be better if the dummy content used some other colors so it would be easier to spot the splitter UI itself? Or maybe just use a simpler placeholder with a bit of text in the middle on both content areas to keep it simple so it wouldn't look so cluttered?

I think it would be best if there was some more realistic content here.

Anyway I think this is still not too bad to publish. It could be improved later.

frontend/demo/component/splitlayout/split-layout-toggle.ts
```diff
@@ -11,14 +17,55 @@ export class Example extends LitElement {
...
   }
+
+  setSplitterPosition(sidebarWidthPercentage: number) {
+    if (!this.splitLayout) throw 'splitLayout not ready';
+    const sidebar = this.splitLayout.children[0] as HTMLElement;
+    const contentArea = this.splitLayout.children[1] as HTMLElement;
+    sidebar.style.flex = '';
+    contentArea.style.flex = '';
+    sidebar.style.width = sidebarWidthPercentage + '%';
+    contentArea.style.width = 100 - sidebarWidthPercentage + '%';
```
tomivirkki, line 60: Use data-binding for this:
`<div style="width: ${this.sidebarWidthPercentage}%">`

`<detail-content style="width: ${100 - this.sidebarWidthPercentage}%">`

Haprog: I'm actually surprised that this works. And it also makes the demo code much simpler since we don't need `firstUpdated`, `@query`, `setSplitterPosition()` and `updateSplitterPosition` anymore.

Since we need to reset the `flex` property (in case user has dragged the splitter manually) it's not very intuitive how this works but it just happens to work because binding to `style` attribute will automatically remove all inline styles that have been set imperatively outside the template when it gets rendered.

Haprog: Updated.

DiegoCardoso:
I agree that the `master-content` color on the `<th>` elements background might confuse with the splitter.

I would suggest tone it down a bit, by removing the background and making the `<tbody>` color a bit brighter (but darker than the splitter color):
image

Another suggestion would be to add some margin/padding on the `master-content` to separate it from the splitter.

Haprog:
Agreed with Jouni and Rolf in Slack that the visuals are good enough for now. So technical review and approval should be all that's needed for this PR now.

articles/ds/components/split-layout/index.asciidoc
```diff
@@ -25,19 +25,19 @@ In the meantime, you can view the previous examples for the web component and th
...
+--
 
-// [source,html]
-// ----
-// include::{root}/frontend/demo/component/splitlayout/split-layout-basic.ts[render,tags=snippet,indent=0,group=TypeScript]
-// ----
+[source,html]
+----
+include::{root}/frontend/demo/component/splitlayout/split-layout-basic.ts[render,tags=snippet,indent=0,group=TypeScript]
+----
```
Lodin, line 34: According to @jouni's recommendation in [Board PR](https://github.com/vaadin/docs/pull/310), we need to include all the code used to create an example, as tabs. So the `master-content` and `detail-content` should be added to each example they are used. 

```suggestion
[source,html]
----
include::{root}/frontend/demo/component/splitlayout/split-layout-basic.ts[render,tags=snippet,indent=0,group=TypeScript]
----

[source,typescript]
----
include::{root}/frontend/demo/component/splitlayout/master-content.ts[group=TypeScript]
----

[source,typescript]
----
include::{root}/frontend/demo/component/splitlayout/detail-content.ts[group=TypeScript]
----
```

Lodin (commented):
I'm going to take a closer look a bit later but I have one comment already.

articles/ds/components/split-layout/index.asciidoc
Haprog: Done.

frontend/demo/component/splitlayout/split-layout-toggle.ts
```diff
@@ -11,14 +16,37 @@ export class Example extends LitElement {
...
+        </div>
+        <detail-content style="width: ${100 - this.sidebarWidthPercentage}%"></detail-content>
       </vaadin-split-layout>
-      <!-- end::snippet[] -->
     `;
   }
+
+  toggleSidebar() {
+    this.sidebarCollapsed = !this.sidebarCollapsed;
+    this.sidebarWidthPercentage = this.sidebarCollapsed ? 13 : 40;
```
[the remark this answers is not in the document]
Lodin: Or since it is used two times in render, to the `const` variable in render's scope:

```js
const sidebarWidthPercentage = this.sidebarCollapsed ? 13 : 40;

return html`
  ...
  <div style="overflow: hidden; width: ${sidebarWidthPercentage}%">
    ...
  </div>
  <detail-content
    style="width: ${100 - sidebarWidthPercentage}%"
  ></detail-content>
  ...
`
```

frontend/demo/component/splitlayout/master-content.ts
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      table {
+        border-collapse: collapse;
+      }
+
+      th,
+      td {
+        border-bottom: 1px solid currentColor;
+        padding: 0.66rem 1rem;
```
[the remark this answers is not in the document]
Haprog: Good idea. I just copied these from Rolf's proto.

articles/ds/components/split-layout/index.asciidoc
jouni: It's not a hard rule, so use your own judgement case by case. But I think it's safer to show everything than only the main source file. But some files could be omitted without making the reader guess what an import might contain.

frontend/demo/component/splitlayout/split-layout-toggle.ts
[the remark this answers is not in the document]
Haprog: Updated.

frontend/demo/component/splitlayout/master-content.ts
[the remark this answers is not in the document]
Haprog: Updated both to use Lumo variables where it made sense.

Haprog:
Updated according to review suggestions and rebased onto latest.

Haprog:
Rebased onto latest again.

Lodin (approved):
LGTM

frontend/demo/component/splitlayout/split-layout-theme-minimal.ts
```diff
@@ -14,9 +16,9 @@ export class Example extends LitElement {
   render() {
     return html`
       <!-- tag::snippet[] -->
-      <vaadin-split-layout>
-        <div>First content element</div>
-        <div>Second content element</div>
+      <vaadin-split-layout style="max-height: 280px;" theme="minimal">
+        <master-content style="overflow: hidden;"></master-content>
```
tomivirkki, line 20: This inline `overflow: hidden` is probably unintentional? The master/detail-content custom elements seem to have overflow hidden by default.

frontend/demo/component/splitlayout/master-content.ts
```diff
@@ -0,0 +1,80 @@
...
+      th {
+        background: var(--lumo-contrast-5pct);
+      }
+    `;
+  }
+
+  constructor() {
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot);
```
tomivirkki, line 42: nit: Why would a Vaadin custom theme be applied inside the placeholder components?

frontend/demo/component/splitlayout/split-layout-theme-minimal.ts
Haprog: Yes, thanks.

Updated.

frontend/demo/component/splitlayout/master-content.ts
Haprog: Good observation. Updated to not use it in master-content and detail-content components.

Haprog:
For some reason the Netlify build now failed with:
```
12:16:07 PM: [ERROR] Command `/opt/buildhome/.nvm/versions/node/v12.18.0/bin/npx --yes --quiet pnpm --shamefully-hoist=true install` failed:
12:16:07 PM: Lockfile is up-to-date, resolution step is skipped ERROR  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up-to-date with package.json
12:16:07 PM: [ERROR] >>> Dependency ERROR. Check that all required dependencies are deployed in pnpm repositories.
12:16:07 PM: [INFO] ------------------------------------------------------------------------
12:16:07 PM: [INFO] BUILD FAILURE
12:16:07 PM: [INFO] ------------------------------------------------------------------------
```

I'm not sure what could have changed. I'm pretty sure previous build was fine and no dependencies were modified in the last push.

tomivirkki (approved):

PR #312, opened 2021-03-25
docs: update to 20.0.0.alpha5, fix TS errors

web-padawan:
When updating docs to Vaadin 20.0.0.alpha4 there was a following error:

```
[14:58:36][Step 2/2] [ERROR] [tsl] ERROR in /opt/agent/work/83e2cf1ab9bef258/frontend/demo/component/avatar/avatar-group-internationalisation.ts(50,11)
[14:58:36][Step 2/2] [ERROR]       TS2739: Type '{ anonymous: string; activeUsers: { one: string; many: string; }; }' is missing the following properties from type 'AvatarGroupI18n': joined, left
```

This is a PR to add missing properties. Please forgive me if I'm mistaken, as I used Google translate 🙈


jouni (approved):
Impressive translation skills 👍 😄

Haprog:
I guess this should update the platform version at the same time to make it compile.

Haprog:
`ERROR  Cannot install with "frozen-lockfile" because pnpm-lock.yaml is not up-to-date with package.json`

tomivirkki (approved):

frontend/themes/theme.js
```diff
@@ -1 +1 @@
-export { applyTheme } from './docs/docs.generated.js';
+export { applyTheme } from 'Frontend/generated/theme-docs.generated.js';
```
tomivirkki, line 1: This needs the change also https://github.com/vaadin/docs/blob/latest/webpack.dspublisher.js#L94

web-padawan: This file no longer exists, and the `generated` folder is already excluded. I will remove that line altogether.

tomivirkki: It doesn't exist in the version control but it's generated by Vaadin on startup (not under `generated` but under the default project theme). Apparently the generated name was changed in the V20 alpha period from `[name].generated.js` to `theme-[name].generated.js`.

tomivirkki: > not under generated but under the default project theme

I didn't actually check if this is true with the new version. If this has also changed (the project theme -generated.js file also appears under `generated`) then the line could be removed, as you suggested.

PR #316, opened 2021-03-26
AppLayout: Java examples

anezthes:
**Note**: We shouldn't use anchors inside tabs. The icon-only examples _should_ have aria-labels. But, I don't think we should spend considerable amounts of time on solutions we know aren't great/optimal.

We really need that nav component.

Anywho...


magi42:
Is this still under work? Is it a draft or ask for review?

anezthes:
> Is this still under work? Is it a draft or ask for review?

Review.

jouni:
Tested the examples locally. If anyone has comments regarding the implementations, add comments here or open a new PR to address them.

PR #318, opened 2021-03-29
fix: remove dspublisher generated files

tomivirkki:
Fixes https://github.com/vaadin/docs-app/issues/250

This PR adds a docker-compose file whose sole purpose is to remove the `docs/target` dir before running the license check. This is needed on Linux hosts, where the generated target files get owned by the root user and so can't be changed by the current user.

Also, any existing build artifacts under `dspublisher/out` get cleaned up on build.


jouni (approved):
If I understood correctly, this helps with restarting DS Publisher. I’m not much of an expert with Docker, so I can’t say much about the commands either. Looks alright.

PR #329, opened 2021-04-01
chore: clean up the default custom header

tomivirkki:
Fixes https://github.com/vaadin/docs-app/issues/247


scripts/dspublisher-init-templates/dsp-global.css
```diff
@@ -1,14 +1,20 @@
 header h3 a {
   display: flex;
   align-items: center;
-  font-size: 25px;
+  font-size: var(--docs-font-size-h1);
```
jouni, line 4: H1 size would be huge 😅 

```suggestion
  font-size: var(--docs-font-size-h4);
```

scripts/dspublisher-init-templates/dsp-global.css
```diff
@@ -1,14 +1,20 @@
...
-  margin-right: 20px;
-}
+  margin-right: 1em;
+  background-image: var(--docs-header-logo);
```
jouni, line 17: IMO the custom property doesn’t add any value. It might be more confusing than helpful.

tomivirkki: Agreed, removed the custom property

jouni (approved):

PR #331, opened 2021-04-06
Add CE message list & input docs

pekam:
Resolves https://github.com/vaadin/collaboration-engine-internal/issues/554

### DIFF
--- a/articles/ce/collaboration-message-list.asciidoc
+++ b/articles/ce/collaboration-message-list.asciidoc
@@ -30,9 +30,9 @@ for more details on how the avatars are rendered, or how to load images from a b
 
 The topic id works in the same way as for other Collaboration Engine features.
 The data in Collaboration Engine is shared among those users who are connected to the same topic.
-In the case of these components, the topic can be considered as a chat room.
-In our example code, the topic id is hard-coded as "general", which means that every
-user in this view will see each others' messages.
+With these components, you can consider the topic as a chat room.
+In the example above, the topic id is hard-coded as "general", which means that every
+user in this view sees each others' messages.
 You can dynamically change the topic with the `setTopic` method,
 for example when the user makes a selection in the UI.
 
@@ -42,7 +42,7 @@ When the list component has `null` topic id, no messages are displayed, and the 
 message input will be automatically disabled.
 
 After constructing the components, the last required step is to place them somewhere in your view.
-Here's an example of placing the message list and input on top of each other in a layout with defined size:
+Here's an example of placing the message list and input on top of each other in a layout with a defined size:
 
 [source,java]
 ----
@@ -52,7 +52,7 @@ include::{root}/src/main/java/com/vaadin/demo/ce/MessageListDocumentation.java[t
 == Using Custom Input Components
 
 Instead of using the provided `CollaborationMessageInput` component, you can also
-configure your own components to submit messages with the `CollaborationMessageList::setSubmitter` method.
+configure your own component to submit messages. The component can be registered with the `CollaborationMessageList::setSubmitter` method.
 Here's an example of submitting messages with `TextField` and `Button` components:
 
 [source,java]
@@ -64,5 +64,5 @@ The `setSubmitter` method takes a callback that is run when a new topic connecti
 (topic id is set to a non-null value).
 This callback should be used to set up any listeners for submitting new messages to `CollaborationMessageList`
 through the provided `ActivationContext`.
-The callback should return another callback for cleaning things up when the topic connection is deactivated (topic id is changed).
+The callback should return another callback for clean-up when the topic connection is deactivated (topic id is changed).
 Note how the submit button is enabled only when there's an active topic connection.
--- a/articles/ce/collaboration-message-list.asciidoc
+++ b/articles/ce/collaboration-message-list.asciidoc
@@ -1,0 +1,68 @@
+---
+title: CollaborationMessageList
+order: 8
+layout: page
+---
+
+[[ce.collaboration-message-list]]
+= Real-Time Chat and Commenting Components
+
+You can add a real-time chat to your application by using the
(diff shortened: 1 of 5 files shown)


articles/ce/collaboration-message-list.asciidoc
```diff
@@ -0,0 +1,68 @@
...
+
+The `CollaborationMessageList` constructor takes two arguments: the information about
+the end user associated with this session, and a topic id.
+The user info is used to render the user name and avatar in the messages submitted by this user.
+Refer to the <<collaboration-avatar-group#ce.configuring-avatars, `CollaborationAvatarGroup` documentation>>
+for more details on how the avatars are rendered, or how to load images from a backend with the `setImageProvider` method.
+
+The topic id works in the same way as for other Collaboration Engine features.
+The data in Collaboration Engine is shared among those users who are connected to the same topic.
+In the case of these components, the topic can be considered as a chat room.
```
Peppe, line 33:
```suggestion
With these components, you can consider the topic as a chat room.
```

articles/ce/collaboration-message-list.asciidoc
```diff
@@ -0,0 +1,68 @@
...
+In our example code, the topic id is hard-coded as "general", which means that every
```
Peppe, line 34:
```suggestion
In the example above, the topic id is hard-coded as "general", which means that every
```

articles/ce/collaboration-message-list.asciidoc
```diff
@@ -0,0 +1,68 @@
...
+user in this view will see each others' messages.
```
Peppe, line 35:
```suggestion
user in this view sees each others' messages.
```

articles/ce/collaboration-message-list.asciidoc
```diff
@@ -0,0 +1,68 @@
...
+You can dynamically change the topic with the `setTopic` method,
+for example when the user makes a selection in the UI.
+
+`CollaborationMessageInput` takes the message list component as its constructor argument.
+It hooks up with the list, to submit messages to its current topic.
+When the list component has `null` topic id, no messages are displayed, and the connected
+message input will be automatically disabled.
+
+After constructing the components, the last required step is to place them somewhere in your view.
+Here's an example of placing the message list and input on top of each other in a layout with defined size:
```
Peppe, line 45:
```suggestion
Here's an example of placing the message list and input on top of each other in a layout with a defined size:
```

articles/ce/collaboration-message-list.asciidoc
```diff
@@ -0,0 +1,68 @@
...
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/ce/MessageListDocumentation.java[tags=message-list-layout,indent=0]
+----
+
+== Using Custom Input Components
+
+Instead of using the provided `CollaborationMessageInput` component, you can also
+configure your own components to submit messages with the `CollaborationMessageList::setSubmitter` method.
```
Peppe, line 55:
```suggestion
Instead of using the provided `CollaborationMessageInput` component, you can also
configure your own component to submit messages. The component can be registered with the `CollaborationMessageList::setSubmitter` method.
```

Peppe: The sentence felt like it has three parts "Instead of ... you can also ... with the ...". It felt like it needed a break after the "instead of .. do this."

articles/ce/collaboration-message-list.asciidoc
```diff
@@ -0,0 +1,68 @@
...
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/ce/MessageListDocumentation.java[tags=message-list-submitter,indent=0]
+----
+
+The `setSubmitter` method takes a callback that is run when a new topic connection is established
+(topic id is set to a non-null value).
+This callback should be used to set up any listeners for submitting new messages to `CollaborationMessageList`
+through the provided `ActivationContext`.
+The callback should return another callback for cleaning things up when the topic connection is deactivated (topic id is changed).
```
Peppe, line 67:
```suggestion
The callback should return another callback for clean-up when the topic connection is deactivated (topic id is changed).
```

Peppe (changes requested):

articles/ce/overview.asciidoc
```diff
@@ -28,16 +28,19 @@ Avatars::
 End users see who else is looking at the same data as they are.
 The avatars are automatically updated when users join or leave.
 
+Real-time discussion::
+End users can submit comments and chat messages in real time.
```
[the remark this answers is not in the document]
pekam: Good point!

I'd still like to mention "messages" or "comments". Otherwise it might not be obvious what "discussion" means (could be voice call).

Also, about "real-time". I've been following this convention:
- Use dash when it's a prefix for something, e.g. "real-time discussion"
- Use space otherwise, e.g. "...in real time."

I think this was recommended somewhere when I googled it.

So how about:
```suggestion
End users can send messages with each other within the application, either in real time or asynchronously.
```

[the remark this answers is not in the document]
Peppe: My grammar fixer always want to have real-time and end-user (also in sentence) so I tend to put those, but your comment makes sense. Let's go with your suggestion.

pom.xml
```diff
@@ -158,6 +158,11 @@
             <artifactId>mailapi</artifactId>
             <version>1.4.3</version>
         </dependency>
+        <dependency>
+            <groupId>com.vaadin</groupId>
+            <artifactId>collaboration-engine</artifactId>
+            <version>3.1.0.alpha5</version>
+        </dependency>
```
pekam, line 165: When this PR is merged, I'll create another PR to remove this temporary version bump right away, so that we remember to remove that when we update to the next platform alpha that includes this CE alpha5.

Peppe (approved):

PR #334, opened 2021-04-06
ConfirmDialog: Java examples

anezthes:


sosa-vaadin (commented):
I checked the Netlify deploy and noticed that with **Java samples**, the code is showing but the working examples **are not being rendered**.

jouni:
Hes, that's expected and a known limitation, unfortunately. We don't have preview builds for rendered Java examples. You need to test them locally.

src/main/java/com/vaadin/demo/component/confirmdialog/ConfirmDialogConfirmButton.java
```diff
@@ -0,0 +1,52 @@
...
+    layout.setJustifyContentMode(FlexComponent.JustifyContentMode.CENTER);
+
+    // tag::snippet[]
+    status = new Span();
+    status.setVisible(false);
+
+    ConfirmDialog dialog = new ConfirmDialog();
+    dialog.setHeader("Export failed");
+    dialog.setText(
+      new Html("An error occurred while exporting <b>Report Q4</b>. Please try again. If the problem persists, please contact <a href=\"mailto:support@company.com\">support@company.com</a>.")
```
sosa-vaadin, line 30: An exception is thrown at this line because of having both the `<a>` and `<b>` nodes at the parent level. Wrap the entire content in a `<span>` to fix this. Replacing the `Html` object with a `Span` object won't work because these tags will be escaped.
```suggestion
      new Html("<span>An error occurred while exporting <b>Report Q4</b>. Please try again. If the problem persists, please contact <a href=\"mailto:support@company.com\">support@company.com</a>.</span>")
```

sosa-vaadin (changes requested):
Fix the exception thrown at the "confirm dialog confirm button" example and you're good to go.

src/main/java/com/vaadin/demo/component/confirmdialog/ConfirmDialogConfirmButton.java
anezthes: Ah, good catch! Changed to paragraph.

sosa-vaadin (approved):
Good to go.

PR #363, opened 2021-04-13
Add TS examples for cookie-consent component

Lodin:


articles/ds/components/cookie-consent/index.asciidoc
```diff
@@ -28,19 +28,107 @@ The usage examples for this component are currently in progress. In the meantime
 - https://vaadin.com/components/vaadin-cookie-consent/java-examples[Java Component Examples]
 ====
 
+.Scaled down examples
+[IMPORTANT]
+The examples on this page are scaled down so that their viewport-size dependant behavior can be demonstrated.
+Some examples change their behavior also based on your browser viewport size.
```
jouni, line 34: If we are going to reuse this content, we should extract it to a shared file and include it from there.

jouni: Regarding the CSS, not all of these are used on this page (no `.mobile` or `.desktop` classes used).

For the scale, 0.8 would be enough for Cookie Consent. If we extract the CSS into a shared include, we can parametrize the scale value using an Asciidoc attribute.

articles/ds/components/cookie-consent/index.asciidoc
```diff
@@ -28,19 +28,107 @@ The usage examples for this component are currently in progress. In the meantime
...
 ----
-include::{root}/frontend/demo/component/cookieconsent/cookie-consent-basic.ts[render,tags=snippet,indent=0,group=TypeScript]
+include::{root}/frontend/demo/component/cookieconsent/cookie-consent-basic.ts[render,frame,tags=snippet,indent=0,group=TypeScript]
 ----
 
-[source,java]
+[source,typescript]
 ----
-include::{root}/src/main/java/com/vaadin/demo/component/cookieconsent/CookieConsentBasic.java[render,tags=snippet,indent=0,group=Java]
+include::{root}/frontend/demo/component/cookieconsent/example-base.ts[group=TypeScript]
```
jouni, line 124: Should we show this part of the source code? I assume it’s not needed in a regular Vaadin application, and therefore not necessary for developers to be able to copy paste into their project. It could end up confusing readers more than being helpful, I think.

Though, the example does use the mixin, but maybe we could add a code comment that says you can safely delete the import and the `extends ExampleMixin(...)` and just use `extends LitElement`. Would be great if there was some other way to bring in the shared functionality, so that we could hide it completely from the source code examples.

frontend/demo/component/cookieconsent/cookie-consent-theming.ts
```diff
@@ -3,12 +3,46 @@ import '@vaadin/flow-frontend/contextMenuConnector.js'; // hidden-full-source-li
...
+        border: none;
+        border-radius: var(--lumo-border-radius);
+        color: var(--lumo-primary-contrast-color);
+        font-size: var(--lumo-font-size-m);
+        line-height: var(--lumo-size-m);
+        min-width: 0;
+        padding: 0 calc(var(--lumo-size-m) / 3 + var(--lumo-border-radius) / 2);
+      }
+    `;
+    document.body.appendChild(style);
```
jouni, line 38: These should rather go into the [docs theme](https://github.com/vaadin/docs/tree/latest/frontend/themes/docs) `styles.css` or `document.css`.

frontend/demo/component/cookieconsent/cookie-consent-theming.ts
```diff
@@ -3,12 +3,46 @@ import '@vaadin/flow-frontend/contextMenuConnector.js'; // hidden-full-source-li
...
-// tag::snippet[]
 @customElement('cookie-consent-theming')
-export class Example extends LitElement {
+export class Example extends ExampleMixin(LitElement) {
+  // tag::snippet[]
+  connectedCallback() {
+    const style = document.createElement('style');
+    style.innerHTML = `
+      div.cc-window.cc-theme-block {
+        background-color: var(--lumo-base-color);
```
yuriy-fix, line 15: Probably need to be confirmed with @anezthes, the styles specified by him in the doc don't look the same as "primary" notification. Compared to the `Primary` example on [CDN](https://cdn.vaadin.com/vaadin-notification/1.6.1/demo/#notification-theme-variants-demos).

yuriy-fix (changes requested):

articles/ds/components/cookie-consent/index.asciidoc
yuriy-fix: I am not sure that we actually need that functional part from `ExampleMixin`. Would it be enough just to show the component itself. So there will be no need in adding extra logic with mixin? 🤔

frontend/demo/component/cookieconsent/cookie-consent-theming.ts
tomivirkki: ...preferably as an `@import` of a separate file specific to this example

articles/ds/components/cookie-consent/index.asciidoc
tomivirkki: Using mixins or a custom superclass with an example generally makes it less copy-pasteable so we've intentionally tried avoiding it.

frontend/demo/component/cookieconsent/cookie-consent-theming.ts
jouni: Yeah, true. I can update the Notification examples to do that as well (also noticed that those examples aren’t even using `applyTheme` for some reason).

articles/ds/components/cookie-consent/index.asciidoc
jouni: > Would it be enough just to show the component itself. So there will be no need in adding extra logic with mixin? 🤔

My assumption is that it’s needed so that the popup shows even if you’ve dismissed it before.

articles/ds/components/cookie-consent/index.asciidoc
```diff
@@ -28,19 +28,107 @@ The usage examples for this component are currently in progress. In the meantime
...
+  min-height: 0 !important;
+  border: 1px solid var(--docs-divider-color-2);
+  border-radius: 0;
+  margin: var(--docs-space-s);
+  margin-bottom: var(--docs-space-xs);
+}
+
+.example {
+  /* Just enough to keep "desktop" mode when docs are at max-width */
+  --scale: 0.58;
```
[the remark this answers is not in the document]
Lodin: If it is not scaled, it take a mobile shape, unfortunately
image

frontend/demo/component/cookieconsent/cookie-consent-theming.ts
Lodin: Hmm, should I add this file to examples then?

articles/ds/components/cookie-consent/index.asciidoc
Lodin: Removed. I made a shared init part instead

articles/ds/components/cookie-consent/index.asciidoc
Lodin: I applied scale 0.8, it fits perfectly. However, I'm not sure I fully understand how I can make a shared include that parametrizes with an Asciidoc attribute

frontend/demo/component/cookieconsent/cookie-consent-theming.ts
jouni: Yes, then that external style sheet should be included as a part of the examples source includes.

articles/ds/components/cookie-consent/index.asciidoc
jouni: I’ll try to do that once I finish updating/fixing the Notification examples.

frontend/demo/component/cookieconsent/cookie-consent-theming.ts
jouni: Before you do it, I suggest you update from latest (or rebase), as I modified the styles.css file a short while ago.

articles/ds/components/cookie-consent/index.asciidoc
jouni: I pushed an update that extracts the shared parts and makes the example scale an attribute that can be customized.

articles/ds/components/cookie-consent/index.asciidoc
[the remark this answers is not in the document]
jouni: Small amount of scaling is okay. We are using 0.8 now.

jouni (approved):
Good enough for me :+1:

yuriy-fix (approved):

PR #367, opened 2021-04-14
chore: add ts-lit-plugin to tsconfig

web-padawan:
WIP, do not merge. 

@vursen as you mentioned that you would like to try this, here is a PR.


jouni:
@web-padawan, can we delete the branch?

web-padawan:
Thanks, deleted. I could not get this working (no errors in webpack build). We could use `lit-analyzer` CLI instead though.

PR #371, opened 2021-04-15
update login-view to use redirectUrl from LoginResult

haijian-vaadin:
Update the doc based on the change in https://github.com/vaadin/flow/pull/10663, would require a new platform version, so mark it as a draft for now


haijian-vaadin:
Close this one in favor of https://github.com/vaadin/docs/pull/406

PR #372, opened 2021-04-15
feat: Add RTE docs and examples

yuriy-fix:


frontend/demo/component/richtexteditor/rich-text-editor-min-max-height.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  @internalProperty()
+  private richText = templates.richTextDelta;
+
+  render() {
+    return html`
+      <!-- tag::snippet[] -->
+      <vaadin-rich-text-editor
+        style="min-height:200px; max-height: 400px;"
+        value=${this.richText}
```
tomivirkki, line 25: Please avoid binding values via attributes

Screenshot 2021-04-16 at 16 37 14

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
```diff
@@ -0,0 +1,53 @@
...
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot);
+  }
+
+  @internalProperty()
+  htmlValue = '<p>Html value of the editor</p>';
+
+  @query('vaadin-rich-text-editor')
+  richTextEditor?: RichTextEditorElement;
```
tomivirkki, line 26: These are probably unintentionally public

articles/ds/components/rich-text-editor/index.asciidoc
```diff
@@ -220,7 +220,7 @@ include::../../../../frontend/demo/component/richtexteditor/rich-text-editor-the
 
 === No Border
 
-Apply the `no-border` theme variant to remove Rich Text Editor’s border, e.g. when the component is wrapped in a container with its own borders.
+Apply the `no-border` theme variant to remove Rich Text Editor’s border, for example when the component is wrapped in a container with its own borders.
```
jouni, line 223:
```suggestion
Apply the `no-border` theme variant to remove Rich Text Editor’s border, for example, when the component is wrapped in a container with its own borders.
```

frontend/demo/component/richtexteditor/rich-text-editor-basic.ts
```diff
@@ -22,7 +22,7 @@ export class Example extends LitElement {
       <!-- tag::snippet[] -->
       <vaadin-rich-text-editor
         style="height: 400px;"
-        value=${this.richText}
+        .value=${this.richText}
```
jouni, line 25: Should be fixed for all attribute values.

```suggestion
        .value="${this.richText}"
```

yuriy-fix: Done

frontend/demo/component/richtexteditor/rich-text-editor-min-max-height.ts
yuriy-fix: Updated.

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
yuriy-fix: Updated.

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
```diff
@@ -0,0 +1,53 @@
...
+          (this.htmlValue = e.detail.value)}"
+      ></vaadin-rich-text-editor>
+
+      <vaadin-text-area
+        @value-changed="${(e: TextFieldValueChanged) => this.setHtmlValue(e.detail.value)}"
+        placeholder="Type html string here to set it as value to the Rich Text Editor above..."
+        style="width: 100%;"
+      ></vaadin-text-area>
+
+      <div><b>Html value of the editor:</b> ${this.htmlValue}</div>
```
tomivirkki, line 43: [The design/UX of this example needs some tuning](https://deploy-preview-372--vaadin-docs-preview.netlify.app/ds/components/rich-text-editor#value-format). Right now, the flow of data is a bit confusing:
- Changes to the RTE are reflected to the div, but not to the text-area
- Changes to the text-area are reflected to the RTE and the div

I'd suggest dropping the div completely, since you already have a container for the HTML-value (the text-area) and then reflect the RTE changes to the TA and vice versa.

To avoid changing the value of the editor you're currently on, the flow of data could be one-way (from the editor you're currently on to the other editor), or alternatively, applied only on change event (instead of the eager value-changed). The second approach is probably nicer to implement for the Java example. @anezthes thoughts?

yuriy-fix: Syncing the `htmlValue` on `change` now. Aligned both the rte and the textarea so there won't be differences in setting the value.

articles/ds/components/rich-text-editor/index.asciidoc
```diff
@@ -8,36 +8,424 @@ page-links:
...
+  }
+
+  .code-block {
+    line-height: calc(var(--docs-line-height-s) * 2.1);
+    font-size: calc(var(--docs-font-size-m) * 1.2);
+    padding-left: calc(var(--docs-space-s) * 1.7);
+  }
+</style>
+++++
+
```
jouni, line 124: Move this block as the last thing in the document. No markup should come before the first heading, it breaks the page layout.

yuriy-fix: Done

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
```diff
@@ -0,0 +1,55 @@
...
+    super();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(this.shadowRoot);
+  }
+
+  @internalProperty()
+  private htmlValue = '';
+
+  @query('vaadin-rich-text-editor')
+  private richTextEditor?: RichTextEditorElement;
```
sosa-vaadin, line 23:
```suggestion
  private richTextEditor!: RichTextEditorElement;
```
Since you're using `@query()` decorator to assign a value to this property, use the definite assignment assertion operator, `!` instead of marking it as optional with `?`. That way, you won't have to use the optional operator any time you want to reference this variable.

sosa-vaadin (approved):
I left a suggestion about a property in one of the examples but it's not really that important and can be left out.

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
yuriy-fix: Done.

PR #396, opened 2021-04-22
fix: remove 'incomplete' icon for button and form-layout

vlukashov:
The 'incomplete' icon appears next to a component in the navigation menu for those components that have neither Java nor TS examples in the new docs site. This is no longer the case for button and form-layout, so let's remove the icon for them.


jouni (approved):
I already removed it from Button moments ago (didn’t notice this PR before).

PR #398, opened 2021-04-23
fix: clarify more on document.css usages in theming embedded components

taefi:
fixes: https://github.com/vaadin/flow/issues/10132


articles/ds/customization/custom-theme.asciidoc
```diff
@@ -155,7 +155,7 @@ These assets can be used in the theme’s style sheets through URIs relative to
 
 To ensure that certain styles are always applied to the document root rather than the shadow root of an embedded Flow application or component, they can be placed in a special style sheet in the theme folder root called [filename]#document.css#.
 This is mainly needed for `@font-face` declarations that are not supported inside web component shadow DOM, and only when the theme is going to be used with <<{articles}/flow/integrations/embedding/overview#, embedded Flow applications or components>> inside another application or used with Design System Publisher.
-Another example is when the theme should be applied to an embedded Flow application or component which is going to be shown in an overlay inside another applications, and since the overlay cannot access the styles from web component's shadow DOM, those styles must be added in a [filename]#document.css#, too.
+Another example is when the theme should be applied to an embedded Flow application or component which is going to be shown in an overlay inside another applications, and since the overlay cannot access the styles from web component's shadow DOM, styles also must be added in [filename]#document.css#, too.
```
caalador: Perhaps add a new line between each section as now it renders as a huge block of text and is hard to read when the parts talk a bit about different things.

caalador, line 158:
```suggestion
Another example is when the theme should be applied to an embedded Flow application or component which is shown in an overlay inside another application.
Since the overlay cannot access the styles from web component shadow DOM, the styles must also be added to [filename]#document.css#.
```

caalador (changes requested):

articles/ds/customization/custom-theme.asciidoc
taefi: Good Point! Thanks! Done.

caalador (approved):

articles/ds/customization/custom-theme.asciidoc
```diff
@@ -156,6 +156,9 @@ These assets can be used in the theme’s style sheets through URIs relative to
 To ensure that certain styles are always applied to the document root rather than the shadow root of an embedded Flow application or component, they can be placed in a special style sheet in the theme folder root called [filename]#document.css#.
 This is mainly needed for `@font-face` declarations that are not supported inside web component shadow DOM, and only when the theme is going to be used with <<{articles}/flow/integrations/embedding/overview#, embedded Flow applications or components>> inside another application or used with Design System Publisher.
 
+Another example is when the theme should be applied to an embedded Flow application or component which is shown in an overlay inside another application.
+Since the overlay cannot access the styles from web component shadow DOM, the styles must also be added to the [filename]#document.css#.
```
jouni, line 160:
```suggestion
Since the overlay cannot access the styles from web component shadow DOM, the styles must also be added to the [filename]#document.css#.
The recommended way to do this is to move such styles into a separate style sheet, and use `@import` to include them in both `styles.css` and `document.css`.
```

jouni: Nit: it would be good to have a cross reference to Flow app embedding docs.

articles/ds/customization/custom-theme.asciidoc
```diff
@@ -156,6 +156,9 @@ These assets can be used in the theme’s style sheets through URIs relative to
 To ensure that certain styles are always applied to the document root rather than the shadow root of an embedded Flow application or component, they can be placed in a special style sheet in the theme folder root called [filename]#document.css#.
 This is mainly needed for `@font-face` declarations that are not supported inside web component shadow DOM, and only when the theme is going to be used with <<{articles}/flow/integrations/embedding/overview#, embedded Flow applications or components>> inside another application or used with Design System Publisher.
 
+Another example is when the theme should be applied to an embedded Flow application or component which is shown in an overlay inside another application.
```
jouni, line 159:
```suggestion
Another example is when the theme should be applied to an <<{articles}/flow/integrations/embedding/overview#,embedded Flow application or component>> which is shown in an overlay inside another application.
```

PR #401, opened 2021-04-23
CP clarify document.css usages in theming embedded components

taefi:
* fix: clarify more on document.css usages in theming embedded components (#398)


caalador:
Is there some way to force the checks to run?

jouni:
I believe I’m the only one who can check the build statuses in Netlify and restart them if needed. We would need to upgrade our subscription if we want more users there (from Pro $19/month → Enterprise $99/month).

I just checked, and it looks like the build is passing. Not sure why the status is not reported here.
Screen Shot 2021-04-26 at 15 31 47

Also, as this PR touches Asciidoc files only, the preview build isn’t that critical. It’s very unlikely that it would break the build. I can merge this one when it’s ready.

caalador (approved):

PR #418, opened 2021-04-30
Externalize the example and fix the errors

### LINKED ISSUE #410: Example in  "Using a Field Web Component" is not working
The code in this page is not working:
https://vaadin.com/docs/latest/fusion/forms/appendix-web-component-field-strategy

In particular:

```typescript
import {AbstractModel, Binder, BinderConfiguration, FieldStrategy, ModelConstructor} from '@vaadin/form';
import { MyCcStrategy } from './my-cc-strategy';

export class MyBinder<T, M extends AbstractModel<T>> extends Binder<T, M> {
    constructor(context: any, model: ModelConstructor<T, M>, config?: BinderConfiguration<T>) {
        super(context, model, config);
    }

    getFieldStrategy(element: any): FieldStrategy {
        if (element.localName === 'my-cc-field') {
            return new MyCcStrategy(element);
        }
        return super.getFieldStrategy(element);
    }
}
```

And:
`onValueChanged(event) {` instead of `onValueChanged(event:CustomFieldValueChanged) {`

To avoid this, it would be better to create a separate example in Typescript that runs and embed this example.

jcgueriaud1:
Should fix #410

### DIFF
--- a/articles/fusion/forms/appendix-web-component-field-strategy.asciidoc
+++ b/articles/fusion/forms/appendix-web-component-field-strategy.asciidoc
@@ -15,39 +15,10 @@ This article explains how to configure the binder for using a custom text-field 
 Let us consider the following LitElement component for using in form binding:
 
 .Example web component field
+
 [source,typescript]
 ----
-import {LitElement, customElement, html, property} from 'lit-element';
-import '@vaadin/vaadin-custom-field';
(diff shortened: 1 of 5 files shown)


sosa-vaadin:
Closing this one since https://github.com/vaadin/docs/pull/421 already fixes the same issue and also changes the text content.

PR #421, opened 2021-05-02
fix: using custom web components in form fields

### LINKED ISSUE #410: Example in  "Using a Field Web Component" is not working

vlukashov:
- code snippets are updated so that they actually compile and work
 - the text content is rephrased to clarify the purpose and the general principles behind using custom web components with the client-side form binder API.

Fixes #410


jcgueriaud1:
I think this PR https://github.com/vaadin/docs/pull/418 fixes the same issue: https://github.com/vaadin/docs/issues/410

I think only one should be merged and the other closed. (Mine doesn't change the text, only fixes and externalize the examples)

articles/fusion/forms/appendix-web-component-field-strategy.asciidoc
```diff
@@ -1,125 +1,58 @@
 ---
-title: Using a Field Web Component
+title: Binding Data to Custom Components
 order: 100
 layout: page
 ---
 
 
-= Binding a Field Web Component
+= Using Custom Web Components in a Form
```
haijian-vaadin, line 8: Maybe it's better to also update the title here to `Binding Data to Custom Components` keep it in sync with the menu item?

haijian-vaadin (changes requested):

vlukashov:
> I think this PR #418 fixes the same issue: #410
> 
> I think only one should be merged and the other closed. (Mine doesn't change the text, only fixes and externalize the examples)

I should have checked the open issues and PRs before starting to fix this on my own :(
Now both me and @jcgueriaud1 have done the same work twice.

I would probably take the PR with the changes to the text content. Would you like to suggest changes to the code samples in this PR, @jcgueriaud1?

articles/fusion/forms/appendix-web-component-field-strategy.asciidoc
vlukashov: 👍

haijian-vaadin (approved):

PR #425, opened 2021-05-03
Deprecate forum in favor of StackOverflow

fluorumlabs:


yuriy-fix (approved):

jouni:
I’d suggest to replace `Describe your changes in imperative mood, for example "make xyzzy do frotz"` with a more realistic example. The current one doesn't mean anything and is therefore hard to understand what is required. A counter-example would also improve, which shows a non-imperative description.

fluorumlabs:
> I’d suggest to replace `Describe your changes in imperative mood, for example "make xyzzy do frotz"` with a more realistic example. The current one doesn't mean anything and is therefore hard to understand what is required. A counter-example would also improve, which shows a non-imperative description.

This is a valid suggestion, but I'd leave it outside of scope of this PR ;)

PR #432, opened 2021-05-03
Cookie consent: Java examples

anezthes:
Unsure how to make the Java examples render inside the iframe. Will need some assistance on that.


articles/ds/components/cookie-consent/index.asciidoc
```diff
@@ -40,10 +40,10 @@ include::../_shared.asciidoc[tag=scaled-examples]
 include::{root}/frontend/demo/component/cookieconsent/cookie-consent-basic.ts[render,frame,tags=snippet,indent=0,group=TypeScript]
 ----
 
-//[source,java]
-//----
-//include::{root}/src/main/java/com/vaadin/demo/component/cookieconsent/CookieConsentBasic.java[render,tags=snippet,indent=0,group=Java]
-//----
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/cookieconsent/CookieConsentBasic.java[render,tags=snippet,indent=0,group=Java]
```
jouni, line 45: Here’s how to render any example, TS or Java, inside an iframe.

```suggestion
include::{root}/src/main/java/com/vaadin/demo/component/cookieconsent/CookieConsentBasic.java[render,frame,tags=snippet,indent=0,group=Java]
```

anezthes: Thanks! Fixed.

src/main/java/com/vaadin/demo/component/cookieconsent/CookieConsentTheming.java
```diff
@@ -1,21 +1,20 @@
...
 import com.vaadin.flow.router.Route;
-import com.vaadin.demo.DemoExporter; // hidden-source-line
 
 @Route("cookie-consent-theming")
 public class CookieConsentTheming extends Div {
 
     public CookieConsentTheming() {
         // tag::snippet[]
-        CookieConsent cookieConsent = new CookieConsent();
+        add("cookie-consent-theming");
```
jouni, line 13: Can this be right? What does this even do? Is there really a `add(String)` method on the `Div` component? I would expect something like `addClassName(String)`.

anezthes: Fixed!

magi42 (commented):
AsciiDoc looks OK. The Java examples would still need a review.

jouni (approved):

PR #451, opened 2021-05-07
Collaboration Engine article titles

jouni:
## Description

The current article titles are a bit long, and some of them are formatted like Java class names, which is discouraged in general.

This change makes the titles short enough to fit on one line, and avoids the class name formatting. That said, those particular articles have very different page headings vs the nav titles. I’m open for suggestions if we can bring them closer together.

(remove unnecessary `ifdef` macros at the same time, those are no longer needed)

## Checklist

- [ ] ~~The issue is created in the corresponding repository and I have referenced it.~~
- [ ] ~~I have added tests to ensure my change is effective and works as intended.~~
- [ ] ~~New and existing tests are passing locally with my change.~~

### DIFF
--- a/articles/ce/collaboration-avatar-group.asciidoc
+++ b/articles/ce/collaboration-avatar-group.asciidoc
@@ -1,5 +1,5 @@
 ---
-title: CollaborationAvatarGroup
+title: Avatar Group
 order: 6
 layout: page
 ---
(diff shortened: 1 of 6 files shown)


magi42 (approved):

magi42:
What's the label check that stays "expected"?

PR #452, opened 2021-05-07
Display a ToC banner on tutorial pages

ozguruysal:
## Description

Display a ToC banner on tutorial pages

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist


jouni (approved):
One more thing that we could do, is add a comment at the beginning of render-banner.ts, so that if anyone goes and browsers the repo, they’ll understand that this is not something you need to worry about or use in your application.

jouni:
@tomivirkki, I want you to have a heads up about this as well. We are adding an `frontend/demo/render-banner.ts` file here, which is not something our users should care about. This could be something we need to address if a custom design system is ever going to be bootstrapped from the v14 branch in the future, or if we want to introduce these marketing banners in the `latest` branch as well.

jouni:
@ozguruysal bummer. Dev vs production build differences bit us again. If you look at production, the inages are the low-res preview versions. The script should query another image element from the DOM.

ozguruysal:
> The script should query another image element from the DOM.

Can you elaborate on that? We can host the images somewhere else like HS too, things will scatter around though.

jouni:
Instead of querying for [`.toc-banner-source img`](https://github.com/vaadin/docs/pull/452/files#diff-8ec55c0558db3af8898869d2155cd7449fe3a28e32371e84b532a8cadbd43101R32), we should query the whole `.toc-banner-source .image` element, and clone/move that into the banner element. There’s a `<noscript>` element inside it which will render the proper image. Or, we can query for `.toc-banner-source .image noscript` and the innerHTML of it. Then we don’t get the low-resolution preview image, but that’s probably okay.

PR #455, opened 2021-05-11
Fix prettier errors

yuriy-fix:
Note: PR is to check if the build is passing.


magi42:
Please have this reviewed and merged as soon as possible, as the problem is apparently blocking all other docs PR checks.

sosa-vaadin (approved):

PR #469, opened 2021-05-17
feat: Add @RouteScope docs for Spring add-on

pleku:
Fixes vaadin/flow#10900

### DIFF
--- a/articles/flow/integrations/spring/scopes.asciidoc
+++ b/articles/flow/integrations/spring/scopes.asciidoc
@@ -115,6 +115,7 @@ See <<../../advanced/application-lifecycle#application.lifecycle.ui,Application 
 NOTE: Unlike with earlier Vaadin versions 7 and 8, the `UI` and thus the `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used, and the browser is refreshed.
 To be able to preserve the beans on refresh, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
 
+[role="since:com.vaadin:vaadin@V21"]
 === RouteScope and RouteScopeOwner
 
 The `@RouteScope` annotation ties the beans to the lifecycle of Vaadin Flow routing components (`@Route`, `RouteLayout`, `HasErrorParameter`).
--- a/articles/flow/integrations/spring/scopes.asciidoc
+++ b/articles/flow/integrations/spring/scopes.asciidoc
@@ -284,6 +284,7 @@ public class User {
 ** In this example, a new `UserProfilePresenter` bean is created every time when the `UserProfileView` view is opened.
 ** The presenter bean stays the same during the time the view is attached to the UI.
 
+[role="since:com.vaadin:vaadin@V21"]
 [[routescope.preserve]]
 === Preserving Beans During Browser Refresh
 
@@ -328,4 +329,3 @@ However, any currently active routing components are preserved even if they are 
 This is due to the nature of the `@PreserveOnRefresh` feature implementation - the `UI` instance itself is not preserved, but routing components are.
 Any bean tied to the `UI` instance with `UIScope` is recreated, and the preserved routing components are moved to the new `UI`.
 To be able to preserve beans during a browser refresh, `@RouteScope` needs to be used like above.
-
--- a/articles/flow/integrations/spring/scopes.asciidoc
+++ b/articles/flow/integrations/spring/scopes.asciidoc
@@ -34,9 +34,7 @@ In addition to standard Spring scopes, the https://vaadin.com/directory/componen
 The `@VaadinSessionScope` annotation maps the Spring beans to the Vaadin session lifecycle.
 It ensures that the same bean instance is used during the whole Vaadin session.
 
-+
-*Example*: using the `@VaadinSessionScope` annotation.
-+
+*Example*: using the `@VaadinSessionScope` annotation
 [source,java]
 ----
 @Route("")
@@ -74,9 +72,7 @@ See <<../../advanced/application-lifecycle#application.lifecycle.session,Applica
 
 The `@UIScope` annotation manages the Spring beans during the `UI` lifecycle.
 
-+
 *Example*: using the `@UIScope` annotation.
-+
 [source,java]
 ----
 @Route("")
--- a/articles/flow/integrations/spring/scopes.asciidoc
+++ b/articles/flow/integrations/spring/scopes.asciidoc
@@ -113,7 +113,7 @@ See <<../../advanced/application-lifecycle#application.lifecycle.ui,Application 
 
 .Preserving UIScope beans
 NOTE: Unlike with earlier Vaadin versions 7 and 8, the `UI` and thus the `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used, and the browser is refreshed.
-To be able to preserve the beans on refresh, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
+To be able to preserve the beans on refresh, you need to use `@RouteScope` instead (available since V21), as described in <<routescope.preserve, the next chapter>>.
 
 [role="since:com.vaadin:vaadin@V21"]
 === RouteScope and RouteScopeOwner
--- a/articles/flow/integrations/spring/scopes.asciidoc
+++ b/articles/flow/integrations/spring/scopes.asciidoc
@@ -20,7 +20,8 @@ In most programming languages, a variable name refers to a unique object within 
 In Spring, an object has a unique identity within a scope, but instead of identifying the object by its variable name, it is identified by its type (object class) and qualifiers, if any.
 
 With Spring Boot, you need to mark your Vaadin related bean classes with the https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/stereotype/Component.html[`@Component`] annotation to make them picked up as managed beans.
-If no scope is specified in addition to the `@Component` annotation, the component is in https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-factory-scopes-singleton[singleton scope].
+Because there is also `Component` class in Vaadin, you can use the `SpringComponent` annotation from the Vaadin Spring add-on to avoid having to use fully qualified names.
+If no scope is specified in addition to the `@SpringComponent` / `@Component` annotation, the component is in https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-factory-scopes-singleton[singleton scope].
 
 In addition to standard Spring scopes, the https://vaadin.com/directory/component/vaadin-spring/overview[Vaadin Spring add-on] introduces three additional scope:
 
@@ -52,7 +53,7 @@ public class Editor extends Div {
     }
 }
 
-@Component
+@SpringComponent
 @VaadinSessionScope
 public class SessionService {
     private String uid = UUID.randomUUID().toString();
@@ -94,7 +95,7 @@ public class Editor extends Div {
     }
 }
 
-@Component
+@SpringComponent
 @UIScope
 public class UIService {
     private String uid = UUID.randomUUID().toString();
@@ -110,17 +111,16 @@ public class UIService {
 
 See <<../../advanced/application-lifecycle#application.lifecycle.ui,Application Lifecycle > Loading a UI>> for more information on UIs.
 
-==== UIScope Beans Are Not Preserved on Refresh
-
-Unlike with earlier Vaadin versions 7 and 8, `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used and the browser is refreshed.
-To be able to preserve the beans, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
+.Preserving UIScope beans
+NOTE: Unlike with earlier Vaadin versions 7 and 8, the `UI` and thus the `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used, and the browser is refreshed.
+To be able to preserve the beans on refresh, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
 
 === RouteScope and RouteScopeOwner
 
 The `@RouteScope` annotation ties the beans to the lifecycle of Vaadin Flow routing components (`@Route`, `RouteLayout`, `HasErrorParameter`).
-Since there can be multiple nested levels of routing components present at once, an additional annotation is needed to specify the _owner_ routing component.
-The owner is specified wit the `@RouteScopeOwner` _qualifier_ which takes as an argument the routing component that the bean has its lifecycle tied into.
-As long as the routing component is part of the active view chain, all beans owned by it remain in the scope.
+Since there can be multiple nested levels of routing components present at once, an additional `@RouteScopeOwner` _qualifier_ annotation can be used to specify the _owner_ routing component.
+Without the owner qualifier, the owner is the currently active routing component at the time of injection.
+As long as the owner routing component is part of the active view chain, all beans owned by it remain in the scope.
 
 When the `@RouteScope` bean itself is a router component, the owner can be any parent `RouterLayout` in the route chain hierarchy, or the bean itself.
 
@@ -129,6 +129,7 @@ See <<../../routing/overview#,Defining Routes With @Route>> and <<../../routing/
 *Example*: sharing a bean between two child views with the same parent layout
 [source, java]
 ----
+@SpringComponent
 @RouteScope
 @RouteScopeOwner(ParentView.class)
 public class RouteService {
@@ -160,7 +161,7 @@ public class ChildAView extends VerticalLayout {
     public ChildAView(
             @Autowired @RouteScopeOwner(ParentView.class)
             RouteService routeService) {
-        setText("Child-a: " + routeService.getText());
+        add(new Text("Child-a: " + routeService.getText()));
     }
 }
 
@@ -170,7 +171,7 @@ public class ChildBView extends VerticalLayout {
     public ChildBView(
             @Autowired @RouteScopeOwner(ParentView.class)
             RouteService routeService) {
-        setText("Child-b: " + routeService.getText());
+        add(new Text("Child-a: " + routeService.getText()));
     }
 }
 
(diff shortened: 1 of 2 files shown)


articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -8,21 +8,33 @@ layout: page
...
 
-== Using Spring Scopes
+== Using Scopes with Vaadin and Spring
 
-In most programming languages, a variable name refers to a unique object within the scope of the variable. In Spring, an object has a unique identity within a scope, but instead of identifying the object by its variable name, it is identified by its type (object class) and qualifiers, if any.
+In most programming languages, a variable name refers to a unique object within the scope of the variable.
+In Spring, an object has a unique identity within a scope, but instead of identifying the object by its variable name, it is identified by its type (object class) and qualifiers, if any.
 
-In addition to standard Spring scopes, the https://vaadin.com/directory/component/vaadin-spring/overview[Vaadin Spring add-on] introduces two additional scopes: `VaadinSessionScope` and `UIScope`.
+With Spring Boot, you need to mark your Vaadin related bean classes with the https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/stereotype/Component.html[`@Component`] annotation to make them picked up as managed beans.
```
denis-anisimov, line 22: Since `Component` exists in Flow there is a `SpringComponent` which worth to mention here.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+** When navigating inside the same browser tab between the `MainLayout` and `Editor`, the text stays the same since the service is the same.
 
-** *Example*: Navigating to the `editor` target.
-+
-[source,java]
+See <<../../advanced/application-lifecycle#application.lifecycle.ui,Application Lifecycle > Loading a UI>> for more information on UIs.
+
+==== UIScope Beans Are Not Preserved on Refresh
+
+Unlike with earlier Vaadin versions 7 and 8, `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used and the browser is refreshed.
```
denis-anisimov, line 115: It's worth to shortly mention may be why: because `UI` instance is not preserved (so UI scoped beans mapped to it are also lost).

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+To be able to preserve the beans, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
```
denis-anisimov, line 116: This is a workaround.
I would recommend do not give a false impression that this is just a direct replacement for UIScope which is preserved in refresh.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+
+When the `@RouteScope` bean itself is a router component, the owner can be any parent `RouterLayout` in the route chain hierarchy, or the bean itself.
+
+See <<../../routing/overview#,Defining Routes With @Route>> and <<../../routing/layout#,Router Layouts and Nested Router Targets>> for more about route targets, route layouts, and the active route chain.
+
+*Example*: sharing a bean between two child views with the same parent layout
+[source, java]
+----
+@RouteScope
+@RouteScopeOwner(ParentView.class)
```
denis-anisimov, line 133: This example uses a bean (not a component) and it requires an explicit injection point.

Another example which ( I would say even more important) is to annotate e.g. the bottom navigation target which allows to reuse the navigation target component even though the intermediate layouts may be recreated.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+** The injected `RouteService` bean instance is the same while the `ParentView` is attached, like when navigating between the child views.
+** When navigating to the `SiblingView`, the `ParentView` is detached. When navigating back to the `ParentView` (or child views), the a new `RouteService` bean is created.
+
+Injecting a "narrower" `RouteScope` bean into "wider" scope, like parent layout's `RouteScope` or `UIScope`, can cause problems.
+If you store a `RouteScope` bean into a wider scope bean, the stored bean instance is not automatically replaced.
+
+==== @RouteScope Without @RouteScopeOwner to Replace @ViewScope From Vaadin 7 / 8
+
+When the `@RouteScopeOwner` annotation is omitted, the bean behaves like it is in https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-factory-scopes-prototype[prototype_ scope].
+This allows you to inject a bean to a routing component once, and keep the bean the same for the lifecycle of the component.
```
denis-anisimov, line 196: This is confusing for me.
As I understand this sentence describes missing owner qualifier behavior.
And this is prototype scope.
As a result : it has no sense to use route scope without owner since there is already prototype scope (at least without my latest PR).

`This allows you to inject a bean to a routing component once, and keep the bean the same for the lifecycle of the component.` : it's confusing since this has no relation to the scope. Something injected is managed by Java object reference rules.
It's a direct reference to the object. So it may not disappear (and GCed as a result) until the instance which refers to it is alive.
This is just regular Java objects behavior and this description makes a wrong impression at least for me.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+This makes sense when you don't want to keep the beans in the memory when the route component is not active because it for example is not opened frequently or consumes a lot of memory on the server.
```
denis-anisimov, line 197: I would say : it doesn't make any sense considering routing.
There is ready-made prototype scope . If prototype scope semantic is needed it should be used directly instead route scope with confusing behavior.

(again : assuming my PR is not merged).

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+Thus using `@RouteScope` without clarifying an owner makes it possible to replace the `@ViewScope` from Vaadin 7 or 8.
```
denis-anisimov, line 198: I'm not sure : does it assume my latest PR is merged ?

Without PR the prototype scope is not a replacement for `ViewScope` .
With the PR the prototype scope may not be mentioned here and should be removed.

denis-anisimov: It's possible to get a bean via `ApplicationContext` :

```

@Autowired
private ApplicationContext context;

private void someMethod(){

context.getBeanByType(SomeBean.class);
}
```

Prototype scope: every time `context.getBeanByType(SomeBean.class)` returns a new instance
`ViewScope` : until we are inside the same view the instance will be the same.

So prototype scope (and route scope without owner) is not a replacement anyhow for `ViewScope`.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+
+.Model-View-Presenter
+NOTE: The following example is based on the _model-view-presenter_ design pattern just for the sake of demo - **it is not a best practice example**.
+It allows splitting different logical parts of the application, but adds a lot of boilerplate code.
+
+*Example*: `@RouteScope` without owner behaves like legacy Vaadin `@ViewScope`
```
denis-anisimov, line 204: No, it doesn't without my PR.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+    }
+}
+
+@Route("user-profile")
+public class UserProfileView extends VerticalLayout {
+
+    private final UserProfilePresenter presenter;
+
+    public UserProfileView(@Autowired UserProfilePresenter presenter) {
+        this.presenter = presenter;
```
denis-anisimov, line 238: It has no any relation to the scopes (unless they are too wide, like UIScope).
Prototype scope will give the same result.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+==== UIScope Beans Are Not Preserved on Refresh
+
+Unlike with earlier Vaadin versions 7 and 8, `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used and the browser is refreshed.
+To be able to preserve the beans, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
+
+=== RouteScope and RouteScopeOwner
+
+The `@RouteScope` annotation ties the beans to the lifecycle of Vaadin Flow routing components (`@Route`, `RouteLayout`, `HasErrorParameter`).
+Since there can be multiple nested levels of routing components present at once, an additional annotation is needed to specify the _owner_ routing component.
+The owner is specified wit the `@RouteScopeOwner` _qualifier_ which takes as an argument the routing component that the bean has its lifecycle tied into.
```
mshabarov, line 122:
```suggestion
The owner is specified with the `@RouteScopeOwner` _qualifier_ which takes as an argument the routing component that the bean has its lifecycle tied into.
```

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +104,224 @@ public class UIService {
...
+    }
+}
+// User entity
+public class User {
+    private String name;
+    // getter and setter omitted
+}
+
+----
+** In this example, a new `UserProfilePresenter` bean is created every time when the `UserProfileView` view is opened.
```
denis-anisimov, line 281: Like with prototype scope.

If I use `ApplicationContext::getBeanByType` instead direct injection the example will be broken.

denis-anisimov (changes requested):

articles/flow/integrations/spring/scopes.asciidoc
pleku: Done

articles/flow/integrations/spring/scopes.asciidoc
pleku: Done

articles/flow/integrations/spring/scopes.asciidoc
pleku: Done

articles/flow/integrations/spring/scopes.asciidoc
pleku: Done - it is explained more later on and this is now a note linking to it

articles/flow/integrations/spring/scopes.asciidoc
pleku: Done

articles/flow/integrations/spring/scopes.asciidoc
pleku: It was outdated since it was not updated after the decision from later on to not require the owner. And you're right that it misses the point. Fixed to highlight the semantics and usage better

articles/flow/integrations/spring/scopes.asciidoc
pleku: Done (removed)

articles/flow/integrations/spring/scopes.asciidoc
pleku: Yes and no - in case of constructor injection and then storing the instance - it is replacement, but it is not the same.
Fixed (taking into account that your PR is merged).

articles/flow/integrations/spring/scopes.asciidoc
pleku: True, but it is not used in the example (and not sure how much it would be used anyway).
But it makes sense to merge your PR as I think it clearly improves things.

articles/flow/integrations/spring/scopes.asciidoc
pleku: The code example is based on the pattern from the legacy bakery and it works by happenstance due to prototype scope... but like pointed out elsewhere it would be broken in case something relied on injecting the bean elsewhere via application context or using method injection or in case it depends on destruction - so it is again proof to your PR

articles/flow/integrations/spring/scopes.asciidoc
pleku: Yes and it should be merged.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +105,227 @@ public class UIService {
...
+To be able to preserve the beans on refresh, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
+
+=== RouteScope and RouteScopeOwner
+
+The `@RouteScope` annotation ties the beans to the lifecycle of Vaadin Flow routing components (`@Route`, `RouteLayout`, `HasErrorParameter`).
+Since there can be multiple nested levels of routing components present at once, an additional `@RouteScopeOwner` _qualifier_ annotation can be used to specify the _owner_ routing component.
+Without the owner qualifier, the owner is the currently active routing component at the time of injection.
+As long as the owner routing component is part of the active view chain, all beans owned by it remain in the scope.
+
+When the `@RouteScope` bean itself is a router component, the owner can be any parent `RouterLayout` in the route chain hierarchy, or the bean itself.
```
mshabarov, line 125: Is it worth to mention that the `@RouteScopeOwner` is optional for router components?

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +105,227 @@ public class UIService {
...
+
+.Preserving UIScope beans
+NOTE: Unlike with earlier Vaadin versions 7 and 8, the `UI` and thus the `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used, and the browser is refreshed.
+To be able to preserve the beans on refresh, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
+
+=== RouteScope and RouteScopeOwner
+
+The `@RouteScope` annotation ties the beans to the lifecycle of Vaadin Flow routing components (`@Route`, `RouteLayout`, `HasErrorParameter`).
+Since there can be multiple nested levels of routing components present at once, an additional `@RouteScopeOwner` _qualifier_ annotation can be used to specify the _owner_ routing component.
+Without the owner qualifier, the owner is the currently active routing component at the time of injection.
```
mshabarov, line 122: > Without the owner qualifier, the owner is the currently active routing component at the time of injection.

This feature is not merged yet, so may be temporary disable this sentence?

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +105,227 @@ public class UIService {
...
+                new RouterLink("Open Child-B", ChildBView.class));
+    }
 }
 ----
+** The injected `RouteService` bean instance is the same while the `ParentView` is attached, like when navigating between the child views.
+** When navigating to the `SiblingView`, the `ParentView` is detached. When navigating back to the `ParentView` (or child views), a new `RouteService` bean is created.
+
+.Injecting to wider scope
+CAUTION: Injecting a "narrower" `RouteScope` bean into "wider" scope, like parent layout's `RouteScope` or `UIScope`, can cause problems.
+If you store a `RouteScope` bean into a wider scope bean, the stored bean instance is not automatically replaced.
```
mshabarov, line 193: > the stored bean instance is not automatically replaced.

This sounds like a limitation and may raise a feeling that Flow wants automatically replace it, but it can't. Would it be better to just say that the 'narrower-scoped' bean would extend its lifespan to 'wider-scoped' bean it is being injected to.

articles/flow/integrations/spring/scopes.asciidoc
mshabarov: However, skip this comment, as other comment says we are about to merge PR revealing this.

articles/flow/integrations/spring/scopes.asciidoc
mshabarov: It is described in other section, so maybe not needed.

articles/flow/integrations/spring/scopes.asciidoc
denis-anisimov: Am I missing something or it's not committed ?
I cannot find any route chain component annotated with `RouteScopeOwner`

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +105,227 @@ public class UIService {
...
+
+==== @RouteScope Without @RouteScopeOwner to Replace @ViewScope From Vaadin 7 / 8
+
+When the `@RouteScopeOwner` annotation is omitted, the owner is the currently active route target.
+In case of nested routing hierarchies, the owner is the "leaf" / "bottom-most" active routing component.
+The bean remains in scope for as long as the routing component stays active.
```
denis-anisimov, line 199: >as the routing component

to avoid confusion which we already have  I suggest to emphasize that this is a route navigation target (not some component in the routing chain).

pleku: Done

articles/flow/integrations/spring/scopes.asciidoc
pleku: Actually it will not go like that with the most recent changes after not requiring the owner - since the route scoped bean will not be in prototype scope but always have an owner, it should be destroyed once the owner is removed.

articles/flow/integrations/spring/scopes.asciidoc
pleku: I think it is so clear that we merge it (based on other discussions in this PR) that I won't remove it.

articles/flow/integrations/spring/scopes.asciidoc
pleku: I can't imagine why would you make a routing component `@RouteScope` without any owner. I sense this will only lead into problems in the end. I would not mention anything about this, reformatted the existing text a bit.

articles/flow/integrations/spring/scopes.asciidoc
pleku: Sorry I meant that there is an explicit injection point for the bean in the components.

> Another example which ( I would say even more important) is to annotate e.g. the bottom navigation target which allows to reuse the navigation target component even though the intermediate layouts may be recreated.

Yes, this might be better, maybe I could introduce it here

pleku: Now it is done - I think it works fine here to showcase the difference between the child views

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +105,227 @@ public class UIService {
...
-** *Example*: Navigating to the `editor` target.
-+
-[source,java]
+See <<../../advanced/application-lifecycle#application.lifecycle.ui,Application Lifecycle > Loading a UI>> for more information on UIs.
+
+.Preserving UIScope beans
+NOTE: Unlike with earlier Vaadin versions 7 and 8, the `UI` and thus the `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used, and the browser is refreshed.
+To be able to preserve the beans on refresh, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
+
+=== RouteScope and RouteScopeOwner
```
mshabarov, line 118:
```suggestion
[role="since:com.vaadin:vaadin@V21"]
=== RouteScope and RouteScopeOwner
```

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -84,16 +105,227 @@ public class UIService {
...
+** When navigating inside the same browser tab between the `MainLayout` and `Editor`, the text stays the same since the service is the same.
 
-** *Example*: Navigating to the `editor` target.
-+
-[source,java]
+See <<../../advanced/application-lifecycle#application.lifecycle.ui,Application Lifecycle > Loading a UI>> for more information on UIs.
+
+.Preserving UIScope beans
+NOTE: Unlike with earlier Vaadin versions 7 and 8, the `UI` and thus the `UIScope` beans are not preserved when the `@PreserveOnRefresh` annotation is used, and the browser is refreshed.
+To be able to preserve the beans on refresh, you need to use `@RouteScope` instead, as described in <<routescope.preserve, the next chapter>>.
```
mshabarov, line 116:
```suggestion
To be able to preserve the beans on refresh, you need to use `@RouteScope` instead (available since V21), as described in <<routescope.preserve, the next chapter>>.
```

mshabarov (changes requested):
As discussed, let's marks route scope sections as available since V21

mshabarov (approved):

articles/flow/integrations/spring/scopes.asciidoc
denis-anisimov: I still don't see any component annotated with `@RouteScope` and `@RouteScopeOwner` .
Only autowired CTOR parameters.

mshabarov (approved):

PR #486, opened 2021-05-19
Fix updated lumo-css-framework import path

marcushellberg:
Fixes the changed path for importing Lumo CSS utility classes


jouni (approved):

Haprog:
I didn't notice this when creating #521 to do the same, but #521 fixes one more reference, so we can probably close this one without merging.

PR #461, opened 2021-05-14
chore: upgrade to V21 and fix Lit 2 usage and TS event names

denis-anisimov:
fixes vaadin/flow#10872

**Update**
Here's a summary of what this PR does.
- Upgrades Vaadin version to V21 and does updates required by breaking changes (in Lit, flow and our components)
  - Fix/update Lit related import paths in TS files and example snippets (additionally make these imports a bit more consistent in code style and ordering since we're touching most of the files anyway)
  - Replace `@internalProperty` decorator with `@state` (this is a rename in Lit API)
  - Fix TS compilation by updating Java classes related to breaking nullability changes in vaadin/flow#8849
  - Use new Vaadin Component TS events as required due to this breaking change vaadin/web-components#265
  - Fix `applyTheme()` usage due to breaking change in Lit 2 (calling it in the constructor does not work anymore as it is now too early. `createRenderRoot()` is now called later by Lit so `this.shadowRoot` is not available in the constructor)
- Adds a section into `upgrading.asciidoc` about Lit 2 related changes

PR #490, opened 2021-05-20
chore: add more abbreviations to exceptions

web-padawan:
## Description

Added some abbreviations to exceptions based on the Vale comments from #461 

## Type of change

- [x] Linter config


magi42 (approved):
Might be OK, but please note that any non-common abbreviations should not be included in the exceptions list, but introduced in the text, such as: "a Progressive Web Application (PWA) ..."
I don't have a strong opinion about what abbreviations are such. Of these, maybe CDI, CRM, CSRF, MDN, POJO, and PWA might be non-obvious to a casual reader and should rather be introduced.

web-padawan:
Thanks. I can remove these from the exceptions then.

PR #499, opened 2021-05-24
Documentation for Utility Classes

anezthes:
Adds documentation for Lumo CSS utility classes.


jouni:
Verified locally that the build works. I think our production deployment is using Node 14+, so it should not fail on the same issue as the preview build (https://github.com/43081j/eslint-plugin-lit/pull/112)

jouni (approved):

PR #501, opened 2021-05-24
fix: clarify routing components scope in Spring

pleku:
Adds a note about how routing components behave out-of-the-box as Spring
initialized beans.

Related to vaadin/flow#10900


articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -12,7 +12,7 @@ layout: page
 
 * *Scopes* are narrower than contexts. While conventional singletons are application-wide, objects managed by a Spring container are "singletons" in a narrower scope. Examples include a user session, a particular UI instance associated with the session, or even a single request. A scope defines the lifecycle of the object, that is its creation, use, and destruction.
 
-More information on beans and scopes in Spring can read from the https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-definition[Spring project documentation].
+More information on beans and scopes in Spring can be read from the https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-definition[Spring project documentation].
```
jouni, line 15:
```suggestion
Read more about the https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-definition[beans and scopes in Spring].
```

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -29,6 +29,12 @@ In addition to standard Spring scopes, the https://vaadin.com/directory/componen
 ** `UIScope` for targeting beans to a browser window or tab opened by the user
 ** `RouteScope` for targeting beans to certain routing components pointed by the `RouteScopeOwner` qualifier
 
+.Routing components are Spring-initialized beans
+IMPORTANT:
```
jouni, line 33:
```suggestion
[IMPORTANT]
```

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -29,6 +29,12 @@ In addition to standard Spring scopes, the https://vaadin.com/directory/componen
...
+By default, all Flow routing components (`@Route`, `RouterLayout` or `HasErrorParameter`) are Spring-initialized and act like a bean in _prototype_ scope even without explicitly adding the `@SpringComponent` and `@Scope("prototype")` annotations.
```
mshabarov, line 34: Do the routing components indeed act like prototypes when `com.vaadin.flow.spring.annotation.SpringComponent` or `org.springframework.stereotype.Component` annotated? 

I believe its a _singleton_ and if some instance of bean class annotated with `@RouteScope` is being auto-wired to this routing component, then the `IllegalStateException` with `No VaadinSession bound to current thread` is thrown, because the sessions are invisible for singleton.

mshabarov: And this is also good to mention.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -29,6 +29,12 @@ In addition to standard Spring scopes, the https://vaadin.com/directory/componen
...
+You can inject other beans to the components, but note that while `@PostConstruct` methods work, `@PreDestroy` methods do not.
+It is possible to widen the scope of the routing components with `@UIScope` and `@RouteScope` as described later.
+
```
mshabarov, line 37: Another thing I would propose to add: when adding `@RouteScope` to the navigation route target, don't forget to annotate it with `@SpringComponent`, otherwise the route scope annotation would not give any effect (no-op). I've run through it two times and it took a while to understand where is the mistake.

pleku: Well, this applies to anything and it should be visible from the examples.  It now emphasized here too

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -12,7 +12,7 @@ layout: page
 
 * *Scopes* are narrower than contexts. While conventional singletons are application-wide, objects managed by a Spring container are "singletons" in a narrower scope. Examples include a user session, a particular UI instance associated with the session, or even a single request. A scope defines the lifecycle of the object, that is its creation, use, and destruction.
 
-More information on beans and scopes in Spring can be read from the https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-definition[Spring project documentation].
+Read more about the https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-definition[beans and scopes in Spring].
```
jouni, line 15: Didn't pay attention to the unnecessary article.

```suggestion
Read more about https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-definition[beans and scopes in Spring].
```

mshabarov (approved):

PR #502, opened 2021-05-24
Changed the first code sample to use VerticalLayout instead of Div

samie:
Introduce new users with more high-level API first.

## Description

This was discussed more widely earlier, but primarily we should introduce using high-level component API for new Vaadin users, and creating a Div might be misleading.

### DIFF
--- a/articles/guide/introduction/index.asciidoc
+++ b/articles/guide/introduction/index.asciidoc
@@ -13,8 +13,8 @@ For example, you can create a simple UI in Java as follows:
 
 [source, java]
 ----
-// Create an HTML element
-Div layout = new Div();
+// Create a layout for other components
+VerticalLayout layout = new VerticalLayout();
 
 // Use TextField for standard text input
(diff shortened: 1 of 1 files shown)


tarekoraby (changes requested):
Isn't `addComponents()` a Framework 8 thing? It should remain `add()` I think.

magi42 (approved):
OK. The purpose of using Div there was just to demonstrate the use of regular HTML elements along with Vaadin components. It apparently wasn't appropriate in that case, so maybe need to demonstrate it in some other way in this or other place.

tarekoraby:
@jouni I think `addComponents()` in this PR is nonexistent in v10+.

jouni:
I damn, I thought someone already changes that.

jouni:
Updated in https://github.com/vaadin/docs/commit/d44c3d1b622003df893f1eb73bfee1f25051dba8

PR #504, opened 2021-05-25
CE: Fix topic-api examples

tulioag:
The partial examples don't compile if you simply omit parts of the code.
I inlined the partial examples so I could add a "return null" to in the
lambda expression.

### DIFF
--- a/articles/ce/topic-api.asciidoc
+++ b/articles/ce/topic-api.asciidoc
@@ -29,7 +29,14 @@ First, create the view with a checkbox but not yet any collaborative functionali
 
 [source,java]
 ----
-include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=topic-view;add-components;!*,indent=0]
+// NOTE: In a real application, use the user id of the logged in user
+// instead
+String userId = System.identityHashCode(UI.getCurrent()) + "";
+UserInfo localUser = new UserInfo(userId, "User " + userId);
+CollaborationEngine.getInstance().openTopicConnection(this, "tutorial",
+        localUser, topic -> {
+          return null;
+        });
 ----
 
 === Define Topic Connection
@@ -43,7 +50,12 @@ The user who is related to the topic connection must be defined.
 
 [source,java]
 ----
-include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=user-info;open-topic;!*,indent=0]
+CollaborationEngine.getInstance().openTopicConnection(this, "tutorial",
+        localUser, topic -> {
+            CollaborationMap fieldValues = topic
+                    .getNamedMap("fieldValues");
+            return null;
+        });
 ----
 
 The activation callback should return a registration that is run when the connection is deactivated.
--- a/articles/ce/topic-api.asciidoc
+++ b/articles/ce/topic-api.asciidoc
@@ -29,14 +29,7 @@ First, create the view with a checkbox but not yet any collaborative functionali
 
 [source,java]
 ----
-// NOTE: In a real application, use the user id of the logged in user
-// instead
-String userId = System.identityHashCode(UI.getCurrent()) + "";
-UserInfo localUser = new UserInfo(userId, "User " + userId);
-CollaborationEngine.getInstance().openTopicConnection(this, "tutorial",
-        localUser, topic -> {
-          return null;
-        });
+include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=topic-view;add-components;!*,indent=0]
 ----
 
 === Define Topic Connection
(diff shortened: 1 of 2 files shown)


jouni (commented):
I think there should be a better way than inline the examples. I assume the problem is that the body of the lambda function is omitted. I also assume it would be okay if it showed something like `...` or `10 lines hidden, click to expand` in there.

tulioag:
Yes, the problem is that if you copy and paste those examples, the code won't compile, which makes it seem as you did something wrong or the example is broken.

>  I also assume it would be okay if it showed something like `...` or  `10 lines hidden, click to expand` in there (i.e. that the current "Expand code" button is too hidden).

How can I do that?

jouni:
> How can I do that?

Ideally you wouldn't have to do anything, and DS Publisher did that automatically for you when you include tagged regions. I haven’t prototype that, how easily that could be done (or if it’s doable at all). It is somewhat possible to fake it with something like this:

```asciidoc
[source,java]
----
include::{root}/path/to/source.java[tag=part-before-omitted-lines]
// some lines hidden, click the "Expand code" button above to show them
include::{root}/path/to/source.java[tag=part-after-omitted-lines]
----
```

path/to/source.java:
```java
// package, imports, etc., everything before the snippet basically

// tag::part-before-omitted-lines[]
        // NOTE: In a real application, use the user id of the logged in user
        // instead
        String userId = System.identityHashCode(UI.getCurrent()) + "";
        UserInfo localUser = new UserInfo(userId, "User " + userId);
        CollaborationEngine.getInstance().openTopicConnection(this, "tutorial",
                localUser, topic -> {
// end::part-before-omitted-lines[]
                    CollaborationMap fieldValues = topic
                            .getNamedMap("fieldValues");
                    Registration registration = checkbox
                            .addValueChangeListener(valueChangeEvent -> {
                                fieldValues.put("isFriday",
                                        valueChangeEvent.getValue());
                            });
                    fieldValues.subscribe(event -> {
                        if ("isFriday".equals(event.getKey())) {
                            checkbox.setValue(Boolean.TRUE
                                    .equals(event.getValue(Boolean.class)));
                        }
                    });
                    return registration;
// tag::part-after-omitted-lines[]
                });
// end::part-after-omitted-lines[]

// all the rest of the file
```

pekam:
In this kind of a step-by-step tutorial, it would be good to have compiling code after each step. So ideally the callback would return `null` until there's a `Registration` that should be returned. Most readers probably don't expand the code examples, and I think we should optimize for how the code examples look when not expanded.

Then we can't use a single source code example, so we need to either write some parts inline, or add a separate method (or even class) for the first steps which return `null` from the callback.

tulioag:
I tried to minimize the amount of inlined code. Now it's just the `Registration registration = null;` line

articles/ce/topic-api.asciidoc
```diff
@@ -43,7 +43,10 @@ The user who is related to the topic connection must be defined.
 
 [source,java]
 ----
-include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=user-info;open-topic;!*,indent=0]
+include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=user-info;open-topic-method-call;!*,indent=0]
+            Registration registration = null;
```
pekam, line 47: I think it's a good idea to write only one line inline like this. But it feels redundant at this point to define a `null` `Registration`. I'd suggest to change this to `return null` and then exclude the `return registration` line also from these first steps.

articles/ce/topic-api.asciidoc
```diff
@@ -55,7 +58,10 @@ In the activation callback, get a map by its name:
...
 [source,java]
 ----
-include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=open-topic;get-map;!*,indent=0]
+include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=user-info;open-topic-method-call;get-map;!*,indent=0]
+            Registration registration = null;
+include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=return-registration;!*,indent=12]
+include::{root}/src/main/java/com/vaadin/demo/ce/TopicView.java[tags=open-topic-method-call-end;!*,indent=8]
 ----
 
 === Pass Values to the Topic
```
pekam, line 67: Looking at the preview, the code snippet in this section now shows the wrong thing.

articles/ce/topic-api.asciidoc
tulioag: Done

articles/ce/topic-api.asciidoc
tulioag: Yes, thanks for noticing. I had changed the tags in `TopicView.java`, I think it's fixed now.

pekam (approved):

PR #506, opened 2021-05-25
chore: add script for updating versions in docs

ZheSun88:
## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes # (issue)

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] I have performed self-review and corrected misspellings.

### DIFF
--- a/scripts/version-update.sh
+++ b/scripts/version-update.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+set -o pipefail
+
+# Script for updating vaadin versions used in docs
+# Usage:  bash ./scripts/version-update.sh <update-version> <currentLatestReleaseMajor>
+
+# update versions in v14 branch
+# sed usage: sed -i "s/<replacing-string>/<replaced-string>/g" "<file>"
+#            use -i to edit files in-place
+#            use /g for matching the replacing-string globally
+updateV14Branch() {
+  #https://github.com/vaadin/docs/blob/v14/articles/guide/start/maven.asciidoc
+  sed -i "s/-DarchetypeVersion.*/-DarchetypeVersion=[replaceable]#$1# \\\/g" "articles/guide/start/maven.asciidoc" || exit 1
+  
+  #https://github.com/vaadin/docs/blob/v14/articles/guide/upgrading/v10-13/index.asciidoc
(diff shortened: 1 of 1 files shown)


tarekoraby (approved):

jouni:
This feels like overkill to me. Or is there some automation that runs this script when a new version is available?

I mean, we could instead rely on Asciidoc attributes (like we at some point did), which defines the version in one place, and then that is included in those articles that need to use the version number.

jouni:
> Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Instead of this placeholder text, it would be good to write at least something about why the changes are needed/made, and not just assume future readers will be able to parse it from the source code.

tarekoraby:
> Or is there some automation that runs this script when a new version is available?
I assumed that this is the case. But @ZheSun88 can kindly confirm.

tarekoraby:
> Edit: I see there’s a clear description in #505. I’ll copy it here.

Done it already :-)

jouni:
Great, if it’s automatically kept up to date. But still, I think we should consolidate that version number in one place, and the script would only update that. Now there’s a risk that a new article is created with a hard-coded version number but it’s not included in the script.

jouni:
Here’s what I would do:
- Create `articles/_vaadin-version.adoc`, with the following contents:
     ```asciidoc
     :vaadin-version: 20.0.0
     ```
- Include that in articles that require it (after the page heading, e.g. `= Packaging a Theme for Reuse`:
     ```asciidoc
     = Packaging a Theme For Reuse
     include::{root}/_vaadin-version.adoc[]
     ```
- Use the attribute where needed:
     ```asciidoc
     .`pom.xml`
     [source,xml,subs=attributes]
     ----
     <vaadin.version>{vaadin-version}</vaadin.version>
     ----
     ```
- Modify the script to update the `_vaadin-version.adoc`

tarekoraby:
Sounds good to me. What do you think @ZheSun88? I know that you have already done us the favor of writing the script to make the list of specific versions automated, and I'm embarrassed to ask you if you can do more, but perhaps the approach suggested by @jouni would make sure that we won't bother you (as much) with doc-related requests in the future.

PR #508, opened 2021-05-25
Revised the CE architecture diagram

magi42:
Revised the diagram, as the font was too small to read.


jouni:
The SVGs are quite massive. I could redraw them in Figma, see if I can optimize the size.

jouni:
The SVG to PNG conversion script is IMO not needed. If that is something that is needed for the PDF book conversion, it should be handled in that process.

magi42:
@jouni Ah, so it seems, it must come from the text-to-path export as the original is just 78 kB. The PNG seems to be half the size, so could use that as well.

But if you think it's better to redraw in Figma, go ahead. Also the 78 kB original seems quite big... Apparently Inkscape likes to store all settings for every element.

magi42:
Don't know if the newer book build system still needs PNGs. In 2016 it couldn't handle SVGs or transparent PNGs for that matter. Sure can remove the PNG for now.

jouni:
Please ask for assistance in the #design-support Slack channel. We still don't have a unified visual style for these diagrams, though, so it’s difficult for a designer to decide how they should be presented.

magi42:
@Peppe Any suggestions for the colors? One constraint is that diagrams in Docs should have white background, because 1) it's more suitable for printing, 2) otherwise there will be an additional white border like there is in the currently CE diagram. Well, unless the diagram is transparent, in which case it will get gray background from the image box...

jouni:
To make the eventual dark mode even better for readers, we could optimize the SVG colors to adapt to the page colors. That requires the use of `image::[opts=inline]`. 

Not a requirement, just wanted to put the idea here so that I might remember it later.

magi42:
I asked from #design-support about the architecture diagram styles and they'll discuss it in the design team at some time. It might be part of an upcoming new style guide. So for now it seems can't get anything quickly. I could make some color changes and such easily, and like change the bullet arrows to regular ones, etc. Or do we go with my current proposition and improve later?

magi42:
OK it now has lines and colors similar to the original, except for reversing the gray background. Switched to using SVG, and that doesn't actually look bad on dark background either.

jouni:
It’s perhaps not the best looking diagram, but I’m not blocking the update. The design team will get back to these eventually.

PR #509, opened 2021-05-25
Remove /start from Get Started guide

tarekoraby:
Remove reference to the deprecated `/start` from the diagram in https://vaadin.com/docs/v20/guide/start.


magi42 (approved):
I don't know what's the current policy, I understand that s.v.c is nowadays preferred, but vaadin.com/start is not invalid, is it? There's many downloadable example projects there that are outside the scope of s.v.c. Almost all of them are for V14, but there's at least one for V19.

Anyhow, go ahead if this is the policy and this is in the latest branch anyhow, not v14.

Oh, the diagram probably should have white background. Doesn't matter much, but you could correct that as well, or leave it to the next iteration.

tarekoraby:
> Oh, the diagram probably should have white background. Doesn't matter much, but you could correct that as well, or leave it to the next iteration.
 
I will have to leave this for a later point indeed, but thanks for noting :-)

PR #510, opened 2021-05-26
Reload on session expiration when included example is loaded

yuriy-fix:
Requires https://github.com/vaadin/docs-app/pull/277
Fixes https://github.com/vaadin/docs-app/issues/273

### DIFF
--- a/frontend/demo/example-resources.ts
+++ b/frontend/demo/example-resources.ts
@@ -1,5 +1,6 @@
 // @ts-ignore
 import { withPrefix } from 'gatsby';
+import { differenceInMinutes } from 'date-fns';
 import './init-flow-namespace';
 import { applyTheme } from 'Frontend/generated/theme';
 // See webpack.config 'externals.Frontend/generated/theme'
@@ -9,22 +10,25 @@ window.__applyTheme = { applyTheme };
 // @ts-ignore
 import('all-flow-imports-or-empty').catch(() => {});
 
-window.addEventListener('included-example-loaded', (e) => {
-  const includeUrl = (e as any).detail;
-  // Make sure example includes Vaadin component
-  if (includeUrl && includeUrl!.indexOf('vaadin') > -1) {
-    // Make sure Vaadin server is up and running
-    fetch(withPrefix(includeUrl)).then((urlFetch) => {
-      if (urlFetch.ok) {
-        fetch(withPrefix('/vaadin/?v-r=heartbeat&v-uiId=0'), {
-          method: 'POST',
-        }).then((data) => {
-          // If session is expired reload the page
-          if (!data.ok) {
-            location.reload();
-          }
-        });
-      }
-    });
-  }
-});
+let reloadDate = new Date();
+
+function testHeartbeat() {
+  fetch(withPrefix('/vaadin/?v-r=heartbeat&v-uiId=0'), { method: 'POST' }).then((data) => {
+    if (!data.ok && differenceInMinutes(reloadDate, new Date()) > 5) {
+      reloadDate = new Date();
+      location.reload();
+    }
+  });
+}
+
+const initialListener = () => {
+  window.removeEventListener('included-example-loaded', initialListener);
+  fetch(withPrefix('/vaadin/index.html')).then((serverData) => {
+    if (serverData.ok) {
+      window.addEventListener('included-example-loaded', () => testHeartbeat());
+      testHeartbeat();
+    }
+  });
+};
+
+window.addEventListener('included-example-loaded', initialListener);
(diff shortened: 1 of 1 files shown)


frontend/demo/example-resources.ts
```diff
@@ -6,3 +9,26 @@ import { applyTheme } from 'Frontend/generated/theme';
 window.__applyTheme = { applyTheme };
 // @ts-ignore
 import('all-flow-imports-or-empty').catch(() => {});
+
+let reloadDate = new Date();
```
tomivirkki, line 13: This probably wouldn't work for preventing an accidental reload loop since `reloadDate` is always initialized on load so you never exceed the 5 min safe period. Store the current time in localStorage before `location.reload()` and compare against a value that was saved earlier (if it exists).

yuriy-fix: On reload the Date is set, so that the next reload cannot happen sooner than in 5 minutes as that value won't get overridden (noticed there was a redundant part of setting it again before reload). 

Just to make sure it works, I have switched the solution to use `localStorage` and compare the value.

frontend/demo/example-resources.ts
```diff
@@ -6,3 +9,29 @@ import { applyTheme } from 'Frontend/generated/theme';
 window.__applyTheme = { applyTheme };
 // @ts-ignore
 import('all-flow-imports-or-empty').catch(() => {});
+
+function testHeartbeat() {
+  fetch(withPrefix('/vaadin/?v-r=heartbeat&v-uiId=0'), { method: 'POST' }).then((data) => {
+    const reloadDate =
+      localStorage.getItem('reloadDate') && new Date(localStorage.getItem('reloadDate') as string);
```
tomivirkki, line 16: "Parsing of date strings with the Date constructor (and Date.parse(), which works the same way) is strongly discouraged due to browser differences and inconsistencies."
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#timestamp_string

Rather store and use the value from `Date.now()` (timestamp) and compare the millisecond difference (no need for Date object / date-fns)

frontend/demo/example-resources.ts
```diff
@@ -6,3 +9,29 @@ import { applyTheme } from 'Frontend/generated/theme';
...
+    if (
+      !data.ok &&
+      (!reloadDate || (reloadDate && differenceInMinutes(new Date(), reloadDate) > 5))
```
tomivirkki, line 19: nit: `(!reloadDate || (reloadDate && ...)) ` checks for `reloadDate` twice

frontend/demo/example-resources.ts
yuriy-fix: Aware of those, but as it is set and checked within the same browser I was assuming it should be Ok.

Anyway the `timestamp` from `Date.now()` looks cleaner, so updated it to that one. Thanks for suggestion! 🙇 

Note: showing explicit calculation `5 * 60 * 1000`, so it will be obvious that it's the difference in 5 minutes.

frontend/demo/example-resources.ts
yuriy-fix: Updated.

frontend/demo/example-resources.ts
```diff
@@ -6,3 +8,25 @@ import { applyTheme } from 'Frontend/generated/theme';
 window.__applyTheme = { applyTheme };
 // @ts-ignore
 import('all-flow-imports-or-empty').catch(() => {});
+
```
tomivirkki, line 11: nit: There could be some comments explaining the purpose of this code (as discussed last retro)

tomivirkki (approved):

frontend/demo/example-resources.ts
```diff
@@ -6,3 +8,32 @@ import { applyTheme } from 'Frontend/generated/theme';
...
+    }
+  });
+}
+
+const initialListener = (e: any) => {
+  window.removeEventListener('included-example-loaded', initialListener);
+  // Make sure flow example is upgraded and requested content from server
+  customElements.whenDefined(e.detail).then(() => {
+    // Make sure server is up and running
+    fetch(withPrefix('/vaadin/index.html')).then((serverData) => {
```
tomivirkki, line 29: Is this fetch still necessary? You now have the `whenDefined` check, which wouldn't resolve if the server didn't respond with an embedded component definition.

yuriy-fix: It is there for the cases included example is not a Vaadin embedded component.
Could be redundant if we assume included examples can be loaded only from Vaadin server. Should I remove that one? 🤔

tomivirkki: Shouldn't be an issue having the check here, even though effectively, in this project all included examples are Vaadin embedded components. It's only run once so it's not that much additional overhead.

tomivirkki (approved):

PR #515, opened 2021-05-28
Suggest excluding vaadin-dev-server module in production mode

joheriks:
Suggest how to exlude `com.vaadin:vaadin-dev-server` in production mode for smaller application. Not that this is not the default in starter applications yet.

Related to vaadin/flow#10892

### DIFF
--- a/articles/guide/production/index.asciidoc
+++ b/articles/guide/production/index.asciidoc
@@ -62,13 +62,37 @@ To create a production build, you can call `mvn clean package -Pproduction`.
 This will build a JAR or WAR file with all the dependencies and transpiled front end resources, ready to be deployed.
 The file can be found in the `target` folder after the build completes.
 
-If you do not have the the production Maven profile in your POM file, the easiest way to get it is to get a project base from https://vaadin.com/start matching your environment (Spring Boot, Jakarta EE, or plain Java), and copy the production profile from the downloaded POM file.
+If you do not have the production Maven profile in your POM file, the easiest way to get it is to get a project base from https://vaadin.com/start matching your environment (Spring Boot, Jakarta EE, or plain Java), and copy the production profile from the downloaded POM file.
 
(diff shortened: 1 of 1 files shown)


haijian-vaadin (changes requested):
Got this exception while trying to run in production mode after following the described steps
```


```
Trying to include `vaadin.enableDevServer = false` in the application.properties file, still the same. Wonder how can I get it work?

articles/guide/production/index.asciidoc
```diff
@@ -62,14 +62,38 @@ To create a production build, you can call `mvn clean package -Pproduction`.
...
+----
+<dependency>
+    <groupId>com.vaadin</groupId>
+    <artifactId>vaadin</artifactId>
+    <exclusions>
+        <exclusion>
+            <groupId>com.vaadin</groupId>
+            <artifactId>vaadin-dev-server</artifactId>
+        </exclusion>
+    </exclusions>
```
haijian-vaadin, line 91: maybe better to wrap the code into <dependencies></dependencies>. Otherwise got compile error when copy-pasting the code into pom.xml

haijian-vaadin (changes requested):

joheriks:
> Got this exception while trying to run in production mode after following the described steps
> 
> Caused by: java.lang.RuntimeException: no DevModeHandlerManager implementation found but but dev server enabled. Either disable by setting vaadin.enableDevServer=false (and run the build-frontend maven goal) or include the vaadin-dev-server dependency
> 
> Trying to include `vaadin.enableDevServer = false` in the application.properties file, still the same. Wonder how can I get it work?

Should be fixed in https://github.com/vaadin/spring/commit/01b88375a0f9901a13e709c0817a08cd47c63fd7

articles/guide/production/index.asciidoc
joheriks: Added some context to the pom fragment.

haijian-vaadin (approved):

PR #523, opened 2021-06-01
Remove reference to start.spring.io

tarekoraby:


jouni (approved):

mstahv:
Why did we do this? Spring Initializr is probably the best way for pro users to create a Vaadin project as they can also configure other things in their app, not **just** UI. Vaadin is **just** the UI and what start.vaadin.com provides for that is not nearly as versatile as Spring Initializr.

tarekoraby:
> Why did we do this?  

Above all other considerations, we shouldn't send developers away from our website unless it's absolutely necessary (for example, to install a prerequisite such as Maven or JDK). If we send them elsewhere, we are significantly reducing the chance of their return to Vaadin (maybe they will get distracted in Spring's website and try a different tech like, say, thymeleaf). 

> Spring Initializr is probably the best way for pro users to create a Vaadin project as they can also configure other things in their app, not just UI. 
If that's true, then we should improve our experience to make sure that the best way to create a Vaadin project (for pro-users and beginners) is provided by us, not others.

mstahv:
IMHO, not giving our users/evaluators the best possible tools (and possibly tools they are already familiar with) to get started increases the likelihood of ditching Vaadin more than getting to different website. We are not selling to average consumers here, they probably open that link to a new tab anyways.

tarekoraby:
In your opinion, what are the features that make Spring Initializer especially attractive compared to our start.vaadin.com? Could we perhaps work on incorporating those on our website?

There are few things to keep in mind if we send people to Spring Initializer:

1. They might get distracted by other technologies there
2. By default, they will only get the LTS version (hence we won't be able to get the opportunity to showcase our latest enhancements)
3. If I'm not mistaken, they will have to add dependencies to Vaadin components themseleves (which is far from obvious for new evaluators, especially if we send them away from our website). 
4. We lose the ability to collect statistics on how well or bad our getting-started experience is doing


Those points aside, we should certainly ensure that Vaadin works with all Spring Boot versions (especially the current default, 2.5.0) before we direct people there.

samie:
Valid point about 2.5.0, but I'll jump in since this has been discussed and analyzed a couple of times. It is not our tool, but it is an awesome tool to get started with Vaadin - especially if you already are a Spring user.  

Main benefit is that you get project without any views and combine Vaadin with other project dependencies (like DB drivers and other backend tools). Vaadin is typically not the only technology used in an application and start.vaadin.com does not currently provide a way to configure those dependencies. 

1. They are not badly distracted (there is no Vaadin competitors there), but offered easy way to configure other libraries and dependencies for the project. But we could always link to a good Vaadin preset, [like this.](https://start.spring.io/#!type=maven-project&language=java&platformVersion=2.4.6.RELEASE&packaging=jar&jvmVersion=11&groupId=org.vaadin&artifactId=example-app&name=example-app&description=Vaadin%20project%20for%20Spring%20Boot&packageName=org.vaadin.example-app&dependencies=vaadin)
2. LTS is the default for new users (who do not know Vaadin well). It is easy for advanced Vaadin users to change it in the pom.xml.
3.  `vaadin-bom` includes all the component dependencies and we also have a special Vaadin Maven plugin support there for production builds. 
4. This is the same as in start.vaadin.com. We only get the download numbers, but nothing else. I hope to fix this particular one to make it useful for all users. 💪 🙂 

My personal thinking is, if they offer anything more than we can offer, it is good and honest to mention. Developers are opinionated about tooling and more ways to get started is generally better for more users.

tarekoraby:
Good points! Thanks for sharing.

mstahv:
Sami has been awake on the lessons, but one very important thing about Spring Initialzr was left out. One can use (and I think most do and I always use) IDE to generate the projects. No need to go with browser anywhere or to do the download-extract-import dance.

If they leave the Java IDE and open their browser they might end up on reactjs.org page 🤪

PR #525, opened 2021-06-02
Docs for vaadin-message-list theme API

tulioag:
Fixes: https://github.com/vaadin/collaboration-engine-internal/issues/595

### DIFF
--- a/articles/ds/components/messages/index.asciidoc
+++ b/articles/ds/components/messages/index.asciidoc
@@ -74,6 +74,26 @@ include::{root}/src/main/java/com/vaadin/demo/component/messages/MessageInputCom
 ----
 --
 
+== Theming
+
+You can style individual messages by adding a theme property to some of the items and providing css using that theme. This can be usefull, for example, to highlight the current user's own messages.
+
+[.example]
+--
+[source,css]
+----
+include::{root}/frontend/themes/docs/components/vaadin-message-with-theme.css[render,tags=snippet,indent=0,group=CSS]
+----
+[source,html]
+----
+include::{root}/frontend/demo/component/messages/message-list-with-theme-component.ts[render,tags=snippet,indent=0,group=TypeScript]
+----
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/messages/MessageListWithThemeComponent.java[render,tags=snippet,indent=0,group=Java]
+----
+--
+
 == Related Components
 
 [cols="1,2"]
--- a/articles/ds/components/messages/index.asciidoc
+++ b/articles/ds/components/messages/index.asciidoc
@@ -83,7 +83,7 @@ This can be useful, for example, to highlight the current user's own messages.
 --
 [source,css]
 ----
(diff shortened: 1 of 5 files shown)


articles/ds/components/messages/index.asciidoc
```diff
@@ -85,6 +74,27 @@ include::{root}/src/main/java/com/vaadin/demo/component/messages/MessageInputCom
...
+== Theming
+
+You can style individual messages by adding a theme property to some items and providing CSS using that theme.
+This can be useful, for example, to highlight the current user's own messages.
+
+[.example]
+--
+[source,css]
+----
+include::{root}/frontend/themes/docs/components/vaadin-message-with-theme.css[render,tags=snippet,indent=0,group=CSS]
```
jouni, line 86: Don't use `render` for source files that don't export a custom element.

Don't use `tags` if you are including all the lines in the file. Then you also don't need `indent` because we want to preserve the indentation as-is from the source file.

Use `group` only to divide examples into TypeScript and Java. Files that are relevant for both can be without a group.

```suggestion
include::{root}/frontend/themes/docs/components/vaadin-message-with-theme.css[]
```

frontend/themes/docs/components/vaadin-message-with-theme.css
```diff
@@ -0,0 +1,18 @@
...
+:host([theme~="current-user"]) [part="name"] {
+    font-weight: 900;
+}
+:host([theme~="current-user"]) [part="time"] {
+    color: #fff;
+}
+:host([theme~="current-user"]) [part="name"]::after{
+    content: " (You)";
+}
+/* end::snippet[] */
```
jouni, line 18:
```suggestion
/* Place in frontend/themes/[my-theme]/components/vaadin-message.css */

:host([theme~="current-user"]) {
    background-color: #000;
    color: #fff;
    border: 2px solid #fff;
    border-radius: 9px;
    font-weight: 900;
  }
:host([theme~="current-user"]) [part="name"] {
    font-weight: 900;
}
:host([theme~="current-user"]) [part="time"] {
    color: #fff;
}
:host([theme~="current-user"]) [part="name"]::after{
    content: " (You)";
}
```

frontend/themes/docs/components/vaadin-message.css
```diff
@@ -0,0 +1 @@
+@import 'vaadin-message-with-theme.css';
```
jouni, line 1: This kind of obscures the place where the CSS should be placed in the project. The reader will just see `vaadin-message-with-theme.css`, perhaps not realizing that the CSS in that file should be placed in the `my-theme/components/vaadin-message.css` file.

This is a problem with other component examples as well, not just here.

jouni: A simple code comment in the CSS file might suffice. I added that as a suggestion.

tulioag: Thanks for the review. I committed your suggestions.

heruan (approved):

PR #526, opened 2021-06-03
chore: Add Vaadin Spring scopes diagram

mshabarov:
## Description

Add a simple diagram to represent visually a Vaadin Scopes overview.

Related-to https://github.com/vaadin/flow/issues/11083

## Type of change

- [ ] Bugfix
- [ ] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] I have performed self-review and corrected misspellings.


pleku (approved):

magi42 (commented):
I get a bit confused for which class the annotations are for? Are they both annotations for a URL or the ProductView class? Should "URL" rather be "Route"?

mshabarov:
> I get a bit confused for which class the annotations are for? Are they both annotations for a URL or the ProductView class? Should "URL" rather be "Route"?

Thanks for your question, @magi42 ! Making this diagram, I aimed to show visually the following:
1. Bean lifecycle scopes are 'nested', i.e. the 'wider' scope overlaps the 'narrower' scope, for example, beans annotated with `@VaadinSessionScope` live longer (or at least equally) than the other beans annotated with`@UIScope`.
2. While the scope meaning is more or less abstract term, there are 'physical' counterparts for them: application running on the server, user's session created on the server (which is an object in the memory), opened browser page and finally the URL typed in the address bar by the user. So, showing these APPLICATION, USER'S SESSION, ... on the left side of squares, and the corresponding annotations on the left side, I was trying to tell the reader that those annotation are bound to a 'physical' things, that's IMO easier to understand, especially for those who uses it for the first time.
3. For `@RouteScope` it's not that easy to show only an empty square...Because this scope always bound to some navigation target (i.e. class annotatated with `@Route`, `@RouterLayout` or `HasErrorParameter`), even if no owner is set. This forced me to show the view hierarchy of the imaginary application and pass round some sub-set of the views.

> I get a bit confused for which class the annotations are for?

This is unfortunately not shown in this picture and not described in the title. All the annotations showed in the picture are targeted to the bean classes, which can be a navigation targets, or other non-UI related classes considered as managed beans in the application (for instance, controllers/services). These bean classes are not show in the picture.

> Are they both annotations for a URL or the ProductView class?

They can be an annotations for `ProductView` technically, but this example says that `ProductView` is an owner (i.e. the scope) for annotated bean. That means, the same like `VaadinSession` means for `@VaadinSessionScope` annotated beans - it limits the lifecycle of the beans.

> Should "URL" rather be "Route"?

It might be a 'Route', but the `Route` is more a 'Vaadin-specific' term, whereas the 'URL' is more commonly used thing (and more clear).  

What I would propose to fix is to change the description of the picture to explain that those annotations are for bean classes.

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -29,8 +29,8 @@ In addition to standard Spring scopes, the https://vaadin.com/directory/componen
 ** `UIScope` for targeting beans to a browser window or tab opened by the user
 ** `RouteScope` for targeting beans to certain routing components pointed by the `RouteScopeOwner` qualifier
 
-.Vaadin Spring scopes overview
-image::images/spring-scopes.png[Vaadin Spring scopes overview and their binding to app, session, page and url]
+.Overview of Vaadin Spring scopes annotations for bean classes
```
pleku, line 32: This diagram is a tricky one to describe, since it has some example case that is not shown in code. I would go with:
```suggestion
.Overview of how Vaadin specific scopes relate to each other when used in navigation targets
```

articles/flow/integrations/spring/scopes.asciidoc
```diff
@@ -29,8 +29,8 @@ In addition to standard Spring scopes, the https://vaadin.com/directory/componen
...
+image::images/spring-scopes.png[Overview of Vaadin Spring scopes annotations for bean classes (including navigation targets) and their binding to app, session, page and url]
```
pleku, line 33:
```suggestion
image::images/spring-scopes.png[Overview of Vaadin specific scopes for bean classes (including navigation targets) and their binding to app, session, page and URL]
```

articles/flow/integrations/spring/scopes.asciidoc
pleku, line 32: Non-blocking comment: I would change the diagram to use `/PATH` instead of `URL` and maybe `BROWSER WINDOW/TAB`. But it is probably fine as-is. The comment about "url is more common than path" doesn't apply IMO since this is highly Vaadin context dependent and you need to understand routes + paths at this point.

pleku (changes requested):

articles/flow/integrations/spring/scopes.asciidoc
magi42: Notice that you can't have commas in the Alt text for images, you need to quote the text in ".

articles/flow/integrations/spring/scopes.asciidoc
magi42: You could avoid that very image captions, and rather explain it in an introducing sentence before the diagram.

magi42 (commented):
Maybe the diagram is good enough, even if there's some issues. Better put this forward.

articles/flow/integrations/spring/scopes.asciidoc
mshabarov: I've improved the image caption as suggested, but kept the caption, instead of separate sentence, because IMO the scopes are well-described in the existing sub sections. This diagram is just a visual representation what was written before.

articles/flow/integrations/spring/scopes.asciidoc
mshabarov: Fixed the alt text and wrapped it to `"`.

articles/flow/integrations/spring/scopes.asciidoc
mshabarov: Looking at this after a while, I agree that the `path` is better fit to Vaadin `route`, so renamed it.

pleku (approved):

mshabarov:
@magi42 `deploy-netlify` failure is a know issue with webpack/workbox incompatible versions:

```
9:28:51 AM: [ERROR] /opt/build/repo/node_modules/.pnpm/webpack-cli@3.3.11_webpack@4.46.0/node_modules/webpack-cli/bin/cli.js:93
9:28:51 AM: [ERROR] 				throw err;
9:28:51 AM: [ERROR] 				^
9:28:51 AM: [ERROR]
9:28:51 AM: [ERROR] Error: Cannot find module 'workbox-build/build/options/schema/webpack-generate-sw'
```

It has been fixed in V20 and will be fixed in V21 as soon as a new flow release will be done. So, to fix it, you need to use https://github.com/vaadin/platform/releases/tag/20.0.6

PR #534, opened 2021-06-07
fix: use timestamp to check if session is still alive

yuriy-fix:
`setPollInterval` will keep the session alive while user is online.
If the user or server is not going online for more than 40 minutes, the tab will be reloaded once the server is available.
Logic won't be executed if the included example is not on the page.

### DIFF
--- a/frontend/demo/example-resources.ts
+++ b/frontend/demo/example-resources.ts
@@ -9,33 +9,34 @@ window.__applyTheme = { applyTheme };
 // @ts-ignore
 import('all-flow-imports-or-empty').catch(() => {});
 
-function testHeartbeat() {
-  fetch(withPrefix('/vaadin/?v-r=heartbeat&v-uiId=0'), { method: 'POST' }).then((data) => {
-    const reloadTimestamp = localStorage.getItem('reloadTimestamp');
-    // Make sure session is still opened, otherwise reload the page
-    if (!data.ok && (!reloadTimestamp || Date.now() - parseInt(reloadTimestamp) > 5 * 60 * 1000)) {
-      // Save the previous reload timestamp to avoid reloading in less than 5 minutes
-      localStorage.setItem('reloadTimestamp', Date.now().toString());
-      location.reload();
-    }
-  });
-}
+let timestamp = Date.now();
+let interval: ReturnType<typeof setInterval>;
+const sessionTimeout = 30 * 60 * 1000;
 
-const initialListener = ((e: CustomEvent<string>) => {
-  window.removeEventListener('included-example-loaded', initialListener);
-  // Make sure flow example is upgraded and requested content from server
-  if (e.detail) {
-    customElements.whenDefined(e.detail).then(() => {
-      // Make sure server is up and running
-      fetch(withPrefix('/vaadin/index.html')).then((serverData) => {
-        if (serverData.ok) {
-          window.addEventListener('included-example-loaded', () => testHeartbeat());
-          testHeartbeat();
-        }
-      });
+const compareTimestamps = () => {
+  if (Date.now() - timestamp > sessionTimeout) {
+    fetch(withPrefix('/vaadin/index.html')).then((serverData) => {
+      if (serverData.ok) {
+        location.reload();
+      }
     });
   }
-}) as EventListener;
+};
+
+const updateTimestamps = () => {
+  timestamp = Date.now();
+  if (interval) {
+    clearInterval(interval);
+  }
+
+  interval = setInterval(compareTimestamps, sessionTimeout);
+};
+
+const initialListener = () => {
+  window.removeEventListener('update-timestamp', initialListener);
+  interval = setInterval(compareTimestamps, sessionTimeout);
+  window.addEventListener('update-timestamp', updateTimestamps);
+};
 
 // Examples are not available when session is expired. Logic prevents that by reloading the page.
-window.addEventListener('included-example-loaded', initialListener);
+window.addEventListener('update-timestamp', initialListener);
--- a/frontend/demo/example-resources.ts
+++ b/frontend/demo/example-resources.ts
@@ -1,5 +1,3 @@
-// @ts-ignore
-import { withPrefix } from 'gatsby';
 import './init-flow-namespace';
 import { applyTheme } from 'Frontend/generated/theme';
 // See webpack.config 'externals.Frontend/generated/theme'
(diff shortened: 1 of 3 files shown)


frontend/demo/example-resources.ts
```diff
@@ -9,33 +9,38 @@ window.__applyTheme = { applyTheme };
...
+    clearInterval(interval);
+  }
+
+  // Make sure interval starts at the time of UI polling the server
+  interval = setInterval(compareTimestamps, sessionTimeout);
+};
+
+// Examples are not available when session expires
+// Event is emitted when UI polls the server. It keeps session alive
+window.addEventListener('update-timestamp', updateTimestamps);
```
tomivirkki, line 46: Will this event be dispatched when you end up in this state? https://github.com/vaadin/docs-app/issues/273#issuecomment-853597932

src/main/java/com/vaadin/demo/DemoExporter.java
```diff
@@ -25,5 +30,16 @@ public DemoExporter() {
 
   @Override
   protected void configureInstance(final WebComponent<T> webComponent, final T demo) {
+    if (!initializedPoll) {
+      UI.getCurrent().setPollInterval((int) Duration.ofMinutes(20).toMillis());
+      UI.getCurrent().addPollListener(e -> emitUpdateTimestamp());
```
tomivirkki, line 35: These listeners keep adding up as you visit more examples and you end up with multiple "update-timestamp" events firing at once.

frontend/demo/example-resources.ts
```diff
@@ -9,33 +9,38 @@ window.__applyTheme = { applyTheme };
...
-    const reloadTimestamp = localStorage.getItem('reloadTimestamp');
-    // Make sure session is still opened, otherwise reload the page
-    if (!data.ok && (!reloadTimestamp || Date.now() - parseInt(reloadTimestamp) > 5 * 60 * 1000)) {
-      // Save the previous reload timestamp to avoid reloading in less than 5 minutes
-      localStorage.setItem('reloadTimestamp', Date.now().toString());
-      location.reload();
-    }
-  });
-}
+// Stores the last time UI polls the server
```
tomivirkki, line 12: Can you move this logic to a separate file that you import here.

tomivirkki (changes requested):

frontend/demo/example-resources.ts
yuriy-fix: I wasn't able to get into that exact state.
However, this logic should prevent getting there when the server and user are online. If one of those is offline, the reloading should happen at the point they are getting back online (if the session is presumably destroyed = after timeout).

src/main/java/com/vaadin/demo/DemoExporter.java
yuriy-fix: Good catch! Checking for the existing poll interval now.
Visiting other examples will still trigger the event as it's considered to be a round-trip that keeps the session alive and the timestamp should be updated 🤔

frontend/demo/example-resources.ts
yuriy-fix: Moved it.

frontend/demo/session-verification.ts
```diff
@@ -0,0 +1,38 @@
...
+
+// Stores the last time UI polls the server
+let timestamp: number;
+let interval: ReturnType<typeof setInterval>;
+const sessionTimeout = 20 * 60 * 1000;
+
+const compareTimestamps = () => {
+  // Event could be emitted with a delay up to 2 seconds after the poll
+  // Check if the timestamp is not updated after two intervals
+  if (Date.now() - timestamp > 2 * sessionTimeout) {
```
tomivirkki, line 12: What happens if the user re-opens the page after `1.5 * sessionTimeout`?

frontend/demo/session-verification.ts
```diff
@@ -0,0 +1,38 @@
...
+    // Make sure we are not reloading the page before the server becomes available
+    fetch(withPrefix('/vaadin/index.html')).then((serverData) => {
```
tomivirkki, line 14: `/vaadin/index.html` [seems to get redirected](https://vaadin.com/docs/latest/vaadin/index.html) to `/vaadin/index` so better use the latter.

frontend/demo/session-verification.ts
```diff
@@ -0,0 +1,38 @@
+// @ts-ignore
+import { withPrefix } from 'gatsby';
+
+// Stores the last time UI polls the server
+let timestamp: number;
+let interval: ReturnType<typeof setInterval>;
+const sessionTimeout = 20 * 60 * 1000;
```
tomivirkki, line 7: This is a magic number. Could it be delivered in the event payload?

src/main/java/com/vaadin/demo/DemoExporter.java
```diff
@@ -25,5 +28,15 @@ public DemoExporter() {
 
   @Override
   protected void configureInstance(final WebComponent<T> webComponent, final T demo) {
+    if (UI.getCurrent().getPollInterval() == -1) {
+      UI.getCurrent().setPollInterval((int) Duration.ofMinutes(20).toMillis());
```
tomivirkki, line 32: This is a magic number. Could it be obtained programmatically instead?

yuriy-fix: It's semi-random (`maxInectiveInterval` - 10 minutes) now: poll should happen at some point before the max inactive interval is reached. Cannot do it at the same time as there could be some delay within roundtrips.

frontend/demo/session-verification.ts
yuriy-fix: Getting value from the event now.

frontend/demo/session-verification.ts
yuriy-fix: Changed.

frontend/demo/session-verification.ts
yuriy-fix: Nothing as `pollInterval` is supposed to keep the session alive.

frontend/demo/example-resources.ts
```diff
@@ -9,33 +7,5 @@ window.__applyTheme = { applyTheme };
 // @ts-ignore
 import('all-flow-imports-or-empty').catch(() => {});
 
-function testHeartbeat() {
-  fetch(withPrefix('/vaadin/?v-r=heartbeat&v-uiId=0'), { method: 'POST' }).then((data) => {
-    const reloadTimestamp = localStorage.getItem('reloadTimestamp');
-    // Make sure session is still opened, otherwise reload the page
-    if (!data.ok && (!reloadTimestamp || Date.now() - parseInt(reloadTimestamp) > 5 * 60 * 1000)) {
-      // Save the previous reload timestamp to avoid reloading in less than 5 minutes
-      localStorage.setItem('reloadTimestamp', Date.now().toString());
```
tomivirkki, line 18: Noticed you decided to remove the "safety net". I assume it was intentional?

tomivirkki (approved):

frontend/demo/example-resources.ts
yuriy-fix: Yes, with the new approach we are not relying on heartbeats checks. The only case it will get into the reloading loop is if we would provide incorrect `sessionTimeout`.

Issue #437, opened 2021-05-05
document the virtual-list component (TypeScript)


vlukashov:
TypeScript examples for https://github.com/vaadin/docs/issues/436

The docs page _may_ optionally include a section comparing the vaadin virtual list component to other virtual lists out there, but this is not a part of the _required_ scope for this task.

Compared to other alternatives, vaadin virtual list:
 - has a better UX for scrolling (doen not immediately scrolls the document after reaching the end of the virtual content)
 - has a better support for screen readers
 - unlimited items

Issue #436, opened 2021-05-05
document the VirtualList component (Java)


vlukashov:
This documentation should be available in the list of all DS components, at https://vaadin.com/docs/latest/ds/components
It should have a similar structure as all other component pages (i.e. examples and API). The emphasiz should be on that it's a _virtual_ list, i.e. capable of displaying large number of items without having adverse performance impact.

PR #538, opened 2021-06-10
docs: virtual-list docs and examples

tomivirkki:
Closes https://github.com/vaadin/docs/issues/436
Closes https://github.com/vaadin/docs/issues/437


articles/ds/components/virtual-list/index.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+  - https://github.com/vaadin/flow-components/releases/tag/{moduleMavenVersion:com.vaadin:vaadin-virtual-list-flow}[Flow {moduleMavenVersion:com.vaadin:vaadin-virtual-list-flow}]
+  - https://github.com/vaadin/web-components/releases/tag/v{moduleNpmVersion:vaadin-virtual-list}[Web Component {moduleNpmVersion:vaadin-virtual-list}]
+---
+= Virtual List
+
+:components-path-prefix: ../
+
+// tag::description[]
+Virtual list displays a vertically aligned set of items inside a scrollable container.
+Each item is rendered per need basis as the user scrolls the list, which makes the component especially suitable for displaying long item lists without sacrificing performance.
```
magi42, line 15:
```suggestion
Each item is rendered on the fly as the user scrolls the list, which makes the component especially suitable for displaying long lists of items without sacrificing performance.
```
I wrote "rendered" here, but should it rather be "generated"? Is the actual data generated or just rendering?

articles/ds/components/virtual-list/index.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+// end::description[]
+
+To use the component, you need to assign it a set of data items and a renderer that's used for rendering each individual data item.
+The height of an item is determined by its content and can change dynamically.
+
+[.example]
```
magi42, line 21: The example doesn't include the sources for the data source, and it is not trivial to figure out what they should be like.

articles/ds/components/virtual-list/index.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+= Virtual List
+
+:components-path-prefix: ../
+
+// tag::description[]
+Virtual list displays a vertically aligned set of items inside a scrollable container.
+Each item is rendered per need basis as the user scrolls the list, which makes the component especially suitable for displaying long item lists without sacrificing performance.
+// end::description[]
+
+To use the component, you need to assign it a set of data items and a renderer that's used for rendering each individual data item.
```
magi42, line 18:
```suggestion
To use the component, you need to assign it a set of data items and a _renderer_ that is used for rendering each individual data item.
```

magi42 (changes requested):
Note the question in the first suggestion.

The file was very short so is it still a draft? If so, please mark it as a draft.

articles/ds/components/virtual-list/index.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+title: Virtual List
+layout: tabbed-page
+tab-title: Usage
+page-links:
+  - https://github.com/vaadin/flow-components/releases/tag/{moduleMavenVersion:com.vaadin:vaadin-virtual-list-flow}[Flow {moduleMavenVersion:com.vaadin:vaadin-virtual-list-flow}]
+  - https://github.com/vaadin/web-components/releases/tag/v{moduleNpmVersion:vaadin-virtual-list}[Web Component {moduleNpmVersion:vaadin-virtual-list}]
+---
+= Virtual List
+
+:components-path-prefix: ../
```
jouni, line 11: Is this attribute used anywhere on this page?

articles/ds/components/virtual-list/index.asciidoc
tomivirkki: Rendered is better since that's what it's about

articles/ds/components/virtual-list/index.asciidoc
tomivirkki: Aligned with other similar examples and included the Person.java / Person.ts files

articles/ds/components/index.asciidoc
```diff
@@ -177,6 +177,11 @@ image::{components-path-prefix}scroller/scroller.png["", opts=inline, role="icon
 include::scroller/index.asciidoc[tag=description]
 <<{components-path-prefix}scroller#,See Scroller>>
 
+[.card]
+=== Virtual List
+image::{components-path-prefix}virtual-list/virtual-list.png["", role="icon screenshot"]
```
yuriy-fix, line 182: nit: all other examples seems to include `opts=inline`. Should it be added here as well? 🤔

frontend/demo/component/virtuallist/virtual-list-basic.ts
```diff
@@ -0,0 +1,96 @@
...
+import '@vaadin/vaadin-details';
+import '@vaadin/vaadin-ordered-layout/vaadin-horizontal-layout';
+import '@vaadin/vaadin-ordered-layout/vaadin-vertical-layout';
+import { getPeople } from 'Frontend/demo/domain/DataService';
+import Person from 'Frontend/generated/com/vaadin/demo/domain/Person';
+import { applyTheme } from 'Frontend/generated/theme';
+import type { VirtualListRenderer } from '@vaadin/vaadin-virtual-list';
+import { DetailsElement } from '@vaadin/vaadin-details';
+
+@customElement('tree-grid-basic')
```
yuriy-fix, line 18:
```suggestion
@customElement('virtual-list-basic')
```

frontend/demo/component/virtuallist/virtual-list-basic.ts
```diff
@@ -0,0 +1,96 @@
...
+        >
+        </vaadin-avatar>
+
+        <vaadin-vertical-layout>
+          <b>${person.firstName} ${person.lastName}</b>
+          <span>${person.profession}</span>
+
+          <vaadin-details
+            .opened="${this.expandedPeople.has(person)}"
+            @click="${(e: CustomEvent) => {
```
yuriy-fix, line 64: nit: we can use `MouseEvent` over here.

articles/ds/components/index.asciidoc
tomivirkki: Good catch. Not sure how I missed it here. Done.

frontend/demo/component/virtuallist/virtual-list-basic.ts
tomivirkki: Done

yuriy-fix (approved):

jouni:
@yuriy-fix, always update the Java/TypeScript/Asciidoc labels to indicate which of those your approval concerns. I assume in this case it's for the Java and TypeScript source code. If yes, replace the `Java` and `TypeScript` labels with `Java OK` and `TypeScript OK`.

articles/ds/components/index.asciidoc
magi42: OK, now the Person is there, but shouldn't the data service be as well? I understand that the fake data service is rather unusual for a business case, so for the sake of abstraction we might not want to show it, but it feels like there's a missing link there.

tomivirkki: The [fake data service ](https://github.com/vaadin/docs/blob/latest/src/main/java/com/vaadin/demo/domain/DataService.java) isn't included in _any_ of the examples. Before it could be included in public examples, it would need some preparation work including documentation and clean up and that falls out of the scope of this PR.

magi42: OK yeah, no worries. The reason why I'm thinking about that is a very frustrated hour a year ago trying to use TreeGrid, where the documentation just used the data source that defined the hierachy, but it wasn't described anywhere. But OK, this was just something to think about.

tomivirkki:
@magi42 The PR still blocked due to the change request

magi42 (approved):
Odd, I didn't notice any pending change request, or do I actually need to approve... Ok I'll approve...

Issue #439, opened 2021-05-05
Update the 'Icons' docs page to show the new vaadin-icon component


vlukashov:
The [Foundation / Icons](https://vaadin.com/docs/latest/ds/foundation/icons) page should be updated to tell about the new `vaadin-icon` web component instead of `iron-icon` with the `vaadin` and `lumo` iconsets.

Also, if there are any changes in the API and / or usage recommendations for the `Icon` Flow component, these changes should be reflected in this docs page as well.

PR #546, opened 2021-06-14
docs: update ds/foundation/icons to use vaadin-icon

tomivirkki:
Closes https://github.com/vaadin/docs/issues/439

### DIFF
--- a/articles/ds/foundation/icons/index.asciidoc
+++ b/articles/ds/foundation/icons/index.asciidoc
@@ -27,7 +27,7 @@ In server-side views (Flow/Java):
 
 [source,java]
 ----
-Icon icon = new Icon('lumo:photo');
+Icon icon = new Icon("lumo", "photo");
 ----
 
 In client-side views (Fusion/TypeScript):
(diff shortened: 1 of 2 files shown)


articles/ds/foundation/icons/index.asciidoc
```diff
@@ -36,19 +36,17 @@ In client-side views (Fusion/TypeScript):
...
 @customElement('photo-icon-example')
 export class PhotoIconExample extends LitElement {
   render() {
-    return html`<iron-icon icon="lumo:photo"></iron-icon>`;
+    return html`<vaadin-icon icon="lumo:photo"></vaadin-icon>`;
   }
 }
 ----
 
-include::../_shared.asciidoc[tag=icon-usage]
```
tomivirkki, line 49: There's currently no bundle import for vaadin-icon + iconset, so instead of updating, removed the include for now

jouni (approved):
Nice catch that there isn't even a "vaadin:photo" icon 😅

PR #558, opened 2021-06-15
feat: Add a login/logout example for View-based access control chapter

### LINKED ISSUE #548: Provide a logout code example in Security Best Practice chapter
Add a basic example how to:
1. Connect the LoginView with the ViewAccessChecker, i.e. how to pass the entered credentials from login form into ViewAccessChecker.
2. Implement a logout functionality for non-Spring applications in this chapter https://vaadin.com/docs/latest/flow/security/best-practices/#view-based-access-control.

This can be added as an example:

```java
public class LoginView extends VerticalLayout implements BeforeEnterObserver,
        ComponentEventListener<AbstractLogin.LoginEvent> {

    private SecurityService securityService;

    // Other class fields omitted

    @Inject
    public LoginView(SecurityService securityService) {
        this.securityService = securityService;

        login.addLoginListener(this);

        // Other code omitted
    }

    @Override
    public void onComponentEvent(AbstractLogin.LoginEvent loginEvent) {
        boolean authenticate = securityService.authenticate(
                loginEvent.getUsername(), loginEvent.getPassword());
        if (authenticate) {
            UI.getCurrent().navigate(HomeView.class);
        } else {
            login.setError(true);
        }
    }
}

public class SecurityService {

    public boolean isAuthenticated() {
        return VaadinServletRequest.getCurrent() != null &&
               VaadinServletRequest.getCurrent().getUserPrincipal() != null;
    }

    public boolean authenticate(String username, String password) {
        if (VaadinServletRequest.getCurrent() != null) {
            VaadinServletRequest request = VaadinServletRequest.getCurrent();
            try {
                request.login(username, password);
                return true;
            } catch (ServletException e) {
                return false;
            }
        }
        return false;
    }

    public void logout() {
        VaadinSession.getCurrent().getSession().invalidate();
        UI.getCurrent().getPage().setLocation("/");
    }

}
```

mshabarov:
## Description

Add an example how to pass Login Form credentials to the security context and how to implement a logout capability.

Fixes #548

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [x] The issue is created in the corresponding repository and I have referenced it.

- [x] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

### DIFF
--- a/articles/flow/security/best-practices.asciidoc
+++ b/articles/flow/security/best-practices.asciidoc
@@ -46,6 +46,7 @@ See <<{articles}/flow/integrations/spring/view-based-access-control#, View-Based
 To enable and use this mechanism, the following should be added to a Vaadin Flow application (if it does not exist already):
 
 - A login view.
+- Logout capability.
 - A `VaadinServiceInitListener` that adds `ViewAccessChecker` as the `BeforeEnterListener` of all UIs.
 - A Service Provider under `META-INF/services` to load above the custom [classname]#ServiceInitListener# via the Java SPI loading mechanism.
 
@@ -58,28 +59,45 @@ It serves to redirect anonymous users to that page before granting access to vie
 ----
 @Route("login")
 @PageTitle("Login")
-public class LoginView extends VerticalLayout implements BeforeEnterObserver {
+public class LoginView extends VerticalLayout implements BeforeEnterObserver,
+        ComponentEventListener<AbstractLogin.LoginEvent> {
+
+    private static final String LOGOUT_SUCCESS_URL = "/";
 
     private LoginForm login = new LoginForm();
+    private SecurityService securityService;
 
-    public LoginView() {
+    @Inject
+    public LoginView(SecurityService securityService) {
+        this.securityService = securityService;
         addClassName("login-view");
         setSizeFull();
 
         setJustifyContentMode(JustifyContentMode.CENTER);
         setAlignItems(Alignment.CENTER);
 
-        login.setAction("login");
+        login.addLoginListener(this);
 
         add(new H1("Test Application"), login);
     }
 
     @Override
     public void beforeEnter(BeforeEnterEvent beforeEnterEvent) {
-        if(beforeEnterEvent.getLocation()
+        if (beforeEnterEvent.getLocation()
             .getQueryParameters()
             .getParameters()
             .containsKey("error")) {
+            login.setError(true);
+        }
+    }
+
+    @Override
+    public void onComponentEvent(AbstractLogin.LoginEvent loginEvent) {
+        boolean authenticated = securityService.authenticate(
+                loginEvent.getUsername(), loginEvent.getPassword());
+        if (authenticated) {
+            UI.getCurrent().getPage().setLocation(LOGOUT_SUCCESS_URL);
+        } else {
             login.setError(true);
         }
     }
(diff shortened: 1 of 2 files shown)


articles/flow/security/best-practices.asciidoc
```diff
@@ -58,37 +59,125 @@ It serves to redirect anonymous users to that page before granting access to vie
...
+            login.setError(true);
+        }
+    }
 }
 ----
 
 In this example, Vaadin's Login Form component is used for the sake of brevity.
 However, feel free to implement your own login view, if you wish.
 
+=== Logout capability
```
mshabarov, line 110:
```suggestion
=== Logout Capability
```

articles/flow/security/best-practices.asciidoc
```diff
@@ -58,37 +59,125 @@ It serves to redirect anonymous users to that page before granting access to vie
...
+}
+----
+
+The way how to get the authenticated user and how to logout the user may vary from application to application.
+Here is a basic example how to do that with Vaadin API:
+
+.[filename]#SecurityService.java#
+[source,java]
+----
+@VaadinServiceScoped
```
pleku, line 150: How is this supposed to work ? I don't think there is a mention of CDI anywhere. This will not work with standard Java DI.

Maybe you should store the user details to the `VaadinSession` with `setAttribute(Class<UserInfo>, UserInfo)` instead ?

pleku (changes requested):

articles/flow/security/best-practices.asciidoc
```diff
@@ -58,37 +59,125 @@ It serves to redirect anonymous users to that page before granting access to vie
...
+public class LoginView extends VerticalLayout implements BeforeEnterObserver,
+        ComponentEventListener<AbstractLogin.LoginEvent> {
+
+    private static final String LOGOUT_SUCCESS_URL = "/";
 
     private LoginForm login = new LoginForm();
+    private SecurityService securityService;
 
-    public LoginView() {
+    @Inject
```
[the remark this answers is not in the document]
mshabarov: Not sure I understand the comment. `@Inject` works here, I tested with demo project. And I believe `@Inject` is a pretty much basic thing and should be knows for devs who start trying security features with Java EE.

articles/flow/security/best-practices.asciidoc
```diff
@@ -58,37 +59,125 @@ It serves to redirect anonymous users to that page before granting access to vie
...
+    public LoginView(SecurityService securityService) {
```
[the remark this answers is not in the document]
mshabarov: Yep, it's better to show it first

articles/flow/security/best-practices.asciidoc
mshabarov: Well, to avoid the CDI here in these examples, I could make this service an util class instead with static methods.
`ViewAccessChecker` relies on `HttpServletRequest::isUserInRole` by default, thus, I don't see how putting the attribute to a session will then let `isUserInRole` to work.

articles/flow/security/best-practices.asciidoc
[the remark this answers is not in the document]
pleku: Yes, `@Inject` works but it may not be obvious for everyone.

But since you added `@VaadinServiceScoped` it this actually doesn't work or would not work like intended

articles/flow/security/best-practices.asciidoc
pleku: Right - well not sure which is better. One thing to note is that `VaadinServletReqquest.getCurrent` will not work for any background thread so the pattern would be a bit fragile if it is just a static utility class. A lot of explaining would be needed to be done to explain how it is not actually recommended

pleku: Oh well, that applies to the whole `ViewAccessChecker` and that should be documented here somewhere too. Probably missed out from the initial code review but I think documenting this is enough

mshabarov: > One thing to note is that VaadinServletReqquest.getCurrent will not work for any background thread so the pattern would be a bit fragile if it is just a static utility class

ViewAccessChecker itself won't work if no request is available: https://github.com/vaadin/flow/blob/d17359acea9f0ef42c2f0d751bb92028ae047d0d/flow-server/src/main/java/com/vaadin/flow/server/auth/ViewAccessChecker.java#L128-L133

I propose to use an utility class and add comments/explanations there that it works like this. Artur suggests using a Shiro, but for the purpose of this chapter it will extra. But mention it might be fine.

pleku (approved):

articles/flow/security/best-practices.asciidoc
```diff
@@ -46,9 +46,58 @@ See <<{articles}/flow/integrations/spring/view-based-access-control#, View-Based
...
+The way how to get the authenticated user and how to login/logout the user may vary from application to application.
+Vaadin View-based access control implementation uses [methodname]#HttpServletRequest::getUserPrincipal()# and [methodname]#HttpServletRequest.isUserInRole(String)# to check whether the currently authenticated user has access to the view.
+Thus, you have to use a solutions, which put the user information into request.
+The example below uses currently active Vaadin servlet request and shows how to:
+
+- Check whether the user is currently authenticated to the application or not.
+The request is not available in background threads, so this method shows authentication state only in Vaadin request processing threads, otherwise it would always return `false`.
+- Authenticate the user with the given credentials.
+When this method is called from background threads, the authentication would always fail.
+- Logout the user by invalidating the http session and redirecting to the home page.
```
magi42, line 64:
```suggestion
=== Handle User's Login and Logout

There are many ways to handle login and logout, and how to get the authenticated user for a request.
The View-based access control implementation in Vaadin uses [methodname]#getUserPrincipal()# and [methodname]#isUserInRole(String)# in [classname]#HttpServletRequest# to check whether the currently authenticated user has access to the view.
Therefore, you need a way to put the user information into the request.

The example below uses the currently active Vaadin servlet request.
It shows how to:

- Check whether the user is currently authenticated to the application or not.
The request is not available in background threads, so the [methodname]#isAuthenticated()# method shows authentication state only in Vaadin request processing threads, otherwise it would always return `false`.
- Authenticate the user with the given credentials.
The [methodname]#authenticate()# method fails if is called from a background thread.
- Logout the user by invalidating the HTTP session and redirecting to the home page.
```

articles/flow/security/best-practices.asciidoc
```diff
@@ -46,9 +46,58 @@ See <<{articles}/flow/integrations/spring/view-based-access-control#, View-Based
...
+                request.login(username, password);
+                return true;
+            } catch (ServletException e) {
+                // login exception handle code omitted
+                return false;
+            }
+        }
+        // This is in a background thread and we cannot access the request to
+        // log in the user
+        return false;
```
magi42, line 91: Consider returning false in the beginning if getCurrent() == null. Maybe more clear structure.

articles/flow/security/best-practices.asciidoc
```diff
@@ -69,26 +121,66 @@ public class LoginView extends VerticalLayout implements BeforeEnterObserver {
...
 }
 ----
 
 In this example, Vaadin's Login Form component is used for the sake of brevity.
 However, feel free to implement your own login view, if you wish.
 
+=== Logout Capability
+
+For handling logout in web applications, one typically uses a logout button.
+Here is a basic implementation of logout button shown on the main layout's header:
```
magi42, line 158:
```suggestion
You typically let the user to log out by using a logout button.

The following example shows a basic implementation of a logout button shown on the header of the main layout:
```

magi42 (changes requested):

PR #563, opened 2021-06-16
Update nullability info

Lodin:
Update nullability info according to the newest updates

### DIFF
--- a/articles/fusion/advanced/endpoints-generator.asciidoc
+++ b/articles/fusion/advanced/endpoints-generator.asciidoc
@@ -72,7 +72,7 @@ public class UserEndpoints {
 
 NOTE: For more information about type mapping between Java and TypeScript, please refer to <<type-conversion#, type conversion page>>.
 
-=== Nullability of types
+=== Type Nullability
 
 Types can be either nullable (optional) or non-nullable (required).
 By default, types are mapped and generated using the Java rules:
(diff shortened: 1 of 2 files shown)


articles/fusion/advanced/endpoints-generator.asciidoc
```diff
@@ -72,10 +72,17 @@ public class UserEndpoints {
...
-To make it non-nullable (required), you need to put a `@Nonnull` annotation on it.
+Types can be either nullable (optional) or non-nullable (required).
+By default, types are mapped and generated using the Java rules:
+
+- Any primitive type (e.g., `int`) is non-nullable.
+- Any reference type (e.g., `String` or `Integer`) is nullable.
+- A collection accepts `null` unless the collection item type is primitive.
+- A map accepts `null` unless the collection item type is primitive.
+
+To make any type non-nullable, you need to put a `@Nonnull` annotation on it.
```
magi42, line 85:
```suggestion
Types are either _nullable_ (optional) or _non-nullable_ (required).
By default, types are mapped and generated using the Java rules:

- Any primitive type, such as `int`, is non-nullable.
- Any reference type, such as `String` or `Integer`, is nullable.
- A collection accepts `null` unless the collection item type is primitive.
- A map accepts `null` unless the collection item type is primitive.

To make a type non-nullable, you need to put a `@Nonnull` annotation on it.
```

magi42 (approved):

magi42:
The text looks OK to me now, if someone could check the TS part.

haijian-vaadin (approved):

PR #569, opened 2021-06-18
Update utility-classes.asciidoc

anezthes:
Previews for most, but not all, classes.

I strongly dislike the look of the previews, but I don't have time to fix/design it at the moment. We'll need to do this properly at some point.

Also, we should give credit to Tailwind at the beginning of the doc, as it's heavily inspired by their work. (I mean, it started with Business App Starter before I at least knew what Tailwind was, but still.)

### DIFF
--- a/articles/ds/foundation/utility-classes.asciidoc
+++ b/articles/ds/foundation/utility-classes.asciidoc
@@ -72,31 +72,48 @@ All utility classes are listed below.
 <style>
 .preview {
   float: none;
-  height: 98px;
-  padding: 0;
-  width: 98px;
 }
 .preview:not([class*=rounded]) {
   border-radius: var(--docs-border-radius-m);
+}
+.preview:not([class*=" h-"]) {
+  height: 98px;
+}
+.preview:not([class*=" w-"]) {
+  width: 98px;
 }
 .preview.grid {
   padding: var(--lumo-space-xs);
   width: 196px;
 }
+/* This class unfortunately overrides some of the utility classes.
+   The examples on this page using .item should be re-written using
+   the utility classes. */
 .item {
   align-items: center;
   background-color: var(--docs-body-text-color);
   border-radius: var(--docs-border-radius-s);
   color: var(--docs-surface-color-1);
-  display: flex;
   font-size: 0.625rem;
   justify-content: center;
-  margin: var(--docs-space-xs);
   min-width: 1.5rem;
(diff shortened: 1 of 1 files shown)


jouni:
Would be great if you could disable Vale spell checking whenever you use lorem ipsum as a placeholder text.

Add `pass:[<!-- vale Vale.Spelling = NO -->]` before the table, and `pass:[<!-- vale Vale.Spelling = YES -->]` after it.

articles/ds/foundation/utility-classes.asciidoc
```diff
@@ -22,39 +22,20 @@ While the classes can be used with Vaadin's web components they are primarily in
...
 
-Then import the classes in your application's `styles.css`.
-
-.`frontend/themes/myapp/styles.css`
-[source,css]
-----
-@import 'lumo-css-framework/all-classes.css';
-----
-
 == Classes
```
DiegoCardoso, line 39: I am not entirely satisfied with my changes here on the instructions to use the utility classes. Could someone else suggest an enhancement to it? ping @jouni @anezthes

articles/ds/foundation/utility-classes.asciidoc
```diff
@@ -22,39 +22,20 @@ While the classes can be used with Vaadin's web components they are primarily in
...
 == Usage
 
-The utility classes can be used with both Flow (Java) and Fusion (TypeScript).
-
-Install the package with the following command:
+The utility classes can be used with both Flow (Java) and Fusion (TypeScript) and they are part of the Vaadin Lumo Styles package.
 
-[source,terminal]
-npm i lumo-css-framework
+To use the classes, make sure that `"utility"` is added to the `"lumoImports"` entry in your application's `theme.json`.
```
rolfsmeds, line 29: How about:

To use the utility classes, make sure that the `"utility"` module is included in the `"lumoImports"` list in your application theme's [configuration](https://vaadin.com/docs/latest/ds/customization/custom-theme-configuration) `theme.json`.

Alternatively, if not using a custom theme folder, you can import the utility class module manually: [followed by Flow and Fusion ways of importing a style module without using theme folder, as shown in test plan]

DiegoCardoso: Uhmm, right. I will apply the suggestions, thanks.

rolfsmeds (approved):

DiegoCardoso (approved):

PR #571, opened 2021-06-18
Add RTE flow examples

yuriy-fix:


src/main/java/com/vaadin/demo/component/richtexteditor/RichTextEditorSetGetValue.java
```diff
@@ -0,0 +1,29 @@
...
+
+  public RichTextEditorSetGetValue() {
+    // tag::snippet[]
+    TextArea textArea = new TextArea("Html Value", "Type html string here to set it as value to the Rich Text Editor above...");
+    textArea.setWidthFull();
+
+    RichTextEditor rte = new RichTextEditor();
+    rte.getStyle().set("max-height", "400px");
+
+    rte.asHtml().addValueChangeListener(e -> textArea.setValue(e.getValue()));
```
DiegoCardoso, line 21: There's something strange happening here. 

If I start typing on RTE, everything works, but if I at any point use the toolbar, then the focus starts to move.

You can see on the example below that from the point I clicked on the bold button, the RTE started to misbehave. You can also notice that the textarea starts receiving the change value indicating that a change event is happening on RTE.


https://user-images.githubusercontent.com/262432/122787408-fa9a4c00-d2bd-11eb-9440-83fc6203cba2.mp4

DiegoCardoso (changes requested):

src/main/java/com/vaadin/demo/component/richtexteditor/RichTextEditorSetGetValue.java
yuriy-fix: Good catch!
Was happening because interaction with toolbar emits `change` event and textarea was resetting value on rte. Fixed by checking if the html value is still the same.

DiegoCardoso (approved):
LGTM 💯

PR #572, opened 2021-06-21
docs: make focus outline always visible in grid cell focus example

sissbruecker:
## Description

Modifies the grid cell focus example to always show the focus outline for demonstration purposes, and adds a hint to the asciidoc about differences to the default behaviour.

Based on feedback from DX testing of the grid cell focus event.

### DIFF
--- a/articles/ds/components/grid/index.asciidoc
+++ b/articles/ds/components/grid/index.asciidoc
@@ -662,7 +662,7 @@ The following keyboard shortcuts are available:
 The cell focus event can be used to get notified when the user changes focus between cells.
 
 By default, the focus outline is only visible when using keyboard navigation.
-For demonstration purposes the example below uses custom styles to also show the focus outline when clicking on cells.
+For demonstration purposes, the example below uses custom styles to also show the focus outline when clicking on cells.
 
 [.example]
 --
--- a/articles/ds/components/grid/index.asciidoc
+++ b/articles/ds/components/grid/index.asciidoc
@@ -661,8 +661,8 @@ The following keyboard shortcuts are available:
 
 The cell focus event can be used to get notified when the user changes focus between cells.
 
-The focus outline is only visible when using keyboard navigation.
-In the example below, try clicking on a cell and then use one of the arrow keys.
+By default, the focus outline is only visible when using keyboard navigation.
+For demonstration purposes the example below uses custom styles to also show the focus outline when clicking on cells.
 
 [.example]
 --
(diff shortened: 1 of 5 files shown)


frontend/themes/docs/components/vaadin-grid-cell-focus.css
```diff
@@ -0,0 +1,12 @@
...
+    content: '';
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    pointer-events: none;
+    box-shadow: inset 0 0 0 2px var(--lumo-primary-color-50pct);
+}
+/* end::snippet[] */
```
jouni, line 12:
```suggestion
:host([theme~='force-focus-outline']) [part~='cell']:focus::before {
    content: '';
    position: absolute;
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
    pointer-events: none;
    box-shadow: inset 0 0 0 2px var(--lumo-primary-color-50pct);
}
```

articles/ds/components/grid/index.asciidoc
```diff
@@ -677,6 +677,11 @@ include::{root}/frontend/demo/component/grid/grid-cell-focus.ts[render,tags=snip
 include::{root}/src/main/java/com/vaadin/demo/component/grid/GridCellFocus.java[render,tags=snippet,indent=0,group=Java]
 ----
 
+[source,css]
+----
+include::{root}/frontend/themes/docs/components/vaadin-grid-cell-focus.css[render,tags=snippet,indent=0,group=CSS]
```
jouni, line 682:
```suggestion
include::{root}/frontend/themes/docs/components/vaadin-grid-cell-focus.css[]
```

jouni (changes requested):
No need for tagged regions if the all the contents of the file are shown. We could use tags if we put the CSS of multiple examples in one file, but I think having individual files per example makes it slightly less confusing for the reader.

No need for `render` in any other source code includes than TS and Java examples that export a web component that can be rendered. CSS files obviously don't do that.

Any source code includes that apply for both TS and Java examples should not define a `group` attribute, so that they are shown under both TS and Java groups. Think of the groups as a collection if source files that are needed to accomplish the rendered example, not a grouping by programming language. It’s more about Fusion/Flow than TS/Java, but we are for now using the more easily recognizable terms.

articles/ds/components/grid/index.asciidoc
```diff
@@ -661,8 +661,8 @@ The following keyboard shortcuts are available:
 
 The cell focus event can be used to get notified when the user changes focus between cells.
 
-The focus outline is only visible when using keyboard navigation.
-In the example below, try clicking on a cell and then use one of the arrow keys.
+By default, the focus outline is only visible when using keyboard navigation.
+For demonstration purposes the example below uses custom styles to also show the focus outline when clicking on cells.
```
fhdhsni, line 665: `For demonstration purposes, ...`?

sissbruecker: Done

fhdhsni (approved):

jouni (approved):

PR #591, opened 2021-06-25
A more proper style guide

magi42:
We need a more proper style guide. We'll have a "quick guide" and a "comprehensive guide" (actual titles to be decided).


jouni:
Build is failing because of `error Broken internal reference #a_to_z.proper_nouns in guide/contributing-docs/quick-style-guide.asciidoc`

jouni:
Looks like the last commit from @dougjewsbury reverted some of the previous commits (like my CSS style updates).

PR #590, opened 2021-06-25
feat: replace iron-icon with vaadin-icon

vursen:
## Description

This PR replaces deprecated `iron-icon` with `vaadin-icon` which is considered as the right way to use icons from now on.

Fixes https://github.com/vaadin/web-components/issues/2112 (issue)

## Type of change

- [x] Feature

## Checklist

- [x] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [x] I have added a description following the guideline.
- [x] The issue is created in the corresponding repository and I have referenced it.
- [x] I have added tests to ensure my change is effective and works as intended.
- [x] New and existing tests are passing locally with my change.
- [x] I have performed self-review and corrected misspellings.

PR #596, opened 2021-06-28
docs: add first set of Flow grid examples

sissbruecker:
## Description

Adds Flow examples for Grid for the following sections:
- Content
- Dynamic height
- Selection
- Columns

Regarding reviewing the changes to the Java files: There already were dummy files for the Java examples, all of which had the same content. I tried removing and readding them, however Github still shows them as changed files. When reviewing I would suggest to use the split view or view the files individually.
For the upcoming examples I already removed the other dummy files, so we will not have this problem again.


frontend/demo/component/grid/grid-basic.ts
```diff
@@ -1,5 +1,6 @@
 import 'Frontend/demo/init'; // hidden-source-line
 import '@vaadin/flow-frontend/gridConnector.js'; // hidden-source-line (Grid's connector)
+import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js'; // hidden-source-line (Legacy template renderer)
```
sissbruecker, line 3: Added this import for all of the Grids' TS examples, otherwise the column headers would not render.

The issue here is that the Flow component needs the template renderer and in theory imports it through the `@JsImport` annotation. However in practice the docs-app seems to have a problem in its setup so that the `@JsImport` annotation do not work correctly. As such all additional JS modules that Flow components need, must be imported somewhere in client-side code.

There are further imports in some of the other TS examples, if the corresponding Flow example needs an additonal component (e.g. combo box needs combo box connector).

I will try to improve this in a separate PR by just adding these imports into one central place.

frontend/demo/component/grid/grid-column-width.ts
```diff
@@ -38,33 +41,27 @@ export class Example extends LitElement {
...
           <vaadin-grid-selection-column></vaadin-grid-selection-column>
-          <vaadin-grid-column path="displayName" width="10em"></vaadin-grid-column>
-          <vaadin-grid-column path="profession" auto-width></vaadin-grid-column>
+          <vaadin-grid-column path="firstName" width="7em" flex-grow="0"></vaadin-grid-column>
+          <vaadin-grid-column path="profession" auto-width flex-grow="0"></vaadin-grid-column>
           <vaadin-grid-column path="email"></vaadin-grid-column>
           <vaadin-grid-column
-            width="7em"
-            header="Manage"
-            .renderer="${this.manageRenderer}"
```
sissbruecker, line 47: Replaced the manage column here with a simple text column - it would be more trouble to set this up with Flow, and it's not absolutely needed for the example.

frontend/demo/component/grid/grid-content.ts
```diff
@@ -49,16 +51,11 @@ export class Example extends LitElement {
           .renderer="${this.statusRenderer}"
           auto-width
         ></vaadin-grid-column>
-        <vaadin-grid-column
-          header="Manage"
-          .renderer="${this.manageRenderer}"
```
sissbruecker, line 54: Removed this one, there are already other columns in this example demonstrating how to use custom content, no need to add another complex custom rendering example.

frontend/demo/component/grid/grid-dynamic-height.ts
```diff
@@ -46,22 +49,19 @@ export class Example extends LitElement {
...
         ></vaadin-combo-box>
         <vaadin-button
           theme="primary"
           @click="${() => {
-            const value = this.selectedItem;
-            const personIdx = this.items.findIndex((p) => +p.id === +value);
-            if (personIdx >= 0) {
-              const person = this.items.splice(personIdx, 1)[0];
-              this.items = [...this.items]; // re-assign the array to refresh the combo-box
+            const person = this.items.find((p) => p.id == parseInt(this.selectedValue));
```
sissbruecker, line 61: Simplified the logic here a bit by removing usage of the item index.

articles/ds/components/grid/index.asciidoc
```diff
@@ -69,12 +76,6 @@ Component renderers are easy to build but slow to render.
 For a given column, they generate a component for each item in the dataset.
 The rendered components are fully controllable on the server side.
 
-[.example]
---
-
-// TODO: Java example for component renderer goes here
```
sissbruecker, line 75: Removed this, instead used a component renderer in the GridContent.java example above.

articles/ds/components/grid/index.asciidoc
```diff
@@ -90,14 +91,6 @@ To use components with template renderers, you must use their HTML format.
 Templates are immutable, meaning the state of the components cannot be managed on the server side.
 The template can however have different representations depending on the state of the item.
 
-// Consider the following example:
-
-[.example]
---
-
-// TODO: Java example for template renderer goes here
```
sissbruecker, line 98: Removed this, instead used a template renderer in the GridContent.java example above.

frontend/demo/component/grid/grid-column-header-footer.ts
```diff
@@ -1,10 +1,11 @@
 import 'Frontend/demo/init'; // hidden-source-line
 import '@vaadin/flow-frontend/gridConnector.js'; // hidden-source-line (Grid's connector)
+import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js'; // hidden-source-line (Legacy template renderer)
 
 import { html, LitElement, render } from 'lit';
 import { customElement, state } from 'lit/decorators.js';
 import '@vaadin/vaadin-grid/vaadin-grid';
-import '@vaadin/vaadin-icons/vaadin-icons';
+import '@vaadin/vaadin-icons/vaadin-iconset';
```
vursen, line 8: Do not forget to make a correct rebase for the current branch after https://github.com/vaadin/docs/pull/590 gets merged. This PR adds correct imports for icons.

sissbruecker: Done.

articles/ds/components/grid/index.asciidoc
fhdhsni: Might be more clear if on [line 71](https://github.com/vaadin/docs/pull/596/files#R71) we could start like:
 ```
 As demonstrated in the example above, custom content can be...
 ```

src/main/java/com/vaadin/demo/component/grid/GridDynamicHeight.java
```diff
@@ -1,38 +1,108 @@
...
+        grid.setAllRowsVisible(true);
         // end::snippet[]
+        grid.addColumn(Person::getFullName).setHeader("Name");
+        grid.addColumn(Person::getEmail).setHeader("Email");
+        grid.addColumn(person -> person.getAddress().getPhone())
+                .setHeader("Phone");
+        grid.addColumn(new ComponentRenderer<>(Button::new, (button, person) -> {
+            button.setThemeName("icon", true);
+            button.setThemeName("error", true);
+            button.setThemeName("tertiary", true);
```
fhdhsni, line 64: also possible`button.addThemeNames("icon", "error", "tertiary");`

src/main/java/com/vaadin/demo/component/grid/GridColumnFreezing.java
```diff
@@ -14,25 +14,36 @@
...
 
         add(grid);
-        // end::snippet[]
     }
 
-    public static class Exporter extends DemoExporter<GridColumnFreezing> { // hidden-source-line
+    private static TemplateRenderer<Person> createPersonRenderer() {
+        return TemplateRenderer.<Person>of(
+                "<vaadin-horizontal-layout style=\"align-items: center;\" theme=\"spacing\">"
+                        + "  <vaadin-avatar img=\"[[item.pictureUrl]]\" name=\"[[item.fullName]]\"></vaadin-avatar>"
```
fhdhsni, line 39: nit: missing `alt` and `height` comparing to ts example

src/main/java/com/vaadin/demo/component/grid/GridColumnReorderingResizing.java
```diff
@@ -15,24 +15,39 @@
...
+        Grid.Column<Person> stateColumn = grid
+                .addColumn(person -> person.getAddress().getState())
+                .setHeader("State").setResizable(true);
+        // end::snippet[]
+
+        HeaderRow headerRow = grid.prependHeaderRow();
+        headerRow.join(firstNameColumn, lastNameColumn).setText("Name");
+        headerRow.join(streetColumn, cityColumn, zipColumn, stateColumn)
+                .setText("Address");
+
```
fhdhsni, line 43: would be if we end the snippet here (`end::snippet[]`)

src/main/java/com/vaadin/demo/component/grid/GridContent.java
```diff
@@ -14,25 +17,53 @@
...
     public GridContent() {
-        // tag::snippet[]
-        Grid<Person> grid = new Grid<>(Person.class);
+        // tag::snippet1[]
+        Grid<Person> grid = new Grid<>(Person.class, false);
+        grid.setSelectionMode(Grid.SelectionMode.MULTI);
+        grid.addColumn(createEmployeeTemplateRenderer()).setHeader("Employee")
+                .setAutoWidth(true).setFlexGrow(0);
+        grid.addColumn(Person::getProfession).setHeader("Profession")
+                .setAutoWidth(true);
```
fhdhsni, line 26: TS side is resizable: `.setResizable(true);`

src/main/java/com/vaadin/demo/component/grid/GridContent.java
```diff
@@ -14,25 +17,53 @@
...
         add(grid);
-        // end::snippet[]
     }
 
-    public static class Exporter extends DemoExporter<GridContent> { // hidden-source-line
+    // tag::snippet2[]
+    private static TemplateRenderer<Person> createEmployeeTemplateRenderer() {
+        return TemplateRenderer.<Person>of(
+                "<vaadin-horizontal-layout style=\"align-items: center;\" theme=\"spacing\">"
+                        + "<vaadin-avatar img=\"[[item.pictureUrl]]\" name=\"[[item.fullName]]\"></vaadin-avatar>"
```
fhdhsni, line 41: nit: missing `alt="User avatar"`

src/main/java/com/vaadin/demo/component/grid/GridDynamicHeight.java
```diff
@@ -1,38 +1,108 @@
...
+            button.addClickListener(e -> this.removeInvitation(person));
```
fhdhsni, line 65: on the TS side the removed person is prepended to the beginning of the list.

articles/ds/components/grid/index.asciidoc
sissbruecker: Good idea, done.

src/main/java/com/vaadin/demo/component/grid/GridDynamicHeight.java
sissbruecker: Nice, done.

src/main/java/com/vaadin/demo/component/grid/GridColumnFreezing.java
sissbruecker: Done.

src/main/java/com/vaadin/demo/component/grid/GridContent.java
sissbruecker: Done

src/main/java/com/vaadin/demo/component/grid/GridColumnReorderingResizing.java
sissbruecker: Done

src/main/java/com/vaadin/demo/component/grid/GridContent.java
sissbruecker: Removed the resizable config in the TS example instead, as I don't see how this is relevant for the example.

src/main/java/com/vaadin/demo/component/grid/GridContent.java
sissbruecker: Done

src/main/java/com/vaadin/demo/component/grid/GridDynamicHeight.java
sissbruecker: Looks like the example was supposed to remove invited people from the combo box options and then add them again when the invitation is removed. However only the second part is implemented, plus I think this example is already elaborate enough without that detail.

So decided to remove that logic from the TS example instead.

sissbruecker: Also added a check to the TS example to only invite persons, that are not yet invited.

fhdhsni (approved):

PR #612, opened 2021-07-01
docs: add second set of Grid Flow examples

sissbruecker:
## Description

This PR adds a second batch of Flow examples for the Grid. The examples cover the following sections:
- Sorting
- Filtering
- Item Details
- Context Menu

### DIFF
--- a/frontend/demo/component/grid/grid-column-filtering.ts
+++ b/frontend/demo/component/grid/grid-column-filtering.ts
@@ -13,7 +13,7 @@ import { getPeople } from 'Frontend/demo/domain/DataService';
 import Person from 'Frontend/generated/com/vaadin/demo/domain/Person';
 import { applyTheme } from 'Frontend/generated/theme';
 
-@customElement('grid-build-in-filtering')
+@customElement('grid-column-filtering')
 export class Example extends LitElement {
   protected createRenderRoot() {
     const root = super.createRenderRoot();
--- a/frontend/demo/component/grid/grid-context-menu.ts
+++ b/frontend/demo/component/grid/grid-context-menu.ts
@@ -1,6 +1,7 @@
 import 'Frontend/demo/init'; // hidden-source-line
 import '@vaadin/flow-frontend/gridConnector.js'; // hidden-source-line (Grid's connector)
 import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js'; // hidden-source-line (Legacy template renderer)
+import '@vaadin/flow-frontend/contextMenuConnector.js'; // hidden-source-line (ContextMenu's connector)
 
 import { html, LitElement, render } from 'lit';
 import { customElement, state } from 'lit/decorators.js';
@@ -8,8 +9,6 @@ import '@vaadin/vaadin-grid/vaadin-grid';
 import '@vaadin/vaadin-context-menu/vaadin-context-menu';
 import '@vaadin/vaadin-list-box/vaadin-list-box';
 import '@vaadin/vaadin-item/vaadin-item';
-import '@vaadin/vaadin-icon/vaadin-icon';
-import '@vaadin/vaadin-icons/vaadin-iconset';
 import type { GridElement, GridEventContext } from '@vaadin/vaadin-grid/vaadin-grid';
 import { getPeople } from 'Frontend/demo/domain/DataService';
 import Person from 'Frontend/generated/com/vaadin/demo/domain/Person';
@@ -41,18 +40,22 @@ export class Example extends LitElement {
   private contextMenuRenderer =
     () => (root: HTMLElement, elem: ContextMenuElement, context: ContextMenuRendererContext) => {
       const { sourceEvent } = context.detail! as { sourceEvent: Event };
-      const grid = elem.firstElementChild as GridElement;
+      const grid = elem.firstElementChild as GridElement<Person>;
 
-      const eventContext = grid.getEventContext(sourceEvent) as GridEventContext<Person>;
-      const person = eventContext.item as Person;
+      const eventContext = grid.getEventContext(sourceEvent);
+      const person = eventContext.item!;
+
+      const clickHandler = (_action: string) => () => {
+        // console.log(`${action}: ${person.firstName} ${person.lastName}`);
+      };
 
       render(
-        html`<vaadin-list-box style="color: var(--lumo-contrast-70pct);">
-          <h6 style="margin: 0 var(--lumo-space-m)">${person.firstName} ${person.lastName}</h6>
-          ${this.createItem('vaadin:pencil', 'Edit')} ${this.createItem('vaadin:trash', 'Delete')}
+        html`<vaadin-list-box>
+          <vaadin-item @click="${clickHandler('Edit')}">Edit</vaadin-item>
+          <vaadin-item @click="${clickHandler('Delete')}">Delete</vaadin-item>
           <hr />
-          ${this.createItem('vaadin:envelope-o', 'Email')}
-          ${this.createItem('vaadin:phone', 'Call')}
+          <vaadin-item @click="${clickHandler('Email')}">Email (${person.email})</vaadin-item>
+          <vaadin-item @click="${clickHandler('Call')}">Call (${person.address.phone})</vaadin-item>
         </vaadin-list-box>`,
         root
       );
@@ -69,16 +72,6 @@ export class Example extends LitElement {
         </vaadin-grid>
       </vaadin-context-menu>
     `;
-  }
-
-  createItem(icon: string, text: string) {
-    return html`<vaadin-item style="font-size: var(--lumo-font-size-s)"
-      ><vaadin-icon
-        icon="${icon}"
-        style="width: var(--lumo-icon-size-s); height: var(--lumo-icon-size-s); margin-right: var(--lumo-space-s)"
-      ></vaadin-icon
-      >${text}</vaadin-item
-    > `;
   }
 
   onContextMenu(e: MouseEvent) {
--- a/frontend/demo/component/grid/grid-external-filtering.ts
+++ b/frontend/demo/component/grid/grid-external-filtering.ts
@@ -1,6 +1,7 @@
 import 'Frontend/demo/init'; // hidden-source-line
 import '@vaadin/flow-frontend/gridConnector.js'; // hidden-source-line (Grid's connector)
-import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js'; // hidden-source-line (Legacy template renderer)
+import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js';
+import { TextFieldValueChangedEvent } from '@vaadin/vaadin-text-field'; // hidden-source-line (Legacy template renderer)
 
 import { html, LitElement, render } from 'lit';
 import { customElement, state } from 'lit/decorators.js';
@@ -17,7 +18,8 @@ import Person from 'Frontend/generated/com/vaadin/demo/domain/Person';
 import { applyTheme } from 'Frontend/generated/theme';
 
 type PersonEnhanced = Person & { displayName: string };
-@customElement('grid-filter')
+
+@customElement('grid-external-filtering')
 export class Example extends LitElement {
   protected createRenderRoot() {
     const root = super.createRenderRoot();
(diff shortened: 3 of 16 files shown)


frontend/demo/component/grid/grid-column-filtering.ts
```diff
@@ -13,7 +13,7 @@ import { getPeople } from 'Frontend/demo/domain/DataService';
 import Person from 'Frontend/generated/com/vaadin/demo/domain/Person';
 import { applyTheme } from 'Frontend/generated/theme';
 
-@customElement('grid-build-in-filtering')
+@customElement('grid-column-filtering')
```
sissbruecker, line 16: There's no built-in column filter for Flow, renamed filter examples to:
- `grid-column-filtering`
- `grid-external-filtering`

frontend/demo/component/grid/grid-context-menu.ts
```diff
@@ -41,18 +40,22 @@ export class Example extends LitElement {
...
-      const person = eventContext.item as Person;
+      const eventContext = grid.getEventContext(sourceEvent);
+      const person = eventContext.item!;
+
+      const clickHandler = (_action: string) => () => {
+        // console.log(`${action}: ${person.firstName} ${person.lastName}`);
+      };
 
       render(
-        html`<vaadin-list-box style="color: var(--lumo-contrast-70pct);">
```
sissbruecker, line 50: This example would have been difficult to recreate in Flow, and IMO it kind of misses the point. While it demonstrates how to use the person data in the menu, it also adds all sorts of visual fluff that is not relevant and is awkward to implement in Flow.

Instead I changed to example to:
- only use simple text items, no custom styles or elements apart from the divider
- demonstrated how to access the person in the menu, by adding the email and phone number to the respective menu items

frontend/demo/component/grid/grid-external-filtering.ts
```diff
@@ -46,19 +48,20 @@ export class Example extends LitElement {
         <vaadin-text-field
           placeholder="Search"
           style="width: 50%;"
-          @value-changed="${(e: CustomEvent) => {
```
sissbruecker, line 49: The example here allowed searching for multiple search terms by splitting up the text input. Again, that would be more code in Java, and is not really required for the example. 

Changed the examples to only search a single term.

frontend/demo/component/grid/grid-item-details-toggle.ts
```diff
@@ -31,7 +31,7 @@ export class Example extends LitElement {
   private items: Person[] = [];
 
   @state()
-  private detailsOpenedItem: Array<Person | null> = [];
+  private detailsOpenedItems: Person[] = [];
```
sissbruecker, line 34: The TS example here actually does not only toggle, but also makes sure that only one item detail is opened at the same time. The Flow API does not expose this kind of state (list of opened details), but only has APIs to check or change the opened state of a single item.

Changed the examples to simply toggle the opened state when clicking on a button, which means that multiple rows can have opened details now.

Given that, the example

frontend/demo/component/grid/grid-rich-content-sorting.ts
```diff
@@ -82,16 +82,14 @@ export class Example extends LitElement {
     const birthday = parseISO(person.birthday);
     render(
       html`
-        <vaadin-horizontal-layout style="align-items: center;" theme="spacing">
```
sissbruecker, line 85: `<vaadin-horizontal-layout>` seems not to be needed here.

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
```diff
@@ -0,0 +1,124 @@
...
+import com.vaadin.flow.component.textfield.TextField;
+import com.vaadin.flow.data.renderer.TemplateRenderer;
+import com.vaadin.flow.data.value.ValueChangeMode;
+import com.vaadin.flow.router.Route;
+
+import java.util.List;
+import java.util.function.Consumer;
+
+@Route("grid-column-filtering")
+public class GridColumnFiltering extends Div {
```
sissbruecker, line 22: This example requires a lot of manual wiring. I tried to add some abstractions to structure the code better, but I'm not sure if that makes the example easier to understand.

src/main/java/com/vaadin/demo/component/grid/GridRichContentSorting.java
```diff
@@ -0,0 +1,81 @@
...
+        return TemplateRenderer.<Person>of(
+                "<vaadin-vertical-layout style=\"line-height: var(--lumo-line-height-m);\">"
+                        + "  <span>[[item.birthday]]</span>"
+                        + "  <span style=\"font-size: var(--lumo-font-size-s); color: var(--lumo-secondary-text-color);\">Age: [[item.age]]</span>"
+                        + "</vaadin-vertical-layout>").withProperty("birthday",
+                GridRichContentSorting::getFormattedPersonBirthday)
+                .withProperty("age", GridRichContentSorting::getPersonAge);
+    }
+
+    private static final DateTimeFormatter birthdayFormatter = DateTimeFormatter
```
fhdhsni, line 60: nit: can be inlined

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public GridColumnFiltering() {
+        // tag::snippet1[]
+        Grid<Person> grid = new Grid<>(Person.class, false);
+        Grid.Column<Person> nameColumn = grid.addColumn(createPersonRenderer());
```
fhdhsni, line 27: `nameColumn.setFlexGrow(0).setWidth("230px");`

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
```diff
@@ -0,0 +1,124 @@
...
+                createFilterHeader("Profession", personFilter::setProfession));
+        // end::snippet1[]
+
+        add(grid);
+    }
+
+    // tag::snippet2[]
+    private static Component createFilterHeader(String labelText,
+            Consumer<String> filterChangeConsumer) {
+        Label label = new Label(labelText);
```
fhdhsni, line 53:
```
label.getStyle().set("padding-top", "var(--lumo-space-m)");
label.getStyle().set("font-size", "var(--lumo-font-size-xs)");
```

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
```diff
@@ -0,0 +1,124 @@
...
+        TextField textField = new TextField();
+        textField.setValueChangeMode(ValueChangeMode.EAGER);
+        textField.setClearButtonVisible(true);
+        textField.setThemeName("small");
+        textField.setWidthFull();
```
fhdhsni, line 58: `textField.getStyle().set("max-width", "100%");`

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
```diff
@@ -0,0 +1,124 @@
...
+        public boolean test(Person person) {
+            boolean matchesFullName = contains(person.getFullName(), fullName);
+            boolean matchesEmail = contains(person.getEmail(), email);
+            boolean matchesProfession = contains(person.getProfession(),
+                    profession);
+
+            return matchesFullName && matchesEmail && matchesProfession;
+        }
+
+        private boolean contains(String value, String searchText) {
```
fhdhsni, line 104: `searchText` -> `searchTerm`

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
```diff
@@ -0,0 +1,124 @@
...
+            return searchText == null || searchText.equals("") || value
```
fhdhsni, line 105: nit: `searchText.equals("")` =>`searchTerm.isEmpty()`

src/main/java/com/vaadin/demo/component/grid/GridContextMenuExample.java
```diff
@@ -0,0 +1,75 @@
...
+                        // System.out.printf("Phone: %s%n", person.getFullName());
+                    }));
+
+            setDynamicContentHandler(person -> {
+                // Do not show context menu when header is clicked
+                if (person == null)
+                    return false;
+                emailItem.setText(
+                        String.format("Email (%s)", person.getEmail()));
+                phoneItem.setText(String.format("Phone (%s)",
```
fhdhsni, line 64: `Call (%s)...`

src/main/java/com/vaadin/demo/component/grid/GridExternalFiltering.java
```diff
@@ -0,0 +1,78 @@
...
+
+import java.util.List;
+
+@Route("grid-external-filtering")
+public class GridExternalFiltering extends Div {
+
+    public GridExternalFiltering() {
+        // tag::snippet[]
+        Grid<Person> grid = new Grid<>(Person.class, false);
+        grid.addColumn(createPersonRenderer()).setHeader("Name");
```
fhdhsni, line 25:
```
   grid.addColumn(createPersonRenderer())
                .setHeader("Name")
                .setFlexGrow(0)
                .setWidth("230px");
```

src/main/java/com/vaadin/demo/component/grid/GridExternalFiltering.java
```diff
@@ -0,0 +1,78 @@
...
+        searchField.setWidth("50%");
+        searchField.setPlaceholder("Search");
+        searchField.setPrefixComponent(new Icon(VaadinIcon.SEARCH));
+        searchField.setValueChangeMode(ValueChangeMode.EAGER);
+        searchField.addValueChangeListener(e -> dataView.refreshAll());
+
+        dataView.addFilter(person -> {
+            String searchTerm = searchField.getValue().trim();
+
+            if ("".equals(searchTerm))
```
fhdhsni, line 42: `searchTerm.isEmpty()`

src/main/java/com/vaadin/demo/component/grid/GridItemDetails.java
```diff
@@ -0,0 +1,89 @@
...
+import com.vaadin.flow.router.Route;
+
+import java.util.List;
+
+@Route("grid-item-details")
+public class GridItemDetails extends Div {
+
+    public GridItemDetails() {
+        // tag::snippet1[]
+        Grid<Person> grid = new Grid<>(Person.class, false);
```
fhdhsni, line 20: the default selection mode between flow and pure WCs seems to be different. So if want to make them similar we have to add this to many examples I guess
```
grid.setSelectionMode(Grid.SelectionMode.NONE);
```

src/main/java/com/vaadin/demo/component/grid/GridItemDetails.java
```diff
@@ -0,0 +1,89 @@
...
+            add(streetField);
+            zipField = new TextField("ZIP code");
+            zipField.setReadOnly(true);
+            add(zipField);
+            cityField = new TextField("City");
+            cityField.setReadOnly(true);
+            add(cityField);
+            stateField = new TextField("State");
+            stateField.setReadOnly(true);
+            add(stateField);
```
fhdhsni, line 67: we can shorten this a bit via something like:
```
  Stream.of(emailField, phoneField, streetField, zipField, cityField, stateField).forEach(filed -> {
                filed.setReadOnly(true);
                add(filed);
            });
```

src/main/java/com/vaadin/demo/component/grid/GridItemDetailsToggle.java
```diff
@@ -0,0 +1,102 @@
...
+import com.vaadin.flow.router.Route;
+
+import java.util.List;
+
+@Route("grid-item-details-toggle")
+public class GridItemDetailsToggle extends Div {
+
+    public GridItemDetailsToggle() {
+        // tag::snippet1[]
+        Grid<Person> grid = new Grid<>(Person.class, false);
```
fhdhsni, line 21: `grid.setSelectionMode(Grid.SelectionMode.NONE);`

src/main/java/com/vaadin/demo/component/grid/GridItemDetailsToggle.java
```diff
@@ -0,0 +1,102 @@
...
+            add(streetField);
+            zipField = new TextField("ZIP code");
+            zipField.setReadOnly(true);
+            add(zipField);
+            cityField = new TextField("City");
+            cityField.setReadOnly(true);
+            add(cityField);
+            stateField = new TextField("State");
+            stateField.setReadOnly(true);
+            add(stateField);
```
fhdhsni, line 81: If you decided to apply the snippet above, don't forget to change this one too

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
fhdhsni: The `Grid` itself should provide a way to make this possible (like the TS side).
Your abstraction isn't hard to understand IMO but I wish it was more reusable (like people could copy and paste it to use it for different grids). 
Maybe we can use a hashmap to store fields dynamically. IDK, maybe it's good enough for what it's trying to show. After all code examples here are not trying to be production ready.

src/main/java/com/vaadin/demo/component/grid/GridItemDetails.java
sissbruecker: Yeah, the defaults are different. Since it's the default in Flow, I'm inclined to not change it in the examples, since that is what the user gets out of the box. I would only consider changing the default if it in some way creates confusion with what is demonstrated.

src/main/java/com/vaadin/demo/component/grid/GridItemDetails.java
sissbruecker: OK, also moved to initialization to the fields.

src/main/java/com/vaadin/demo/component/grid/GridRichContentSorting.java
sissbruecker: What do you mean by inclined?

src/main/java/com/vaadin/demo/component/grid/GridColumnFiltering.java
sissbruecker: Looking at the example again, I think the pattern is clear and developers should be able to apply it to their own use-cases with little effort.

A generic solution would be nice, but is realistically out of scope for our work. It would also need to handle different data types etc. so there is quite a bit more work. Might be a good idea to explore this as part of CF?

fhdhsni (approved):

PR #614, opened 2021-07-02
Add a link explaning the guard directive used in the code samples

gilberto-torrezan:


articles/ds/components/dialog/index.asciidoc
```diff
@@ -43,6 +43,9 @@ include::{root}/frontend/demo/component/dialog/dialog-basic.ts[render, tags=*, i
 ----
 --
 
+[IMPORTANT]
+The typescript examples in this page use an inline `.renderer` function, which is encapsulated by the `guard` directive in Lit. This is done for performance reasons. For more details on the `guard` directive, please refer to the https://lit.dev/docs/templates/directives/#guard[official documentation].
```
tomivirkki, line 47: The content of the asciidoc should be relevant to the Design System and ideally language independent. Using Lit guards is a technical detail in some examples so this notice would work better as a code comment next to where the `guard` is used.

In future, we'll replace the `guard` directives in the examples with something similar to https://github.com/web-padawan/lit-vaadin-helpers

magi42, line 47:
```suggestion
The TypeScript examples in this article use an inline `.renderer` function, which is encapsulated by the `guard` directive in Lit.
The examples do so for performance reasons.
For more details on the `guard` directive, refer to the https://lit.dev/docs/templates/directives/#guard[official Lit documentation].
```

magi42 (approved):

articles/ds/components/dialog/index.asciidoc
magi42: You can have language-specific notes also in example blocks for each language. Shouldn't use admonitions there though.

I made some language fixes, but maybe this issue should be solved.

gilberto-torrezan: How do I add notes to the example blocks? I seems to be just linking to an external file.

magi42: You can have description text above the `[source, Java]` line (leave one empty line between the text). This works inside `[.example]` boxes. Each language (group) can have such description. See for example the code example in https://vaadin.com/docs/latest/flow/application/routing-navigation

jouni:
Oh, damn. I didn't notice this targeted the v20 branch, and I rebased it on latest 🤦

jouni:
I’ll make a new branch, sorry for messing up this one.

jouni:
Pushed to `latest` in https://github.com/vaadin/docs/commit/32ef65648ce0666fc8d95271c56b3bd76a06410e

jouni:
Not sure if it’s relevant to cherry-pick to v20, as that branch will only be around for a month more or so.

PR #622, opened 2021-07-05
docs: add first batch of Flow Notification examples

vursen:
## Description

This PR adds Flow examples for the Notification component.

- Shows simple text notification instead of rich content notification in all samples except for `Error Theme Variant`, `Interactive Elements` and `Icons and Other Rich Formatting`.
- Disables autohide on the `Error Theme Variant` sample.
- Adds some code comments for the Java examples to explain an actual duration.
- Adds a disclaimer about [the `flow-component-renderer` styling issue](https://github.com/vaadin/flow-components/issues/1876)

## Sections

- [x] Basic 
- [x] Theme Variants
- [x] Position


frontend/demo/component/notification/notification-basic.ts
```diff
@@ -23,28 +26,30 @@ export class Example extends LitElement {
...
         Try it
       </vaadin-button>
 
       <!-- tag::snippet[] -->
       <vaadin-notification
         position="middle"
-        .opened="${this.notificationOpen}"
-        @opened-changed="${(e: any) => (this.notificationOpen = e.detail.value)}"
+        .opened="${this.notificationOpened}"
+        @opened-changed="${(e: any) => (this.notificationOpened = e.detail.value)}"
```
sissbruecker: We should be able to type the event here.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
```diff
@@ -0,0 +1,49 @@
...
+public class NotificationBasic extends Div {
+
+  public NotificationBasic() {
+    // tag::snippet[]
+    Notification notification = new Notification();
+    notification.setPosition(Notification.Position.MIDDLE);
+
+    Div notificationText = new Div(new Text("Financial report generated"));
+
+    Button notificationButton = new Button(new Icon("lumo", "cross"));
```
sissbruecker, line 23: Would suggest `closeButton`

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
```diff
@@ -0,0 +1,49 @@
...
+    Button button = new Button("Try it");
+    button.addClickListener(event -> {
+      notification.setOpened(true);
+    });
+
+    notification.addOpenedChangeListener(event -> {
+      button.setEnabled(!event.getSource().isOpened());
+    });
+
+    add(button, notification);
```
sissbruecker, line 43: We usually don't show this line, as it's something that the user quickly learns, and then it's just a distraction to have it in each example.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
```diff
@@ -0,0 +1,49 @@
...
+
+    Button notificationButton = new Button(new Icon("lumo", "cross"));
+    notificationButton.addThemeVariants(ButtonVariant.LUMO_TERTIARY_INLINE, ButtonVariant.LUMO_ICON);
+    notificationButton.getElement().setAttribute("aria-label", "Close");
+    notificationButton.addClickListener(event -> {
+      notification.setOpened(false);
+    });
+
+    HorizontalLayout notificationLayout = new HorizontalLayout(notificationText, notificationButton);
+    notificationLayout.setSpacing(true);
```
sissbruecker, line 31: Spacing is enabled by default for the Flow HorizontalLayout.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
```diff
@@ -0,0 +1,49 @@
...
+
+@Route("notification-basic")
+public class NotificationBasic extends Div {
+
+  public NotificationBasic() {
+    // tag::snippet[]
+    Notification notification = new Notification();
+    notification.setPosition(Notification.Position.MIDDLE);
+
+    Div notificationText = new Div(new Text("Financial report generated"));
```
sissbruecker, line 21: Could use a paragraph instead.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
```diff
@@ -0,0 +1,49 @@
...
+    notification.add(notificationLayout);
+
+    Button button = new Button("Try it");
+    button.addClickListener(event -> {
+      notification.setOpened(true);
+    });
+
+    notification.addOpenedChangeListener(event -> {
+      button.setEnabled(!event.getSource().isOpened());
+    });
```
sissbruecker, line 41: Not 100% necessary to show in the snippet IMO, this is not something that the user needs to learn from this example.

frontend/demo/component/notification/notification-basic.ts
sissbruecker: Nevermind, you already did this 👍

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: Which variant do you mean?

- `notificationText` and `closeButton`
- `notificationText` and `notificationCloseButton`

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: I looked over and seems that some of the examples hide this line, but not all.

I wonder if it would be reasonable to keep this line only for the first example in case the user comes to the notification article from the search results. What do you think?

Anyway, it is also fine to hide this line.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: Good point! 

But it seems quite fishy that the Flow default behavior is different from the Web Components one where you need to set spacing manually. 

I wonder if it's been done so intentionally.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: You are right. Not only the listener but rather the whole "Try it" button should be hidden here to align it with the TypeScript example.

vursen: Fixed.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: Good suggestion, thanks! Fixed.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: Got rid of this line from the snippet as well as the "Try it" button.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: There is a similar situation for the `duration` property whose default value is inconsistent between Flow- and Web- components.

Web Component: [5000ms](https://github.com/vaadin/web-components/blob/master/packages/vaadin-notification/src/vaadin-notification.js#L265-L268)
Flow Component: [0ms](https://github.com/vaadin/flow-components/blob/master/vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java#L128)

vursen: Updated the PR so that the notification's duration is now set manually for the Flow examples.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
sissbruecker: `closeButton` is enough IMO.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: Oh, probably using a paragraph here is not the best alternative since the paragraph has some default margins.

image

vursen: 1. 
```java
Div notificationText = new Div(new Text("Financial report generated"));
```

2. 
```java
Div notificationText = new Div();
notificationText.add("Financial report generated");
```

3.
```java
Paragraph notificationText = new Paragraph("Financial report generated");
notificationText.getStyle().set("margin", "0");
```

vursen: Even resetting `margin-bottom`, `margin-top`, something weird happens with cascade in the case of using paragraphes.

image

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
vursen: Fixed.

src/main/java/com/vaadin/demo/component/notification/NotificationContrast.java
```diff
@@ -0,0 +1,51 @@
...
+import com.vaadin.demo.DemoExporter; // hidden-source-line
+
+@Route("notification-contrast")
+public class NotificationContrast extends Div {
+
+  public NotificationContrast() {
+    // tag::snippet[]
+    Notification notification = new Notification();
+    notification.setPosition(Notification.Position.MIDDLE);
+    notification.setDuration(5000);
```
vursen, line 21: It might be not necessary to keep these lines inside the snippet since the example aims at showing how to use different themes.

frontend/demo/component/notification/notification-position.ts
```diff
@@ -34,24 +42,26 @@ export class Example extends LitElement {
     `;
   }
 
-  show(e: MouseEvent) {
+  show(event: MouseEvent) {
+    const button = event.target as ButtonElement;
     // Use the button label as the location
-    const position = (e.composedPath()[2] as HTMLElement).textContent as NotificationPosition;
+    const position = button.textContent as NotificationPosition;
```
vursen, line 48: Not sure if this change makes anything better.

frontend/demo/component/notification/notification-position.ts
```diff
@@ -34,24 +42,26 @@ export class Example extends LitElement {
...
 
     const notification = new NotificationElement();
     notification.position = position;
```
vursen, line 51: Would it be better to end the snipper right here since the example is about the `position` property?

src/main/java/com/vaadin/demo/component/notification/NotificationPosition.java
```diff
@@ -0,0 +1,71 @@
...
+    closeButton.addClickListener(event -> {
+      notification.close();
+    });
+
+    HorizontalLayout layout = new HorizontalLayout(text, closeButton);
+    layout.setAlignItems(Alignment.CENTER);
+    layout.setFlexGrow(1, text);
+    layout.setSizeFull();
+    notification.add(layout);
+    notification.open();
```
vursen, line 65: When opening a notification that hasn't been manually added to DOM via `add(...)`, the Flow [will automatically remove it](https://github.com/vaadin/flow-components/blob/master/vaadin-notification-flow-parent/vaadin-notification-flow/src/main/java/com/vaadin/flow/component/notification/Notification.java#L309-L310) from DOM as soon as it gets closed. So that you don't need to remove the notification element manually as well as the TypeScript example does.

src/main/java/com/vaadin/demo/component/notification/NotificationPosition.java
```diff
@@ -0,0 +1,71 @@
...
+    Button button = new Button(position.getClientName());
+    button.addClickListener(event -> show(position));
+    return button;
+  }
+  // end::createButton[]
+
+  // tag::show[]
+  private void show(Notification.Position position) {
+    Notification notification = new Notification();
+    notification.setPosition(position);
```
vursen, line 48: Wouldn't it be better to end the `show` snippet over here since the example is pretty much about notification positions?

src/main/java/com/vaadin/demo/component/notification/NotificationContrast.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public NotificationContrast() {
+    // tag::snippet[]
+    Notification notification = new Notification();
+    notification.addThemeVariants(NotificationVariant.LUMO_CONTRAST);
+    // end::snippet[]
+    notification.setPosition(Notification.Position.MIDDLE);
+    notification.setDuration(5000);
+
+    Div notificationText = new Div(new Text("5 tasks deleted"));
```
vursen, line 26: **proposal:** `Div statusText = new Div(...)` ?

src/main/java/com/vaadin/demo/component/notification/NotificationContrast.java
```diff
@@ -0,0 +1,53 @@
...
+
+    Button closeButton = new Button(new Icon("lumo", "cross"));
+    closeButton.addThemeVariants(ButtonVariant.LUMO_TERTIARY_INLINE, ButtonVariant.LUMO_ICON);
+    closeButton.getElement().setAttribute("aria-label", "Close");
+    closeButton.addClickListener(event -> {
+      notification.close();
+    });
+
+    HorizontalLayout notificationLayout = new HorizontalLayout(notificationText, closeButton);
```
vursen, line 35: **proposal:** `HorizontalLayout layout = new HorizontalLayout(...)` ?

src/main/java/com/vaadin/demo/component/notification/NotificationContrast.java
```diff
@@ -0,0 +1,53 @@
...
+    notificationLayout.setAlignItems(Alignment.CENTER);
+    notification.add(notificationLayout);
+
+    Button button = new Button("Try it");
```
vursen, line 39: **proposal:** `Button tryButton = new Button(...)` ?

frontend/themes/docs/notification.css
```diff
@@ -0,0 +1,10 @@
+/* tag::snippet[] */
+/*
+  A workaround to prevent `flow-component-renderer` from affecting the cascade.
+  The related issue: https://github.com/vaadin/flow-components/issues/1876.
+*/
+vaadin-notification-card flow-component-renderer,
```
sissbruecker, line 6: This might be more appropriate as a component specific theme file (`components` folder).

If you decide to change this, it's good to add a comment to the top of the file that shows the file path where the user would have to put the file to take an effect. For example as done in `frontend/themes/docs/components/vaadin-grid-cell-focus.css`

frontend/demo/component/notification/notification-position.ts
sissbruecker: This is a lot better IMO.

src/main/java/com/vaadin/demo/component/notification/NotificationContrast.java
sissbruecker: Since there is only one button, it's good enough IMO.

frontend/demo/component/notification/notification-basic-preview.ts
```diff
@@ -1,8 +1,6 @@
 import 'Frontend/demo/init'; // hidden-source-line
+import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js'; // hidden-source-line (Legacy template renderer)
```
sissbruecker, line 2: Not required in preview components, as there is no Java equivalent.

frontend/demo/component/notification/notification-error.ts
```diff
@@ -22,36 +28,44 @@ export class Example extends LitElement {
...
+        @click="${() => (this.notificationOpened = true)}"
+        .disabled="${this.notificationOpened}"
       >
         Try it
       </vaadin-button>
 
       <!-- tag::snippet[] -->
       <vaadin-notification
         theme="error"
+        duration="0"
```
sissbruecker, line 40: We can also add a comment above the `vaadin-notification´ tag to explain that / why duration was set to zero.

frontend/demo/component/notification/notification-retry-preview.ts
```diff
@@ -18,8 +19,9 @@ export class Example extends LitElement {
     return html`
       <vaadin-notification-card theme="error" slot="middle">
         <div>Failed to generate report</div>
-        <div style="width: 2em"></div>
-        <vaadin-button theme="tertiary-inline">Retry</vaadin-button>
+        <vaadin-button theme="tertiary-inline" style="margin-left: var(--lumo-space-xl);">
```
sissbruecker, line 22: Not sure why the margin here was added, and the corresponding example was not changed / does not contain the margin.

frontend/themes/docs/styles.css
```diff
@@ -3,7 +3,9 @@
 @import './basic-layouts.css';
 @import './cookie-consent-theming.css';
 @import './login-rich-content.css';
-@import './notification-examples.css';
+@import './notification.css';
+@import './notification-rich-example.css';
+@import './notification-position-example.css';
```
sissbruecker, line 8: I would consider adding this to the example sources, someone might be interested in why the buttons are so neatly arranged.

src/main/java/com/vaadin/demo/component/notification/NotificationError.java
```diff
@@ -0,0 +1,54 @@
...
+    Button button = new Button("Try it");
+    button.addClickListener(event -> {
+      notification.open();
+    });
+
+    notification.addOpenedChangeListener(event -> {
+      button.setEnabled(!event.getSource().isOpened());
+    });
+
+    add(button, notification);
```
sissbruecker, line 49: I think the snippet here is problematic, in that it doesn't show that you need to:
- explicitely call `notification.open()` when using the constructor variant
- that you need to add the notification to the component

...both of which are not required when using `Notification.show()`.

That could lead to confusion then someone copies this and nothing shows up.

frontend/demo/component/notification/notification-retry-preview.ts
vursen: Yeah, seems that it has appeared here after a rebasing. Fixed.

src/main/java/com/vaadin/demo/component/notification/NotificationError.java
vursen: Good point!

1. Another option could be to create and open the notification only on-demand, however, it seems to complicate the `setEnabled` logic a bit.

```java
show(detachCallback...) {
  Notification notification = new Notificatication(...);
  notification.open();
  notification.addDetachListener(event -> detachCallback())
}
```

2. Or extend the snippet to grab the entire constructor.

src/main/java/com/vaadin/demo/component/notification/NotificationBasic.java
```diff
@@ -0,0 +1,35 @@
...
+      // tag::snippet[]
+      // When creating a notification using the `show` static method,
+      // the duration is 5-sec by default.
+      Notification notification = Notification.show("Financial report generated");
+      // end::snippet[]
+      notification.setPosition(Notification.Position.MIDDLE);
+
+      notification.addDetachListener(detachEvent -> {
+        button.setEnabled(true);
+      });
```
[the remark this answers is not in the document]
sissbruecker: Actually, when we want to show off some API, it's better if it is on a single line. For plumbing around the example (like in the snippet above, toggling the button state) it's OK to do it inline.

src/main/java/com/vaadin/demo/component/notification/NotificationError.java
vursen: Although, I think that having an example that adds the notification to DOM is also necessary. But we could demonstrate such a case in the "Rich Content" section, for example.

frontend/demo/component/notification/notification-error.ts
```diff
@@ -22,36 +28,45 @@ export class Example extends LitElement {
...
-        @click="${() => (this.notificationOpen = true)}"
-        .disabled="${this.notificationOpen}"
+        @click="${() => (this.notificationOpened = true)}"
+        .disabled="${this.notificationOpened}"
       >
         Try it
       </vaadin-button>
 
       <!-- tag::snippet[] -->
+      <!-- The duration is set to 0-sec to prevent the notification from auto-close. -->
```
vursen, line 38:
```suggestion
      <!-- Setting the duration to 0-sec disables auto-close for the notification. -->
```

src/main/java/com/vaadin/demo/component/notification/NotificationError.java
vursen: Now I'm thinking that the first option is probably best in this case.

Consider the example:

```java
public NotificationRetry() {
  Button button = new Button("Try it");
  button.addClickListener(clickEvent -> {
    button.setEnabled(false);

    Notification notification = show();

    notification.addDetachListener(detachEvent -> button.setEnabled(true));
  });

  add(button);
}

// tag::show[]
private Notification show() {
  // When creating a notification using the constructor,
  // the duration is 0-sec by default, which means the notification does not close automatically.
  Notification notification = new Notification();
  notification.setPosition(Notification.Position.MIDDLE);
  notification.addThemeVariants(NotificationVariant.LUMO_ERROR);

  Div statusText = new Div(new Text("Failed to generate report"));

  Button retryButton = new Button("Retry");
  retryButton.addThemeVariants(ButtonVariant.LUMO_TERTIARY_INLINE);
  retryButton.getElement().getStyle().set("margin-left", "var(--lumo-space-xl)");
  retryButton.addClickListener(event -> {
    notification.close();
  });

  Button closeButton = new Button(new Icon("lumo", "cross"));
  closeButton.addThemeVariants(ButtonVariant.LUMO_TERTIARY_INLINE);
  closeButton.getElement().setAttribute("aria-label", "Close");
  closeButton.addClickListener(event -> {
    notification.close();
  });

  HorizontalLayout notificationLayout = new HorizontalLayout(statusText, retryButton, closeButton);
  notificationLayout.setAlignItems(Alignment.CENTER);
  notification.add(notificationLayout);
  notification.open();

  return notification;
}
// end::show[]
```

frontend/themes/docs/notification.css
vursen: The thread is no more actual since we decided to add a disclaimer instead of a code comment.

src/main/java/com/vaadin/demo/component/notification/NotificationError.java
vursen: Updated the PR so that the 1st variant now is being used.

articles/ds/components/notification/index.asciidoc
```diff
@@ -216,6 +263,23 @@ include::{root}/frontend/demo/component/notification/notification-position.ts[re
...
 - Avoid using positions that may obstruct important parts of the UI, such as navigation.
 
+.The Flow styling issue
+[WARNING]
+====
+At the moment, Flow renders notifications with rich content via
+intermediate `<flow-component-renderer />` which may break the CSS cascade
+and prevent the content from taking full width.
+
+To work around this issue, add the following CSS snippet globally:
```
sissbruecker, line 273:
```suggestion
.Flow styling issue
[WARNING]
====
The Flow notification component renders rich content via an intermediate `<flow-component-renderer>` element.
This can prevent notification content from using the full width, when using top stretch or bottom stretch positioning.
To work around this issue, add the following CSS rule to a global style sheet:
```

Also it would be good if the hint would not use a red border. This is more of an edge-case, but it currently looks like something super important that takes the readers focus away from the content. I would suggest to make it yellow if possible, or just use a regular note.

vursen: Great corrections, thanks! Fixed.

articles/ds/components/notification/index.asciidoc
```diff
@@ -216,6 +263,21 @@ include::{root}/frontend/demo/component/notification/notification-position.ts[re
 - For a consistent user experience, use one or two positions throughout the application.
 - Avoid using positions that may obstruct important parts of the UI, such as navigation.
 
+.The Flow styling issue
```
sissbruecker, line 266:
```suggestion
.Flow styling issue
```

sissbruecker (approved):

PR #623, opened 2021-07-05
Java: Date Picker examples

anezthes:
"Format" TS example is missing locale, so the suggested format is incorrect for everyone that doesn't have a locale that follows DD/MM/YYYY.

Flow/Java version works as expected.


src/main/java/com/vaadin/demo/component/datepicker/DatePickerAutoOpen.java
```diff
@@ -1,21 +1,20 @@
...
 
     public DatePickerAutoOpen() {
+        DatePicker datePicker = new DatePicker("Start date");
         // tag::snippet[]
-        DatePicker datePicker = new DatePicker("Birthday");
-
-        add(datePicker);
+        datePicker.setAutoOpen(false);
         // end::snippet[]
+        add(datePicker);
```
sissbruecker, line 16: Nit: Add a newline after the snippet end. Since the whole line is removed, the blocks will just collapse into one, hurting readability a bit. Same could be applied to the other examples in this PR.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerIndividualInputFields.java
```diff
@@ -1,21 +1,43 @@
...
+
+import java.time.LocalDate;
+import java.time.Month;
+import java.util.stream.IntStream;
 
 @Route("date-picker-individual-input-fields")
 public class DatePickerIndividualInputFields extends Div {
 
     public DatePickerIndividualInputFields() {
+        LocalDate now = LocalDate.now(ZoneId.of("Europe/Helsinki"));
```
sissbruecker, line 22: `LocalDate.now(ZoneOffset.UTC)` would be more appropriate here to keep it neutral. There's also `LocalDate.now(ZoneId.systemDefault())` which would be the best choice if we assume that our customers mostly host their applications in the same zone as their users are in.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerIndividualInputFields.java
```diff
@@ -1,21 +1,43 @@
...
+import java.time.ZoneId;
+import java.time.format.TextStyle;
+import java.util.Locale;
+
+import java.time.LocalDate;
+import java.time.Month;
+import java.util.stream.IntStream;
 
 @Route("date-picker-individual-input-fields")
 public class DatePickerIndividualInputFields extends Div {
```
sissbruecker, line 19: I don't think this example has any practical value:
- It doesn't actually contain a working solution
- Date handling is **hard**, we should not recommend building something like this on your own, providing this naive setup as a start (e.g. not every month has 31 days).

I would suggest to remove the examples, and possibly the section alltogether.

Alternatively recommend the native date picker, for which mobile browsers have optimized UIs.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerInitialPosition.java
```diff
@@ -1,21 +1,25 @@
...
+
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.time.temporal.TemporalAdjusters;
 
 @Route("date-picker-initial-position")
 public class DatePickerInitialPosition extends Div {
 
     public DatePickerInitialPosition() {
+        LocalDate now = LocalDate.now(ZoneId.of("Europe/Helsinki"));
```
sissbruecker, line 16: `LocalDate.now(ZoneOffset.UTC)` would be more appropriate here to keep it neutral. There's also `LocalDate.now(ZoneId.systemDefault())` which would be the best choice if we assume that our customers mostly host their applications in the same zone as their users are in.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerInitialPosition.java
```diff
@@ -1,21 +1,25 @@
...
+        DatePicker datePicker = new DatePicker("Q4 deadline");
         // tag::snippet[]
-        DatePicker datePicker = new DatePicker("Birthday");
-
-        add(datePicker);
+        datePicker.setInitialPosition(now.with(TemporalAdjusters.lastDayOfYear()));
```
sissbruecker, line 19: I would suggest to include constructing the `LocalDate` into the snippet. Otherwise it's not clear what kind of parameter the method accepts.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerMinMax.java
```diff
@@ -1,21 +1,26 @@
...
-import com.vaadin.demo.DemoExporter; // hidden-source-line
+
+import java.time.LocalDate;
+import java.time.ZoneId;
 
 @Route("date-picker-min-max")
 public class DatePickerMinMax extends Div {
 
     public DatePickerMinMax() {
+        LocalDate now = LocalDate.now(ZoneId.of("Europe/Helsinki"));
```
sissbruecker, line 15: `LocalDate.now(ZoneOffset.UTC)` would be more appropriate here to keep it neutral. There's also `LocalDate.now(ZoneId.systemDefault())` which would be the best choice if we assume that our customers mostly host their applications in the same zone as their users are in.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerMinMax.java
```diff
@@ -1,21 +1,26 @@
...
+        DatePicker datePicker = new DatePicker("Appointment date");
         // tag::snippet[]
-        DatePicker datePicker = new DatePicker("Birthday");
-
-        add(datePicker);
+        datePicker.setMin(now);
```
sissbruecker, line 18: I would suggest to include constructing the `LocalDate` into the snippet. Otherwise it's not clear what kind of parameter the method accepts.

src/main/java/com/vaadin/demo/component/datetimepicker/DateTimePickerAutoOpen.java
```diff
@@ -3,21 +3,18 @@
...
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.router.Route;
-
 import com.vaadin.demo.DemoExporter; // hidden-source-line
 
 @Route("date-time-picker-auto-open")
 public class DateTimePickerAutoOpen extends Div {
 
     public DateTimePickerAutoOpen() {
+        DateTimePicker dateTimePicker = new DateTimePicker("Start date");
```
sissbruecker, line 12: I think it would be useful to add something regarding "time" to the label.

src/main/java/com/vaadin/demo/component/datetimepicker/DateTimePickerInitialPosition.java
```diff
@@ -1,30 +1,23 @@
...
-import java.time.LocalDateTime;
-import java.util.Arrays;
-
-import com.vaadin.demo.DemoExporter; // hidden-source-line
 
 @Route("date-time-picker-initial-position")
 public class DateTimePickerInitialPosition extends Div {
 
     public DateTimePickerInitialPosition() {
+        DateTimePicker dateTimePicker = new DateTimePicker("Q4 deadline");
```
sissbruecker, line 14: "Q4 deadline" has a time? That's very specific 😄

package.json
```diff
@@ -1,7 +1,17 @@
 {
   "dependencies": {
+    "@polymer/iron-a11y-announcer": "3.0.2",
```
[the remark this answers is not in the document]
sissbruecker: done.

src/main/java/com/vaadin/demo/component/datetimepicker/DateTimePickerInitialPosition.java
sissbruecker: Changed to align with meeting examples.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerIndividualInputFields.java
sissbruecker: Changed examples to a minimal working solution, added note that this is not a production quality implementation.

frontend/demo/component/datepicker/date-picker-individual-input-fields.ts
```diff
@@ -34,17 +35,70 @@ export class Example extends LitElement {
...
+  @state()
+  selectedYear?: number;
+  @state()
+  selectedMonth?: string;
+  @state()
+  selectedDay?: number;
+  @state()
+  selectableDays: number[] = [];
+
+  private handleYearChange(e: ComboBoxSelectedItemChangedEvent<number>) {
```
sissbruecker, line 48: Previously this example would only show three combo boxes with fixed selections, e.g. every month had 31 days available for selection.

Extended the examples a bit to make sort of sure that only valid date combinations can be picked. It's still a very naive implementation. For this added a note below the example to indicate that this is only for demonstration purposes and not production ready.

frontend/demo/component/datetimepicker/date-time-picker-range.ts
```diff
@@ -1,10 +1,10 @@
 import 'Frontend/demo/init'; // hidden-source-line
```
sissbruecker, line 1: Just noting here that this PR modifies several date-**time-**picker examples as well. Easy to mix them up if you are not aware.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerDateFormatIndicator.java
```diff
@@ -1,21 +1,25 @@
...
 @Route("date-picker-date-format-indicator")
 public class DatePickerDateFormatIndicator extends Div {
 
     public DatePickerDateFormatIndicator() {
+        DatePicker datePicker = new DatePicker("Start date");
+        datePicker.setLocale(new Locale("fi", "FI"));
         // tag::snippet[]
-        DatePicker datePicker = new DatePicker("Birthday");
+        datePicker.setPlaceholder("DD.MM.YYYY");
+        datePicker.setHelperText("Format: DD.MM.YYYY");
```
sissbruecker, line 18: TS + Java are intentionally different here it seems (see OP description). The Java example will initialize the input pattern based on the language codes from the locale, and the TS example initializes the pattern...based on something else? For me it's still using an english pattern even though my system settings are set to use german patterns.

Probably OK to ignore this for now since it's only a best practice example, and there is no dedicated feature for setting the input pattern.

src/main/java/com/vaadin/demo/component/datetimepicker/DateTimePickerInitialPosition.java
```diff
@@ -1,30 +1,34 @@
...
-        add(dateTimePicker);
+        DateTimeFormatter formatter = DateTimeFormatter.ISO_DATE;
+        LocalDate startOfNextMonth = LocalDate.now(ZoneId.systemDefault())
+                .with(TemporalAdjusters.firstDayOfNextMonth());
+        String startOfNextMonthIsoString = formatter.format(startOfNextMonth);
+
+        dateTimePicker.getElement().executeJs(
+                "this.initialPosition = $0",
+                startOfNextMonthIsoString
+        );
```
sissbruecker, line 28: There's no API for this in DateTimePicker. Note that initial position != initial value, position is the scroll position when opening the overlay.

frontend/demo/component/datetimepicker/date-time-picker-range.ts
```diff
@@ -28,18 +28,19 @@ export class Example extends LitElement {
...
-        @value-changed="${(e: CustomEvent) => (this.startDateTime = e.detail.value)}"
-      ></vaadin-date-time-picker>
-
-      <vaadin-date-time-picker
-        label="End date and time"
-        .min="${this.startDateTime}"
-        .value="${this.endDateTime}"
-        @value-changed="${(e: CustomEvent) => (this.endDateTime = e.detail.value)}"
-      ></vaadin-date-time-picker>
+      <vaadin-horizontal-layout theme="spacing">
```
fhdhsni, line 31: I get horizontal scroll bar. we can use a `div` like its Java counterpart.

src/main/java/com/vaadin/demo/component/datepicker/DatePickerIndividualInputFields.java
```diff
@@ -1,20 +1,98 @@
...
-        DatePicker datePicker = new DatePicker("Birthday");
+        LocalDate now = LocalDate.now(ZoneId.systemDefault());
+
+        List<Integer> selectableYears = IntStream.range(
+                now.getYear() - 99,
+                now.getYear() + 1)
+                .boxed().collect(Collectors.toList());
+
+        yearPicker = new ComboBox<>("Year");
+        yearPicker.setItems(selectableYears);
```
fhdhsni, line 37:
```suggestion
        yearPicker = new ComboBox<>("Year", selectableYears);
```

src/main/java/com/vaadin/demo/component/datepicker/DatePickerIndividualInputFields.java
```diff
@@ -1,20 +1,98 @@
...
+        yearPicker.setWidth(6, Unit.EM);
+        yearPicker.addValueChangeListener(e -> {
+            updateMonthPicker();
+            updateDayPicker();
+        });
+
+        monthPicker = new ComboBox<>("Month");
+        monthPicker.setItems(Month.values());
```
fhdhsni, line 45:
```suggestion
        monthPicker = new ComboBox<>("Month", Month.values());
```

src/main/java/com/vaadin/demo/component/datepicker/DatePickerIndividualInputFields.java
sissbruecker: Done

src/main/java/com/vaadin/demo/component/datepicker/DatePickerIndividualInputFields.java
sissbruecker: Done

frontend/demo/component/datetimepicker/date-time-picker-range.ts
sissbruecker: Done

fhdhsni (approved):

PR #630, opened 2021-07-07
docs: add third batch of Flow Grid examples

sissbruecker:
## Description

Adds Flow examples for the Grid component for the following sections:
- Drag and Drop
- Inline Editing
- Styling Rows and Columns

### DIFF
--- a/frontend/demo/component/grid/grid-drag-rows-between-grids.ts
+++ b/frontend/demo/component/grid/grid-drag-rows-between-grids.ts
@@ -5,11 +5,7 @@ import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js'; // hi
 import { css, html, LitElement, render } from 'lit';
 import { customElement, state } from 'lit/decorators.js';
 import '@vaadin/vaadin-grid/vaadin-grid';
-import type {
-  GridDragStartEvent,
-  GridDropEvent,
-  GridItemModel,
-} from '@vaadin/vaadin-grid/vaadin-grid';
+import type { GridDragStartEvent, GridItemModel } from '@vaadin/vaadin-grid/vaadin-grid';
 import { getPeople } from 'Frontend/demo/domain/DataService';
 import Person from 'Frontend/generated/com/vaadin/demo/domain/Person';
 import { applyTheme } from 'Frontend/generated/theme';
@@ -74,19 +70,16 @@ export class Example extends LitElement {
           drop-mode="on-grid"
           @grid-dragstart="${this.startDraggingItem}"
           @grid-dragend="${this.clearDraggedItem}"
-          @grid-drop="${(event: GridDropEvent<Person>) => {
-            const { dropTargetItem } = event.detail;
+          @grid-drop="${() => {
             const draggedPerson = this.draggedItem as Person;
-            if (dropTargetItem !== draggedPerson) {
-              const draggedItemIndex = this.grid2Items.indexOf(draggedPerson);
-              if (draggedItemIndex >= 0) {
-                // remove the item from its previous position
-                this.grid2Items.splice(draggedItemIndex, 1);
-                // re-assign the array to refresh the grid
-                this.grid2Items = [...this.grid2Items];
-                // re-assign the array to refresh the grid
-                this.grid1Items = [...this.grid1Items, draggedPerson];
-              }
+            const draggedItemIndex = this.grid2Items.indexOf(draggedPerson);
+            if (draggedItemIndex >= 0) {
+              // remove the item from its previous position
+              this.grid2Items.splice(draggedItemIndex, 1);
+              // re-assign the array to refresh the grid
+              this.grid2Items = [...this.grid2Items];
+              // re-assign the array to refresh the grid
+              this.grid1Items = [...this.grid1Items, draggedPerson];
             }
           }}"
         >
@@ -103,19 +96,16 @@ export class Example extends LitElement {
           drop-mode="on-grid"
           @grid-dragstart="${this.startDraggingItem}"
           @grid-dragend="${this.clearDraggedItem}"
-          @grid-drop="${(event: GridDropEvent<Person>) => {
-            const { dropTargetItem } = event.detail;
+          @grid-drop="${() => {
             const draggedPerson = this.draggedItem as Person;
-            if (dropTargetItem !== draggedPerson) {
-              const draggedItemIndex = this.grid1Items.indexOf(draggedPerson);
-              if (draggedItemIndex >= 0) {
-                // remove the item from its previous position
-                this.grid1Items.splice(draggedItemIndex, 1);
-                // re-assign the array to refresh the grid
-                this.grid1Items = [...this.grid1Items];
-                // re-assign the array to refresh the grid
-                this.grid2Items = [...this.grid2Items, draggedPerson];
-              }
+            const draggedItemIndex = this.grid1Items.indexOf(draggedPerson);
+            if (draggedItemIndex >= 0) {
+              // remove the item from its previous position
+              this.grid1Items.splice(draggedItemIndex, 1);
+              // re-assign the array to refresh the grid
+              this.grid1Items = [...this.grid1Items];
+              // re-assign the array to refresh the grid
+              this.grid2Items = [...this.grid2Items, draggedPerson];
             }
           }}"
         >
--- a/frontend/demo/component/grid/grid-row-reordering.ts
+++ b/frontend/demo/component/grid/grid-row-reordering.ts
@@ -4,6 +4,7 @@ import '@vaadin/vaadin-template-renderer/src/vaadin-template-renderer.js'; // hi
 
 import { html, LitElement, render } from 'lit';
 import { customElement, state } from 'lit/decorators.js';
+import '@vaadin/vaadin-avatar/vaadin-avatar';
 import '@vaadin/vaadin-grid/vaadin-grid';
 import type {
   GridDragStartEvent,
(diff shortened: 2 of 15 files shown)


src/main/java/com/vaadin/demo/component/grid/PersonWithRating.java
```diff
@@ -0,0 +1,55 @@
...
+        result.setSubscriber(person.isSubscriber());
+        result.setMembership(person.getMembership());
+        result.setPictureUrl(person.getPictureUrl());
+        result.setProfession(person.getProfession());
+        result.setAddress(person.getAddress());
+        result.setManagerId(person.getManagerId());
+        result.setManager(person.isManager());
+        result.setStatus(person.getStatus());
+
+        double rating = ThreadLocalRandom.current().nextDouble(0, 10);
```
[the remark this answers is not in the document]
sissbruecker: @Lift-dev ignore

frontend/demo/component/grid/grid-drag-rows-between-grids.ts
```diff
@@ -74,19 +70,16 @@ export class Example extends LitElement {
           drop-mode="on-grid"
           @grid-dragstart="${this.startDraggingItem}"
           @grid-dragend="${this.clearDraggedItem}"
-          @grid-drop="${(event: GridDropEvent<Person>) => {
-            const { dropTargetItem } = event.detail;
```
sissbruecker, line 78: There is no drop target item when `drop-mode="on-grid"`

frontend/demo/component/grid/grid-drag-rows-between-grids.ts
```diff
@@ -103,19 +96,16 @@ export class Example extends LitElement {
...
```
sissbruecker, line 107: Same here

src/main/java/com/vaadin/demo/component/grid/GridBufferedInlineEditor.java
```diff
@@ -0,0 +1,107 @@
...
+            firstNameValidationMessage.setText("");
+            lastNameValidationMessage.setText("");
+            emailValidationMessage.setText("");
+        });
+
+        List<Person> people = DataService.getPeople();
+        grid.setItems(people);
+
+        getThemeList().clear();
+        getThemeList().add("spacing-s");
```
sissbruecker, line 99: Could not find any constants for spacing variants

src/main/java/com/vaadin/demo/component/grid/GridDragDropFilters.java
```diff
@@ -0,0 +1,89 @@
...
+                    .equals(draggedPerson.getManagerId());
+
+            if (newManager == null || isSameManager)
+                return;
+
+            draggedPerson.setManagerId(newManager.getId());
+            treeData.removeItem(draggedPerson);
+            treeData.addItem(newManager, draggedPerson);
+
+            treeDataProvider.refreshAll();
```
sissbruecker, line 61: Ideally this should only refresh the two root nodes affected by the change using `refreshItem`, but that did not work.

Also the refresh has a nasty visual effect, where the child nodes are removed and then a split second later (after re-fetching the data) are added again. Checked the old demo, which has the same effect. Searched for this a bit and there are several discussions about TreeGrid data updates not being super reliable or broken. I think this is the best we can do.

src/main/java/com/vaadin/demo/component/grid/GridDragRowsBetweenGrids.java
```diff
@@ -0,0 +1,91 @@
...
+        grid2.addDropListener(e -> {
+            dataView1.removeItem(draggedPerson);
+            dataView2.addItem(draggedPerson);
+        });
+        grid2.addDragEndListener(this::handleDragEnd);
+        // end::snippet[]
+
+        Div container = new Div(grid1, grid2);
+        setGridStyles(grid1);
+        setGridStyles(grid2);
```
sissbruecker, line 55: Can move these to `setupGrid`

src/main/java/com/vaadin/demo/component/grid/GridUnbufferedInlineEditor.java
```diff
@@ -0,0 +1,101 @@
...
+        getThemeList().clear();
+        getThemeList().add("spacing-s");
+        add(grid, firstNameValidationMessage, lastNameValidationMessage,
+                emailValidationMessage);
+    }
+
+    private static void addCloseHandler(Component textField,
+            Editor<Person> editor) {
+        textField.getElement().addEventListener("keydown", e -> editor.cancel())
+                .setFilter("event.code === 'Escape'");
```
sissbruecker, line 95: Chose to only use <kbd>Escape</kbd> here, instead of using <kbd>Enter</kbd>. Enter might suggest that it commits the change, whereas unbuffered always commits on blur.

src/main/java/com/vaadin/demo/component/grid/PersonWithRating.java
```diff
@@ -0,0 +1,55 @@
+package com.vaadin.demo.component.grid;
+
+import com.vaadin.demo.domain.Person;
+
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.util.concurrent.ThreadLocalRandom;
+
+public class PersonWithRating extends Person {
```
sissbruecker, line 9: This is 100% plumbing IMO, so I extracted it and did not list the file in the example.

src/main/java/com/vaadin/demo/component/grid/ValidationMessage.java
```diff
@@ -0,0 +1,35 @@
+package com.vaadin.demo.component.grid;
+
+import com.vaadin.flow.component.HasText;
+import com.vaadin.flow.component.html.Span;
+import com.vaadin.flow.component.icon.Icon;
+import com.vaadin.flow.component.icon.VaadinIcon;
+import com.vaadin.flow.component.orderedlayout.HorizontalLayout;
+
+class ValidationMessage extends HorizontalLayout implements HasText {
```
sissbruecker, line 9: Also extracted this one, since its purpose is also jsut facilitating the example. However decided to list the file in the example, since it might be useful for someone.

frontend/demo/component/grid/grid-row-reordering.ts
```diff
@@ -50,16 +51,15 @@ export class Example extends LitElement {
...
-            const draggedItemIndex = this.items.indexOf(draggedPerson);
+            const draggedItemIndex = this.items.indexOf(this.draggedItem);
             this.items.splice(draggedItemIndex, 1);
             // re-insert the item at its new position
             const dropIndex =
               this.items.indexOf(dropTargetItem) + (dropLocation === 'below' ? 1 : 0);
-            this.items.splice(dropIndex, 0, draggedPerson);
+            this.items.splice(dropIndex, 0, this.draggedItem);
             // re-assign the array to refresh the grid
             this.items = [...this.items];
```
fhdhsni, line 64: not exactly related to this PR but can be simplified to sth like:
```
            this.items = this.items
                .filter(item => item !== this.draggedItem)
                .flatMap((item: Person) => {
                  if(item !== dropTargetItem) return item;
                  return dropLocation === 'below' ? [dropTargetItem, this.draggedItem] : [this.draggedItem, dropTargetItem]; 
            });
```

src/main/java/com/vaadin/demo/component/grid/GridDragDropFilters.java
```diff
@@ -0,0 +1,89 @@
...
+        treeGrid.addColumn(Person::getLastName).setHeader("Last name");
+        treeGrid.addColumn(Person::getEmail).setHeader("Email");
+
+        return treeGrid;
+    }
+
+    private List<Person> getStaff(Person manager) {
+        return people.stream()
+                .filter(p -> manager.getId().equals(p.getManagerId()))
+                .collect(Collectors.toList());
```
fhdhsni, line 83: nit:
we can do the following to avoid going through all the items for each `manager`
```
private Map<Integer, List<Person>> staffGroupedByMangers;

public GridDragDropFilters() {
    staffGroupedByMangers =
                    people.stream()
                            .filter(person -> person.getManagerId() != null)
                            .collect(Collectors.groupingBy(Person::getManagerId, Collectors.toList()));
}

 private List<Person> getStaff(Person manager) {
      return staffGroupedByMangers.getOrDefault(manager.getId(), Collections.emptyList());
}

File: src/main/java/com/vaadin/demo/component/grid/GridStyling.java [hunk 12]
```diff
@@ -0,0 +1,48 @@
...
+        grid.addColumn(PersonWithRating::getProfession).setHeader("Profession");
+        grid.addColumn(PersonWithRating::getFormattedRating)
+                .setHeader("Customer rating (0-10)");
+
+        grid.setClassNameGenerator(person -> {
+            if (person.getRating() >= 8)
+                return "high-rating";
+            if (person.getRating() <= 4)
+                return "low-rating";
+            return "";
```
fhdhsni, line 29: can we return `null`?

src/main/java/com/vaadin/demo/component/grid/GridRowReordering.java
```diff
@@ -0,0 +1,81 @@
...
+import com.vaadin.flow.data.renderer.TemplateRenderer;
+import com.vaadin.flow.router.Route;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@Route("grid-row-reordering")
+public class GridRowReordering extends Div {
+
+    private Person draggedPerson;
```
fhdhsni, line 20: `draggedItem`? `draggedPerson` sounds funny :)

frontend/demo/component/grid/grid-row-reordering.ts
sissbruecker: Mhh, it's more compact but the `flatMap` adds more complexity for the average reader IMO - they are more likely to be familiar with `splice` rather than `flatMap`. So I would suggest to leave as is.

Clever solution through!

src/main/java/com/vaadin/demo/component/grid/GridDragDropFilters.java
sissbruecker: Done

src/main/java/com/vaadin/demo/component/grid/GridRowReordering.java
sissbruecker: Fair enough 😄. I wanted to make the name more domain related, but maybe that's going to far.

src/main/java/com/vaadin/demo/component/grid/GridStyling.java
sissbruecker: Yes, seems to work 👍

fhdhsni (approved):

PR #631, opened 2021-07-07
docs: add fourth batch of Flow Grid examples

sissbruecker:
## Description

Adds Flow examples for the Grid component for the theme variants section.

### DIFF
--- a/src/main/java/com/vaadin/demo/component/grid/GridColumnAlignment.java
+++ b/src/main/java/com/vaadin/demo/component/grid/GridColumnAlignment.java
@@ -12,7 +12,6 @@ import com.vaadin.flow.component.grid.ColumnTextAlign;
 import com.vaadin.flow.component.grid.Grid;
 import com.vaadin.flow.component.html.Div;
 import com.vaadin.flow.data.renderer.LocalDateRenderer;
-import com.vaadin.flow.data.renderer.TextRenderer;
 import com.vaadin.flow.router.Route;
 import com.vaadin.demo.DemoExporter; // hidden-source-line
 import com.vaadin.demo.domain.DataService;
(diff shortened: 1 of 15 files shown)


src/main/java/com/vaadin/demo/component/grid/GridColumnAlignment.java
```diff
@@ -27,8 +26,7 @@ public GridColumnAlignment() {
                 GridColumnAlignment::generateRandomDueDate, "MM/dd/yyyy"))
                 .setHeader("Due");
         // tag::snippet[]
-        grid.addColumn(new TextRenderer<>(
-                GridColumnAlignment::generateRandomAmountText))
+        grid.addColumn(GridColumnAlignment::generateRandomAmountText)
```
sissbruecker, line 29: Removed usages of TextRenderer from other Grid examples.

src/main/java/com/vaadin/demo/component/grid/GridWrapCellContent.java
```diff
@@ -0,0 +1,53 @@
...
+        Address address = person.getAddress();
+
+        return String
+                .format("%s %s %s %s", address.getStreet(), address.getCity(),
+                        address.getZip(), address.getState());
+    }
+
+    private static TemplateRenderer<Person> createAvatarRenderer() {
+        return TemplateRenderer.<Person>of(
+                "<vaadin-avatar img=\"[[item.pictureUrl]]\" name=\"[[item.fullName]]\" alt=\"User avatar\"></vaadin-avatar>")
```
vursen, line 46: Do we prefer `TemplateRenderer` rather than `ComponentRenderer` here for some performance reasons?

vursen (approved):

src/main/java/com/vaadin/demo/component/grid/GridWrapCellContent.java
sissbruecker: @vursen Yes, ComponentRenderer needs to create/manage actual component instances server-side, also it requires a server round-trip to render the components. In general prefer TemplateRenderer (later LitRenderer), though ComponentRenderer is easier to use when you need to handle some events. So if there is interaction in an example it's OK to use ComponentRenderer if it makes the example simpler.

PR #636, opened 2021-07-08
mention the scroll to index issue in TreeGrid

sosa-vaadin:
## Description

Add a mention to an issue with the `scrollToIndex(index)` method in TreeGrid.

Related issue: https://github.com/vaadin/flow-components/issues/1016

### DIFF
--- a/articles/ds/components/tree-grid/index.asciidoc
+++ b/articles/ds/components/tree-grid/index.asciidoc
@@ -68,15 +68,15 @@ You can scroll to a specific item by calling the [methodname]`scrollToIndex(inde
 However, the effective index of an item changes depending on which nodes are expanded or collapsed and other factors.
 What portion of the grid is visible also affects how an item's effective index is calculated.
 
-Scrolling right after expanding or collapsing nodes in a single action (for example, a button click) will result in a scroll action before children of expanded nodes are rendered and children of collapsed nodes removed.
-Since the effective index of some items changes after the expand or collapse, the grid might have ended up scrolling to a different item than intended.
+.Index of Items Can Vary
+CAUTION: Items in a TreeGrid are not guaranteed to hold a specific index exclusively.
 
 The `scrollToIndex(index)` method was designed to work in a pre-order traversal method.
 Yet, TreeGrid doesn't know about all children (or the children of those children) a node has.
-In addition, nodes are loaded lazily for performance reasons, increasing the lack of knowledge of the complete structure of the tree.
+In addition, nodes are loaded in a lazy fashion for performance reasons, increasing the lack of knowledge of the complete structure of the tree.
 
-.Index of Items Can Vary
-CAUTION: Items in a TreeGrid are not guaranteed to hold a specific index exclusively.
+Scrolling right after expanding or collapsing nodes in a single action (for example, a button click) results in a scroll action before children of expanded nodes are rendered and children of collapsed nodes removed.
+Since the effective index of some items changes after the expand or collapse, the grid might have ended up scrolling to a different item than intended.
 
 == Rich Content
 
--- a/articles/ds/components/tree-grid/index.asciidoc
+++ b/articles/ds/components/tree-grid/index.asciidoc
@@ -91,14 +91,13 @@ Tree Grid is not meant to be used as a navigation menu.
 
 === Avoid Using Scroll to Index
 
-The effective index of an item changes depending on which nodes are expanded or collapsed and other factors.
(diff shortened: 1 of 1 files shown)


articles/ds/components/tree-grid/index.asciidoc
```diff
@@ -89,6 +89,17 @@ include::{root}/src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags
 
 Tree Grid is not meant to be used as a navigation menu.
 
+=== Avoid Using Scroll to Index
+
+In contrast to Grid, the `scrollToIndex(index)` method in TreeGrid was designed to work in a pre-order traversal method.
+The effective index of an item depends on the complete hierarchy of the tree.
+Nodes are loaded in a lazy fashion for performance reasons.
+Without the knowledge of the complete hierarchy, TreeGrid can't reliably calculate an exact scroll index.
```
sissbruecker, line 97:
```suggestion
TreeGrid uses lazy loading for performance reasons and does not know about the complete hierarchy.
Without the knowledge of the complete hierarchy, TreeGrid can't reliably calculate an exact scroll position.
```

articles/ds/components/tree-grid/index.asciidoc
```diff
@@ -89,6 +89,17 @@ include::{root}/src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags
...
+
+In contrast to Grid, the [methodname]#scrollToIndex(index)# method in TreeGrid works in a pre-order traversal method.
+The effective index of an item depends on the complete hierarchy of the tree.
+TreeGrid uses lazy loading for performance reasons and does not know about the complete hierarchy.
+Without the knowledge of the complete hierarchy, TreeGrid can't reliably calculate an exact scroll position.
+
+.Index of Items Can Vary
+CAUTION: Items in a TreeGrid are not guaranteed to hold a specific index exclusively.
+The behaviour of *this method in TreeGrid* is not deterministic and *should be avoided*.
+
```
magi42, line 102:
```suggestion
The [methodname]#scrollToIndex()# method in [classname]#TreeGrid# works by traversing through the Tree Grid according to their hierarchical order as they would be displayed on the screen, row by row.
However, Tree Grid uses lazy loading for performance reasons and does not know about the complete hierarchy.
Without the knowledge of the complete hierarchy, Tree Grid cannot reliably calculate an exact scroll position.
Therefore, the effective index of an item depends on the complete hierarchy of the tree.

.Index of items can vary
CAUTION: Items in a Tree Grid are not guaranteed to hold a specific index exclusively.
The behavior of the [methodname]#scrollToIndex()# method in Tree Grid *is not deterministic* and *should be avoided*.

```

magi42 (approved):
The "pre-order traverse" is a pretty deep technical term, so I simplified it. Please check that it's still technically correct.

articles/ds/components/tree-grid/index.asciidoc
```diff
@@ -89,6 +89,17 @@ include::{root}/src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags
 
 Tree Grid is not meant to be used as a navigation menu.
 
+=== Avoid Using Scroll to Index
```
rolfsmeds, line 92: This feels like an unncessarily detailed explanation of _why_ scrollToIndex is problematic. We really just need to explain that scrollToIndex is not reliable and thus should be avoided.

articles/ds/components/tree-grid/index.asciidoc
```diff
@@ -89,6 +89,17 @@ include::{root}/src/main/java/com/vaadin/demo/domain/Person.java[group=Java,tags
...
+
+The [methodname]#scrollToIndex()# method in [classname]#TreeGrid# works by traversing through the Tree Grid according to their hierarchical order as they would be displayed on the screen, row by row.
+However, Tree Grid uses lazy loading for performance reasons and does not know about the complete hierarchy.
+Without the knowledge of the complete hierarchy, Tree Grid cannot reliably calculate an exact scroll position.
+Therefore, the effective index of an item depends on the complete hierarchy of the tree.
+
+.Index of items can vary
```
rolfsmeds, line 99: This admonition title doesn't do a very good job of summarizing the problem, which is that scrollToIndex is not reliable.

I propose omitting the long explanation above in favor of an admonition like:

> .scrollToIndex not reliable
> CAUTION: The behavior of the scrollToIndex method in Tree Grid is *not deterministic* due to lazy loading hierarchical data. Usage of this method is not recommended.

rolfsmeds (changes requested):

rolfsmeds (approved):

PR #637, opened 2021-07-09
doc: Added note about z-indeces incompatibility when using MPR

TatuLund:


magi42 (approved):
I simplified the heading a bit as Vaadin 7/8 is implied with MPR. Use single sentence per line.

PR #640, opened 2021-07-10
Revised Fusion application basics

magi42:
This is somewhat WIP, but I'd suggest merging now and then improve further later. There's some revision of Fusion basics, with the reactive intro moved from the tutorial, some stuff moved to a separate Configuring article.

I noticed a conflict with an updated example, but I moved the original example to a separate file. Please check if there's something wrong with my example.


frontend/demo/fusion/application/basic/my-view.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
+    `;
+  }
+
+  render() {
+    return html`
+      <vaadin-button theme="primary" @click="${this.clickHandler}"> Click me </vaadin-button>
+      <vaadin-notification id="notification" duration="2000">
+        <template>Hello, World!</template>
+      </vaadin-notification>
```
haijian-vaadin, line 27: `<vaadin-notification>` component doesn't work in the demo, better to replace with a simple alert, as in https://vaadin.com/docs/latest/fusion/application/events.

articles/fusion/application/basics.asciidoc
```diff
@@ -6,9 +6,61 @@ layout: page
...
+The following example is given in both imperative (Flow/Java) and reactive (Fusion/TypeScript):
+
+[.example.show-code]
+--
+
+In imperative UI programming, like with Vaadin Flow, there are two states: the data model and the UI. 
+It is your responsibility as a developer to keep them in sync. 
+You need to keep references to elements so you can update their values when the underlying state (the `Contact`) changes. 
+
+[source,java]
```
haijian-vaadin, line 30: For some reason, the java example doesn't work, it only shows blank.

articles/fusion/application/basics.asciidoc
```diff
@@ -17,175 +69,64 @@ You do not normally need to modify the file.
 `frontend/index.ts`::
 The TypeScript entry point of the application.
 This is where you set up routing and other application tasks.
+To navigate between views, you need to define routes (see <<../routing/overview#,Defining Routes>> for more details).
 
-[role="since:com.vaadin:vaadin@15"]
-== Prerequisites
+The routing configuration also starts a server session and loads any server-side views implemented with Vaadin Flow.
```
haijian-vaadin, line 74: It kinda indicates that Fusion would always create a session. It only creates one if there are server-side views, if not, it would stateless (from v22).

haijian-vaadin (changes requested):

articles/fusion/application/basics.asciidoc
magi42: Rephrased.

frontend/demo/fusion/application/basic/my-view.ts
magi42: Ok, using that.

haijian-vaadin (approved):

PR #648, opened 2021-07-15
feat: add SplitLayout Flow examples

fhdhsni:


src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
```diff
@@ -1,21 +1,61 @@
...
+        splitLayout.setMaxHeight("280px");
+        add(splitLayout);
+    }
+
+    // tag::setstyles[]
+    private void setStyles() {
+        button.setIcon(sidebarCollapsed ? rightArrowIcon : leftArrowIcon);
+        int sidebarWidthPercentage = sidebarCollapsed ? 13 : 40;
+        masterContainer.getElement().executeJs("this.style = `overflow: hidden; width: ${$0}%`", sidebarWidthPercentage);
+        detailContent.getElement().executeJs("this.style = `width: ${$0}%`",100 - sidebarWidthPercentage);
```
fhdhsni, line 56: There should be a better approach I think. Any idea?

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
```diff
@@ -1,21 +1,61 @@
...
+        MasterContent masterContent = new MasterContent();
+
+        sidebarCollapsed = false;
+
+        button.addClickListener(event -> {
+            sidebarCollapsed = !sidebarCollapsed;
+            setStyles();
+        });
+        button.getElement().setAttribute("aria-label", "Expand/collapse sidebar");
+        button.addThemeVariants(ButtonVariant.LUMO_TERTIARY, ButtonVariant.LUMO_ICON);
```
vursen, line 37: Setting the Lumo Icon theme may be redundant here. 

`setStyles()` calls `button.setIcon(...)` which automatically [adds](https://github.com/vaadin/flow-components/blob/65ccf0b22f24053d922e0db7c25d04752daf8ecb/vaadin-button-flow-parent/vaadin-button-flow/src/main/java/com/vaadin/flow/component/button/Button.java#L208) the Lumo Icon theme to the button if no other content is provided except the icon.

```suggestion
        button.addThemeVariants(ButtonVariant.LUMO_TERTIARY);
```

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
vursen: Have you tried to change styles via `getStyle().set(...)`? Would it work?

```suggestion
        detailContent.getStyle().set("width", 100 - sidebarWidthPercentage + "%")
```

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
```diff
@@ -1,21 +1,61 @@
...
+
+        SplitLayout splitLayout = new SplitLayout(masterContainer, detailContent);
         // end::snippet[]
+
+        splitLayout.setMaxHeight("280px");
+        add(splitLayout);
+    }
+
+    // tag::setstyles[]
+    private void setStyles() {
```
vursen, line 52: I'm wondering if it would make sense to name this method somehow like `setSidebarCollapsed`?

Example:
```java
private void setSidebarCollapsed(value) {
  sidebarCollapsed = value;
  
  // Update styles...
}
```

Although, I'm not sure 100% if this change will make things better. What do you think?

articles/ds/components/split-layout/index.asciidoc
```diff
@@ -150,21 +100,15 @@ This is useful when the user wants to toggle between certain positions.
 include::{root}/frontend/demo/component/splitlayout/split-layout-toggle.ts[render,tags=snippet,indent=0,group=TypeScript]
 ----
 
-[source,typescript]
-----
-include::{root}/frontend/demo/component/splitlayout/master-content.ts[group=TypeScript]
```
vursen, line 155: As I understand, at this moment `master-content.ts` and `detail-content.ts` are not shown anywhere which may confuse the user since these components are pretty often used in the examples. Also the same for `MasterContent.java`, `DetailContent.java`.

I guess it would be nice to have these components shown in the examples as well.

@sissbruecker @fhdhsni What do you think?

fhdhsni: Yeah I thought `split-layout` is a simple component and it's obvious that `master-content` and `detail-content` are placeholders. And code-wise they don't have anything useful. If I include their TS files I also should include their Java counterparts, which is:
```
public class MasterContent extends Component implements HasSize {}
```
again doesn't contain anything useful related to `split-layout`

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
fhdhsni: that doesn't work because of the styles added via the client when you change the divider. I need to remove that style and set my own style, but couldn't find a way.

vursen: I checked the branch locally and I found that the splitter (divider) manages the content width via `flex: 1 1 ...px`. 

Maybe you could use `flex: 1 1 ${percentage}%` rather than `width: ${percentage}%`.

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
vursen: At least, I would prefer `setStyles()` not to set/update the button icon as this action doesn't have anything with styles.

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
vursen: I gave it a shot, and it seems to work at first sight.

```java
int sidebarWidthPercentage = sidebarCollapsed ? 13 : 40;

masterContainer.getElement().getStyle().set("flex", "1 1 " + sidebarWidthPercentage + "%");
detailContent.getElement().getStyle().set("flex", "1 1 " + (100 - sidebarWidthPercentage) + "%");
```

fhdhsni: that works but it diverges from other examples (they're all using `width`) and it kinda relies on the implementation details. IDK. @sissbruecker any idea?

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
fhdhsni: yeah `setStyles` is not a good name. How about `update()`?
Also If you think it'd be better to haver a setter for `sidebarCollapsed` we can define a separate function for that

sissbruecker: I would suggest `updateSidebar`

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
sissbruecker: The Flow component has a dedicated API for setting the splitter position (sets the size in percentage for the first child):
```
splitLayout.setSplitterPosition(sidebarCollapsed ? 13 : 40);
```

Note that the Typescript example here is flawed: custom splitter positions set by the user will be overriden if there is a change to some other state in the component that triggers a rerender. To solve it, instead of binding the style attribute to an expression, the callback for the button click could manipulate the style on the element.

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutInitialSplitterPosition.java
```diff
@@ -12,9 +10,17 @@
 
     public SplitLayoutInitialSplitterPosition() {
         // tag::snippet[]
-        SplitLayout splitLayout = new SplitLayout(new Span("Side 1"), new Span("Side 2"));
-        add(splitLayout);
+        MasterContent master = new MasterContent();
+        master.setWidth("50%");
+
+        DetailContent detail = new DetailContent();
+        detail.setWidth("50%");
```
sissbruecker, line 17: The Flow component has a dedicated API for setting the splitter position (sets the size in percentage for the first child):
```
splitLayout.setSplitterPosition(50);
```

fhdhsni: done

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
fhdhsni: done

src/main/java/com/vaadin/demo/component/splitlayout/SplitLayoutToggle.java
fhdhsni: done

sissbruecker (approved):

PR #649, opened 2021-07-15
docs: add second batch of Notification Flow examples

vursen:
## Description

This PR adds Flow examples for the Notification component.

## Sections
- [x] Interactive Elements


src/main/java/com/vaadin/demo/component/notification/NotificationKeyboardA11y.java
```diff
@@ -0,0 +1,85 @@
...
+    notification.addThemeVariants(NotificationVariant.LUMO_CONTRAST);
+
+    Div statusText = new Div(new Text("5 tasks deleted"));
+
+    Button undoButton = new Button();
+    undoButton.addThemeVariants(ButtonVariant.LUMO_PRIMARY);
+    undoButton.getElement().getStyle().set("margin-left", "var(--lumo-space-xl)");
+    undoButton.getElement().executeJs(
+      "const isMac = /Macintosh|MacIntel|MacPPC|Mac68K/.test(window.navigator.platform);" +
+      "this.textContent = `Undo ${isMac ? '⌘' : 'Ctrl-'}Z`;"
```
vursen, line 50: Decided not to use `<span aria-hidden="true" />` here as the Flow Button component doesn't have a public API to add rich content.

frontend/demo/component/notification/notification-keyboard-a11y.ts
```diff
@@ -22,50 +26,77 @@ export class Example extends LitElement {
...
+          <div>5 tasks deleted</div>
+          <vaadin-button
+            style="margin-left: var(--lumo-space-xl);"
+            theme="primary"
+            @click="${() => (this.notificationOpened = false)}"
+          >
+            Undo
+            <!-- Ideally, this should be hidden if the
+                 device does not have a physical keyboard -->
+            ${this.isMac ? '⌘' : 'Ctrl-'}Z
```
vursen, line 70: The Flow Button component doesn't have a public API for adding rich content wherefore I decided to remove `<span aria-hidden="true" />` from here as well.

frontend/demo/component/notification/notification-keyboard-a11y.ts
```diff
@@ -22,50 +26,77 @@ export class Example extends LitElement {
...
+  connectedCallback() {
+    super.connectedCallback();
+
+    document.addEventListener('keydown', this.onKeyDown);
+  }
+
+  disconnectedCallback() {
+    super.disconnectedCallback();
+
+    document.removeEventListener('keydown', this.onKeyDown);
```
vursen, line 90: The previous version of the example used to not remove the global `keydown` event listener that could lead to a memory leak.

fhdhsni (approved):

fhdhsni:
changed [`NotificationKeyboardA11y`](https://github.com/vaadin/docs/pull/649/commits/0077480119f8376627a9e7dbd3ee72b5794c2d77) to use the `notification` as the lifecycle owner when setting up shortcuts. Before that it wasn't removing the listener because `this` was never detaching. It was also preventing the TS example from receiving the shortcuts because `Shortcuts.addShortcutListener` prevents propagation under the hood.

src/main/java/com/vaadin/demo/component/notification/NotificationUndo.java
```diff
@@ -0,0 +1,67 @@
...
+    add(button);
+  }
+
+  public Notification show() {
+    // tag::snippet[]
+    Notification notification = new Notification();
+    notification.setDuration(10000);
+    notification.addThemeVariants(NotificationVariant.LUMO_CONTRAST);
+
+    Div statusText = new Div(new Text("5 tasks deleted"));
```
sissbruecker, line 37: Span would be more appropriate IMO

src/main/java/com/vaadin/demo/component/notification/NotificationLink.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  public Notification show() {
+    // tag::snippet[]
+    // When creating a notification using the constructor,
+    // the duration is 0-sec by default which means that
+    // the notification does not close automatically.
+    Notification notification = new Notification();
+
+    Div text = new Div(
```
sissbruecker, line 37: Span would be more appropriate IMO

src/main/java/com/vaadin/demo/component/notification/NotificationKeyboardA11y.java
sissbruecker: Using the unicode `⌘` is problematic for an A11y example. According to an onlline search that character is:
```
U+2318 : PLACE OF INTEREST SIGN {operating system key (ISO 9995-7)}
```
Announcing that would be super confusing. Options:
- give rich content another try, and hide the shortcut from being announced for screen readers
- add an aria-label without the shortcut to the button and check that it overrides the text content

Maybe have a quick chat with Serhii about this, he probably has the most experience in the team ATM.

Also we need to align the TS example if we change this.

src/main/java/com/vaadin/demo/component/notification/NotificationKeyboardA11y.java
```diff
@@ -0,0 +1,73 @@
...
+    add(button);
+  }
+
+  // tag::snippet[]
+  public Notification show() {
+    Notification notification = new Notification();
+    notification.setDuration(10000);
+    notification.addThemeVariants(NotificationVariant.LUMO_CONTRAST);
+
+    Div statusText = new Div(new Text("5 tasks deleted"));
```
sissbruecker, line 37: Span would be more appropriate IMO

src/main/java/com/vaadin/demo/component/notification/NotificationKeyboardA11y.java
```diff
@@ -0,0 +1,73 @@
...
+    // tag::snippet[]
+
+    return notification;
+  }
+  // end::snippet[]
+
+  // tag::setupUndoShortcut[]
+  public void setupUndoShortcut(Notification notification) {
+    Shortcuts.addShortcutListener(notification, notification::close, Key.KEY_Z, KeyModifier.META);
+    Shortcuts.addShortcutListener(notification, notification::close, Key.KEY_Z, KeyModifier.CONTROL);
```
sissbruecker, line 67: This isn't actually working for me. Tested on macOS, Chrome + FF

Wait, the shortcut for me is actually <key>Cmd+Y</key>. So it's due to the keyboard layout (I'm using ISO-de). Maybe we can do a quick check if that can be fixed, if not we can file a bug in Flow.

src/main/java/com/vaadin/demo/component/notification/NotificationRetry.java
```diff
@@ -0,0 +1,69 @@
...
+
+  private Notification show() {
+    // tag::snippet[]
+    // When creating a notification using the constructor,
+    // the duration is 0-sec by default which means that
+    // the notification does not close automatically.
+    Notification notification = new Notification();
+    notification.addThemeVariants(NotificationVariant.LUMO_ERROR);
+
+    Div statusText = new Div(new Text("Failed to generate report"));
```
sissbruecker, line 39: Span would be more appropriate IMO

src/main/java/com/vaadin/demo/component/notification/NotificationUndo.java
fhdhsni: As far as I know semantically doesn't make any difference. Also if we go with a `Span`,`margin-left` should be applied manually. `p` makes more sense but again needs additional styling.

src/main/java/com/vaadin/demo/component/notification/NotificationKeyboardA11y.java
fhdhsni: Since this is only used for macs and Voice over announces that as `command` I think it's fine.

src/main/java/com/vaadin/demo/component/notification/NotificationKeyboardA11y.java
fhdhsni: It should be working now.
done.

src/main/java/com/vaadin/demo/component/notification/NotificationLink.java
fhdhsni: see above. let me know if you think using `p` worth it (because of the additional styling)

sissbruecker (approved):

PR #655, opened 2021-07-22
refactor: move Flow-specific imports to central place

sissbruecker:
## Description

Currently, when using Flow components that import JS modules through `@JsImport`, these imports are not added automatically to the build, and instead are added manually, to each corresponding TS example when needed in a Java example. That is error-prone when implementing new examples and hard to maintain when a `@JsImport` annotation is changed at some point. A recent example is the deprecation of `TemplateRenderer`, which was then added as `@JsImport`, but was not added to dozens of examples which stopped working.

This change moves all Flow component specific JS imports into one central file, which is then imported into every example. This guarantees that these JS modules will always be available in all examples, and makes changes to `@JsImport`s easier to maintain, because only a single file needs to be changed then.

The downside is a larger payload for all pages using examples, around 500kb unminified.

### DIFF
--- a/frontend/demo/init-flow-components.ts
+++ b/frontend/demo/init-flow-components.ts
@@ -1,0 +1,28 @@
+// Imports for Flow component specific JS modules
+// These need to be imported manually, as the
+// webpack setup in this project ignores @JsModule
+// annotations on the Flow components.
+// Rather than adding the relevant imports to each
+// specific example, which is error prone and hard
+// to maintain, we import these modules once
+// globally for all examples.
+
+// General Flow modules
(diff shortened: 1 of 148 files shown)


frontend/demo/init-flow-components.ts
```diff
@@ -0,0 +1,28 @@
+// Imports for Flow component specific JS modules
```
sissbruecker, line 1: All Flow specific imports are added to this file.

The changes in the other files are just about removing the redundant imports that are now imported in this file.

sosa-vaadin (approved):

PR #656, opened 2021-07-22
fix: context menu checkable example being checkable

sissbruecker:
## Description

In the `ContextMenuCheckable.java` Flow example the items in the context menu actually weren't checkable.

- Fixed the Flow example to make the items checkable
- Changed the TS example to remove redundant state and remove logic that would recreate the items array on each render


src/main/java/com/vaadin/demo/component/contextmenu/ContextMenuCheckable.java
```diff
@@ -5,6 +5,7 @@
 import com.vaadin.demo.domain.Person;
 import com.vaadin.flow.component.Html;
 import com.vaadin.flow.component.contextmenu.ContextMenu;
```
sissbruecker, line 7:
```suggestion
import com.vaadin.flow.component.contextmenu.ContextMenu;
```

fhdhsni (approved):

PR #671, opened 2021-07-29
Enhance SEO of tutorials

tarekoraby:
Currently, Vaadin-related search terms return outdated v7 and v8 results for the tutorials. This PR, along with #669 and #670, aims to fix that.

### DIFF
--- a/articles/flow/tutorials/in-depth-course/index.adoc
+++ b/articles/flow/tutorials/in-depth-course/index.adoc
@@ -1,11 +1,11 @@
 ---
 title: In-Depth Course
 order: 20
-page-title: "Tutorial: Building Modern Web Applications With Spring Boot and Vaadin"
+page-title: "Vaadin Tutorial: Building Modern Web Applications With Spring Boot and Vaadin"
 description: This tutorial series teaches you how to build, test, and deploy a modern web app from scratch. You'll learn to use Spring Boot and Vaadin. 
 ---
 
-= Building Modern Web Applications With Spring Boot and Vaadin
+= Vaadin Tutorial: Building Modern Web Applications With Spring Boot and Vaadin
 
 video::quiQg9bgFs8[youtube]
 
--- a/articles/flow/tutorials/quick-start-tutorial.adoc
+++ b/articles/flow/tutorials/quick-start-tutorial.adoc
@@ -2,10 +2,10 @@
 title: Quick Start Tutorial
 order: 10
 layout: page
-page-title: Vaadin Flow Quick Start Tutorial
+page-title: Vaadin Quick Start Tutorial
 description: Learn to build and deploy a modern web application 100% in Java.
 ---
-= Vaadin Flow Quick Start Tutorial
+= Vaadin Quick Start Tutorial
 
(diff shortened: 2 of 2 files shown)


articles/flow/tutorials/in-depth-course/index.adoc
```diff
@@ -1,11 +1,11 @@
 ---
 title: In-Depth Course
 order: 20
-page-title: "Tutorial: Building Modern Web Applications With Spring Boot and Vaadin"
+page-title: "Vaadin Tutorial: Building Modern Web Applications With Spring Boot and Vaadin"
```
marcushellberg, line 4: Vaadin is already in the title. The purpose here was to keep the title approachable also to those who don't know Vaadin. I think having "Vaadin tutorial" as a prefix will make it less likely that people who don't know Vaadin would click on it.

articles/flow/tutorials/quick-start-tutorial.adoc
```diff
@@ -2,10 +2,10 @@
 title: Quick Start Tutorial
 order: 10
 layout: page
-page-title: Vaadin Flow Quick Start Tutorial
+page-title: Vaadin Quick Start Tutorial
```
marcushellberg, line 5: Flow needs to be in the title. There is also a Fusion quick start tutorial, it will be confusing if they have the same title.

articles/flow/tutorials/in-depth-course/index.adoc
tarekoraby: That’s possible, and yours is a valid concern. On the other hand, without this change, google is less likely to suggest this tutorial  to people who google “Vaadin”. The problem with this title is that it’s missing “Vaadin Tutorial”, not just “Vaadin”.

articles/flow/tutorials/quick-start-tutorial.adoc
tarekoraby: Confusing maybe for people sufficiently familiar with Vaadin to know that there is such things as Flow and Fusion, and the differences between them. However, for newbies, adding qualifiers such as Flow will only increase their confusion. And here again, google will less likely promote this result for people just googling “Vaadin” or “Vaadin Tutorial “ if we add “Flow”

tarekoraby: > it will be confusing if they have the same title
The idea is to favor Flow's tutorials for those who only know Vaadin (but not Flow vs. Fusion). For Fusion tutorial, the term "Fusion" should remain.

tarekoraby:
As discussed, I reverted the changes to the in-depth tutorial so that it remains approachable for people not familiar with Vaadin.

tarekoraby:
@magi42 SEO issues were already discussed several times in #vaadincom, with @marcushellberg, and with Ömer Tümer. And as far as I understand, the most effective fix is indeed to change the title and H1 headings. Here we are going to only change the title hoping that this would improve the SEO performance of the Flow quickstart tutorial.

magi42:
Perhaps have the heading "Vaadin Quick Start Tutorial (Flow)"? Or "Vaadin Quick Start Tutorial: Flow".

Hard to say what matters.

PR #686, opened 2021-08-09
chore: update vaadin version to 22.0.0 alpha1

sissbruecker:
## Description

Updates `latest` to use Vaadin platform v22.0.0-alpha1.
- [x] changes `vaadin.version` to `v22.0.0-alpha1`
- [x] contains automated changes to package.json
- [x] updates version selector to include mention of v22
- [x] fixes grid component example that broke due to update
- [x] fixes Fusion examples that broke due to update


jouni (approved):
Can't say much about the TS changes, but otherwise looks fine.

Thanks for fixing the broken Messags cross ref which I introduced!

PR #688, opened 2021-08-09
fix: Prevent Login examples from stealing focus

### LINKED ISSUE #677: Disable autofocus in Login samples to prevent erratic scrolling
As mentioned in https://github.com/vaadin/vaadin-login/issues/121#issuecomment-888206239, the default autofocus in Login component makes[ the docs page](https://vaadin.com/docs/latest/ds/components/login) virtually unusable since the samples keep stealing focus.

`no-autofocus` flag shoud be applied to all samples, together with a comment explaining that it's just there for the docs

jouni:
Prevent Login examples from stealing focus when browsing the documentation.

Fixes #677


jouni:
The Java/Flow API is still missing from the component, but I tentatively added the `setNoAutofocus(true)` method calls in there already.

sissbruecker:
@jouni Since the Java API for that attribute is missing, we could alternatively just use `loginForm.getElement().setAttribute("no-autofocus", "")`, to get the issue fixed ASAP. I think that's reasonable, as the line is not part of the snippet, and we have several other places in the Java examples where we have to drop down to the basic Flow API.

jouni:
Yeah, that makes perfect sense. Didn’t cross my mind.

sissbruecker (approved):
@jouni I adapted the Java examples to use `setAttribute`. I also fixed some examples, where the Java example was rendered in addition to the preview, which caused some blank space.

You can give it a second look and merge if you're fine with the changes.

PR #692, opened 2021-08-11
feat: add information about readonly types in generated TypeScript

Lodin:
Updates the documentation according to the latest Fusion changes.


articles/fusion/application/accessing-backend.asciidoc
```diff
@@ -147,18 +151,170 @@ For example, the generated TypeScript module for the Java endpoint defined in
...
+----
+
+== Entities
+
+Endpoint methods can return or receive as a parameter any data structure (entity).
+In this case, the TypeScript generator creates a TypeScript interface for it.
+Let's take a look at the example.
+
+
+`City.java`
```
haijian-vaadin, line 177: Since these are newly added source code, it's better to put the code into separate files and reference to them via `include::`,  e.g.
```
.`SecurityConfig.java`
[source,java]
----
include::{root}/src/main/java/com/vaadin/demo/fusion/security/authentication/SecurityConfigDemo.java[tags=public-resources]
----
```
in https://github.com/vaadin/docs/edit/latest/articles/fusion/security/spring-login.asciidoc

haijian-vaadin (changes requested):

articles/fusion/application/accessing-backend.asciidoc
Lodin: Looks like code moved to a separate file is compiled as an executable code while my examples are only visual. So I guess it is better to leave inlined.

articles/fusion/application/accessing-backend.asciidoc
```diff
@@ -135,30 +112,64 @@ For more information about supported types, see <<../advanced/type-conversion#,
...
+
+`CountryEndpoint.java`
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/fusion/accessingbackend/CountryEndpoint.java[]
 ----
 
+The TypeScript output is the following:
+
+`City.ts`
```
haijian-vaadin, line 143: Could remove this, it looks redundant on the page.
Screenshot 2021-08-13 at 11 01 13

articles/fusion/application/accessing-backend.asciidoc
```diff
@@ -135,30 +112,64 @@ For more information about supported types, see <<../advanced/type-conversion#,
...
-export async function addOne(number: number) {
-  return await client.call('CounterEndpoint', 'addOne', {number});
-}
+`City.java`
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/fusion/accessingbackend/City.java[]
+----
+
+`CountryEndpoint.java`
```
haijian-vaadin, line 135: Could remove this, it looks redundant on the page.

articles/fusion/application/accessing-backend.asciidoc
```diff
@@ -135,30 +112,64 @@ For more information about supported types, see <<../advanced/type-conversion#,
...
 
+The TypeScript output is the following:
+
+`City.ts`
+[source,typescript]
+----
+include::{root}/frontend/generated/com/vaadin/demo/fusion/accessingbackend/City.ts[]
+----
+
+`CountryEndpoint.ts`
```
haijian-vaadin, line 149: Could remove this, it looks redundant on the page.

articles/fusion/application/accessing-backend.asciidoc
```diff
@@ -135,30 +112,64 @@ For more information about supported types, see <<../advanced/type-conversion#,
...
- *
- * This module has been generated from CounterEndpoint.java
- * @module CounterEndpoint
- */
+include::{root}/frontend/generated/CounterEndpoint.ts[]
+----
+
+== Entities
+
+Endpoint methods can return or receive as a parameter any data structure (entity).
```
haijian-vaadin, line 125:
```suggestion
An Endpoint method can return or receive a parameter as an entity (non-primitive type).
```

articles/fusion/application/accessing-backend.asciidoc
```diff
@@ -135,30 +112,64 @@ For more information about supported types, see <<../advanced/type-conversion#,
...
+In this case, the TypeScript generator creates a TypeScript interface for it.
```
haijian-vaadin, line 126:
```suggestion
In this case, the generator also creates a TypeScript interface for the entity.
```

src/main/java/com/vaadin/demo/fusion/accessingbackend/CountryEndpoint.java
```diff
@@ -0,0 +1,33 @@
...
+            new City("Turku", "Finland"), new City("Berlin", "Germany"),
+            new City("London", "UK"), new City("New York", "USA"));
+
+    /**
+     * A method that returns a collection of entities.
+     */
+    public List<City> getCities(int numberOfCities) {
+        return numberOfCities <= cities.size() ?
+                cities.subList(0, numberOfCities - 1) : cities;
+    }
```
haijian-vaadin, line 25: Maybe we can simplify it a bit the endpoint by only showing the `getCityName()` method? The page is getting long with all the source code.
I understand the point of showing one method using it as param the other as return value, but still maybe one is enough.

articles/fusion/application/accessing-backend.asciidoc
haijian-vaadin: Or maybe we could return the whole CountryEndpoint source code? It's kinda repetitive since we showed the CountEndpoint.
Might be better to keep the Entity section about Entity interface generation, not to repeat the Endpoint generation.

haijian-vaadin (changes requested):

articles/fusion/application/accessing-backend.asciidoc
Lodin: I wanted to show how the endpoint with an entity is generated. I believe it is better to show the code here because only the code for an entity might look unclear.

articles/fusion/application/accessing-backend.asciidoc
Lodin: Oh, nice catch. Why they appeared this way, I wonder.

src/main/java/com/vaadin/demo/fusion/accessingbackend/CountryEndpoint.java
Lodin: Right. I'll try to combine everything in one method then

Lodin: @haijian-vaadin, I replaced unnecessary second method with an internal class. I believe it is another good usecase to show for the user. Unfortunately, the amount of code increased. What do you think about it?

haijian-vaadin:
I think it looks good to me now, just in the read-only part, we could mention one thing, that is, instead of mutating the data, it's better to create a new object, as you showed in the slack discussion.

> you shouldn't mutate any of your data implicitly. Especially if the data you're working with came from the backend. So the best approach here is to create a new object over the clone of a previous one:
> ```
> let user: User = {first: "John", last: "Doe"};
> user = {...user, someOptionalThing: 123};
> ```

haijian-vaadin:
Looks good, as discussed in the daily, should we update the upgrading guide here https://github.com/vaadin/docs/blob/latest/articles/guide/upgrading.asciidoc?

Lodin:
Oh, sure

haijian-vaadin (approved):

PR #695, opened 2021-08-15
Updated HaaS and updated Antlers integration (latest)

paulroemer:
Integration of the HaaSv2 used for the new website design.

- HaaS loader URL changed to <haas-service>/v2/haas-loader.js
- HaaS v2 resources are part of the hubspot-theme repository (not Antlers anymore)
- haas-bundle replaced by haas.css that is created by Tailwind and only includes needed styles
- missing styles (formerly provided by haas-bundle) included explicitely


dspublisher/theme/init.js
```diff
@@ -3,10 +3,22 @@ const fs = require(`fs`);
...
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/1-foundation/theme.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/2-components/button.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/2-components/dropdown.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/2-components/links.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/2-components/lists.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/2-components/tag.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/3-patterns/navigation.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/5-sections/addsearch.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/5-sections/cookie.css',
+    'https://cdn.vaadin.com/website/antlers/v2/assets/css/3-patterns/card.css'
```
jouni, line 21: I assume we don't have a "bundle" anymore, and all style sheets need to be loaded individually?

dspublisher/theme/styles/vcom/global.css
```diff
@@ -1,7 +1,21 @@
 /* These style sheets are fetched from vaadin.com during build (see gatsby-node.js) */
-@import 'https://cdn.vaadin.com/vaadin-design-system/latest/assets/icons/css/line-awesome.min.css';
-@import '../antlers/assets/css/haas-bundle.css';
+@import 'https://cdn.vaadin.com/website/antlers/v2/assets/icons/css/line-awesome.min.css';
+@import 'https://cdn.vaadin.com/website/hubspot-theme/v2/haas/css/haas.css';
 @import '../antlers/assets/css/3-patterns/card.css';
```
jouni, line 4: Duplicate line.

```suggestion
```

jouni (commented):
Looks good on the surface (changes), but the preview deployment doesn’t look as it should.
Screen Shot 2021-08-16 at 12 21 50

dspublisher/theme/init.js
paulroemer: Yes, no prebuilt bundle anymore.

paulroemer:
@jouni, the production HaaS is not updated, yet. I hoped you are using https://preview.vaadin.com/haas-service for your preview builds.

dspublisher/theme/styles/vcom/global.css
paulroemer: fixed in 8c2a6e982b4049505669def0ad818b3acc8894fa

jouni:
Right. I think we’ve had this discussion before, that we would like to use the production HaaS when previewing the docs deployment, so that we get to review the expected result what we will also get in production.

We can change it to use the HaaS preview for this PR for a moment, and then switch back once the new version of HaaS goes to production.

dspublisher/theme/header.ts
```diff
@@ -10,8 +10,8 @@ export default class Example extends LitElement {
     script.onload = () => (window as any).haas.loader.initMenu();
     script.src =
       window.location.hostname == 'preview.vaadin.com'
```
jouni, line 12:
```suggestion
      window.location.hostname == 'preview.vaadin.com' || window.location.hostname.endsWith('netlify.app')
```

jouni: @paulroemer, would this be enough?

paulroemer:
Well, it's pretty likely that I will push the HaaS updates to production in the next days. No need to change the behavior of your deployment.

dspublisher/theme/header.ts
paulroemer: Yeah, that will work of course. But let's not change the default behavior but make sure HaaSv2 is in production and works as expected.

paulroemer:
@jouni, HaaSv2 was deployed to production. You can check https://deploy-preview-695--vaadin-docs-preview.netlify.app/

Looks like there are still some styling issues?

jouni:
Okay, thanks! Yeah, I think most custom properties have been renamed, and none of the colors are applied correctly. I’ll work on updating the integration.

PR #718, opened 2021-08-30
chore: update Vaadin version to 22.0.0.alpha2

web-padawan:
## Description

Updated the Vaadin version and fixed one docs example to use the renamed `TextArea` element class.


ZheSun88 (approved):

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
```diff
@@ -7,7 +7,7 @@ import '@vaadin/vaadin-text-field/vaadin-text-area';
 
 import { applyTheme } from 'Frontend/generated/theme';
 import { RichTextEditorElement } from '@vaadin/vaadin-rich-text-editor/vaadin-rich-text-editor';
```
manolo, line 9: Out of curiosity, why `RichTextEditorElement` and  `TextArea` class names are not aligned, I mean  the `Element` suffix, or is it a wip?

web-padawan: This is one of the upcoming breaking changes in V22, see https://github.com/vaadin/web-components/issues/1993

PR #730, opened 2021-09-07
CE: Add MessageManager documentation

heruan:
Adds `MessageManager` documentation.

### DIFF
--- a/articles/ce/components/collaboration-message-list.asciidoc
+++ b/articles/ce/components/collaboration-message-list.asciidoc
@@ -87,6 +87,9 @@ through the provided [classname]#ActivationContext#.
 The callback should return another callback for clean-up when the topic connection is deactivated (topic id is changed).
 Note how the submit button is enabled only when there's an active topic connection.
 
+.Submit messages via [classname]#MessageManager#
+TIP: Messages can also be submitted to the list programmatically using a [classname]#MessageManager# instance with the same topic id, as described in the xref:{articles}/ce/managers/message-manager#ce.message-manager[Message Manager documentation].
+
 == Customizing Messages
 
 [classname]#CollaborationMessageList# enables changing the properties of [classname]#MessageListItem# items after they are generated.
--- a/articles/ce/components/collaboration-message-list.asciidoc
+++ b/articles/ce/components/collaboration-message-list.asciidoc
@@ -31,14 +31,14 @@ The [classname]#CollaborationMessageList# constructor takes two arguments: the i
 the end user associated with this session, and a topic id.
 The user info is used to render the user name and avatar in the messages submitted by this user.
 Refer to the <<collaboration-avatar-group#ce.configuring-avatars, `CollaborationAvatarGroup` documentation>>
-for more details on how the avatars are rendered, or how to load images from a backend with the [methodname]#setImageProvider# method.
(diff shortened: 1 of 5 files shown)


articles/ce/managers/message-manager.asciidoc
```diff
@@ -0,0 +1,98 @@
+---
+title: Message Manager
+order: 2
+layout: page
+---
+
+[[ce.message-manager]]
+= Manage User Presence
+
+Collaboration Engine includes a manager to handle text messages in a topic, for example subscribing to incoming messages and submit new ones.
```
Legioth, line 10: "programmatically" might be a good keyword here to make it immediately clear how this is different from `CollaborationMessageList`.

articles/ce/managers/message-manager.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+The manager can also be configured to persist messages to a custom backend (for example a database) so that messages are restored when the application restarts.
+It provides a simple and flexible way to manage messages in a topic and makes it easy to create custom components with collaborative messaging features.
```
Legioth, line 12: This sounds mostly like trying to sell something, but it doesn't really help the reader understand what it is. The main problem is empty words like "simple" and "easy".

articles/ce/managers/message-manager.asciidoc
```diff
@@ -0,0 +1,98 @@
+---
+title: Message Manager
+order: 2
+layout: page
+---
+
+[[ce.message-manager]]
```
Legioth, line 7: Might be good to somehow point out that `MessageManager` and `CollaborationMessageList` are compatible with each other if pointed at the same topic?

articles/ce/components/collaboration-message-list.asciidoc
```diff
@@ -106,41 +69,21 @@ You can set the message persister in the message list's constructor:
...
-stream should include the messages sent exactly at the time that you can get with [methodname]#getSince()# from the [classname]#FetchQuery#.
-This ensures that all messages are fetched even when multiple messages have the same timestamp.
-The component takes care of filtering out the duplicates, which are expected when returning the latest message again.
-
 == Using Custom Input Components
 
-Instead of using the provided `CollaborationMessageInput` component, you can also
-configure your own component to submit messages. The component can be registered with the `CollaborationMessageList::setSubmitter` method.
-Here's an example of submitting messages with `TextField` and `Button` components:
+Instead of using the provided [classname]#CollaborationMessageInput# component, you can also
```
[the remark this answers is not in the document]
heruan: Sure! I've added notes cross-referencing each other to suggest this possibility.

articles/ce/managers/message-manager.asciidoc
heruan: Added a note 👍

articles/ce/managers/message-manager.asciidoc
heruan: We want to sell something 😄 Jokes apart, since then we provide example on how to use it, I think it's fine here.

Legioth: Words like that will just make people frustrated in case they're having problems.

Instead, we could say something like this:
> It provides a flexible way to manage messages in a topic and helps create custom components with collaborative messaging features.

Legioth (approved):

magi42 (approved):

PR #756, opened 2021-09-20
Moved upgrading guide to its own section

magi42:
- [x] Refactor Flow Upgrading Guide
- [x] Refactor Fusion Upgrading Guide

### DIFF
--- a/articles/shared/upgrading/_common-steps.asciidoc
+++ b/articles/shared/upgrading/_common-steps.asciidoc
@@ -2,7 +2,7 @@
 
 These steps apply regardless of what version you are upgrading from.
 
-. Delete `node_modules` folder, `package.json`, and `pnpm-lock.yaml`
+. Delete the `node_modules` folder and either lock file: `package-lock.json` (with npm) or `pnpm-lock.yaml` (with pnpm).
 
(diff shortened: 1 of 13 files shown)


articles/shared/upgrading/_common-steps.asciidoc
```diff
@@ -0,0 +1,29 @@
+== Common Steps
+
+These steps apply regardless of what version you are upgrading from.
+
+. Delete `node_modules` folder, `package.json`, and `pnpm-lock.yaml`
```
pleku, line 5: Now that I see this, this is not the right thing to do. I think it should mention that you delete either `package-lock.json` (with npm) or `pnpm-lock.yaml` (with pnpm). `package.json` should not be deleted in case they have their own customizations.

pleku (changes requested):
The structure seems OK, but I noticed one bad instruction

articles/shared/upgrading/_common-steps.asciidoc
pleku:
```suggestion
. Delete the `node_modules` folder and either lock file: `package-lock.json` (with npm) or `pnpm-lock.yaml` (with pnpm).
```

PR #758, opened 2021-09-21
docs: add section for Notification.show helper

sissbruecker:
## Description

Adds new section for the Notification.show helper method, now that it is available in both Java + TS. Also updated the basic TS examples to use the helper.

I decided to introduce a new section for this. Although the helper is used in several other examples I though it would be easier to discover this way. Also was easier to explicitely hint at the support for Lit templates in the TS example.

Related to: https://github.com/vaadin/components-team-tasks/issues/593

### DIFF
--- a/articles/ds/components/notification/index.asciidoc
+++ b/articles/ds/components/notification/index.asciidoc
@@ -433,6 +433,24 @@ include::{root}/src/main/java/com/vaadin/demo/component/notification/Notificatio
 ----
 --
 
+== Static Helper
+
+For simple, one-off notifications, it is more convenient to use the static `show` helper.
+The helper manages the notification's lifecycle, and adds and removes it from the DOM automatically.
+
+[.example]
+--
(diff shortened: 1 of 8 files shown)


articles/ds/components/notification/index.asciidoc
```diff
@@ -441,6 +433,24 @@ include::{root}/src/main/java/com/vaadin/demo/component/notification/Notificatio
 ----
 --
 
+== Static Helper
+
+For simple, one-off notifications, it is more convenient to use the static `show` helper.
+The helper manages the notification's lifecycle, and adds and removes it from the DOM automatically.
```
magi42, line 439:
```suggestion
For simple, one-off notifications, it is convenient to use the static `show()` helper method.
The helper manages the notification's lifecycle, and adds and removes it from the DOM automatically.
```

yuriy-fix (approved):

PR #797, opened 2021-10-07
Update handling-view-state.adoc

sindrehauglum:
## Description

The file name is ListView.java

## Type of change

- [x ] Bugfix
- [ ] Feature


jouni:
Apologies for taking so long to merge this PR. The preview deployment was blocking it because we need to explicitly allow the build to run (in Netlify) for external contributors.

Thanks for the fix!

PR #825, opened 2021-10-19
docs: port button component docs to 14

sissbruecker:
## Description

Ports button component docs to v14.

- asciidoc: commented out several references to non-existing docs, added TODO
- asciidoc: commented out related components section, added TODO
- several visual differences regarding colors due to theme changes in V22


sissbruecker:
@DiegoCardoso 

> One question, though: how do we make sure that all the TODOs added on this and in other components examples are going to be addressed properly?

My plan was to port every component first, and then open a separate PR to restore the references. I might even continue commenting out **every** reference, even if it is already ported, to avoid cluttering this up too much. Assuming we can get this finished in around two weeks, that should be OK.

PR #874, opened 2021-11-01
docs: port notification component docs to 14

sissbruecker:
## Description

Ports notification docs to v14.

- asciidoc: removed static helper section, this was added specifically for the web component in 22, Java examples already use the static show method when appropriate
- demos: in V14 middle-positioned notifications have a default width of `20em`, which forces more line breaks when compared to V22


tomivirkki (changes requested):
The following examples don't work when opened in a new tab (may be missing some imports in the TS file):
- "Icons and Other Rich Formatting"
- "Position"

sissbruecker:
> The following examples don't work when opened in a new tab (may be missing some imports in the TS file):
> 
>     * "Icons and Other Rich Formatting"
> 
>     * "Position"

@tomivirkki Good catch! Fixed the missing imports.

tomivirkki (approved):

PR #879, opened 2021-11-02
CE: Mention V14 compatibility upto CE 3.2

heruan:
This adds a note to the Collaboration Engine overview page to mention that in V14 CE is compatible upto version 3.2.


articles/ce/overview.asciidoc
```diff
@@ -13,6 +13,10 @@ order: 1
 https://vaadin.com/collaboration[Collaboration Engine] is a solution for building real-time collaboration features into web apps with a few lines of code.
 This means that users can collaboratively edit and communicate in real time, directly within secure Vaadin web applications.
 
+.Adding Collaboration Engine to your project
```
jouni, line 16: I’d try to grab the readers attention a bit more with the title. As the point about this admonition is to highlight the compatibility with different Vaadin versions, maybe this:

```suggestion
.Compatibility with Vaadin versions
```

heruan: That's definitely better, thanks Jouni!

tulioag (approved):

PR #888, opened 2021-11-02
docs(upgrade): add a note note about TypeScript catch type change

platosha:
Fixes vaadin/fusion#108


frontend/demo/fusion/errorhandling/catch-error.ts
```diff
@@ -6,8 +6,8 @@ export async function callEndpoint() {
   try {
     await DataEndpoint.getViewData();
   } catch (error: any) {
-    console.warn(error.message); // "Not implemented"
     if (error instanceof EndpointError) {
+      console.warn(error.message); // "Not implemented"
```
Haprog, line 10: Was this line moved intentionally? We can now also remove the `any` from the catch line right?

Haprog (changes requested):

articles/shared/upgrading/_changes.asciidoc
```diff
@@ -28,6 +28,35 @@ to:
 import { EndpointError } from '@vaadin/fusion-frontend';
 ----
 
+=== TypeScript 4.4 Default Catch Variable Type
+
+Vaadin has changed the TypeScript dependency version to 4.4.
+Starting from this version, the default catch variable type changes in TypeScript from `any` to `unknown`.
+In practice, the error handling code that did not involve `instanceof` guards needs changes.
```
dougjewsbury, line 35:
```suggestion
As a result, there need to be changes to the error handling code that did not involve `instanceof` guards.
```

frontend/demo/fusion/errorhandling/catch-error.ts
platosha: Yes, intentional, as `instanceof` type guard . Let me remove it.

Haprog (approved):

jouni:
@Haprog, when you update the labels, remember to remove the corresponding "not OK" label. For example, when you add `TypeScript OK` label, remember to remove the `TypeScript` label.

Haprog:
> @Haprog, when you update the labels, remember to remove the corresponding "not OK" label. For example, when you add `TypeScript OK` label, remember to remove the `TypeScript` label.

Ok, thanks. I thought the other label means "requires TS review" so it didn't make sense to me to remove it. I can follow your advice for the future.

jouni:
FYI, note that this same article is currently used for both the Flow and Fusion upgrading guides. Changes only relevant for Fusion should be shown only inside the `ifdef::fusion[]` ... `endif::fusion[]` tags, and similarly with Flow `ifdef::flow[]`.

I noticed this today, and updated it so that the section added by this PR is only shown in the Fusion upgrading guide.

PR #892, opened 2021-11-04
fix: update virtual list TS example

### LINKED ISSUE #891: Bug in the Virtual List TypeScript example
There is a bug in the Virtual List component TypeScript example at https://vaadin.com/docs/latest/ds/components/virtual-list

Expanding the details of an item also causes every Nth item appear with details expanded as you scroll down the virtual list.

https://user-images.githubusercontent.com/1222264/140095804-f917fc8e-3638-4bd4-ae27-486b8dee6f66.mp4

This is somehow related to using Lit in the item renderer; the example works as expected if you add the following to the beginning of `personCardRenderer`:
```ts
delete (root as any)._$litPart$;
root.textContent = '';
```
This isn't the proper fix however since it causes the item elements getting cleared and completely rebuilt on every render.

The example works in the current V21 documentation https://vaadin.com/docs/v21/ds/components/virtual-list most likely because it uses an older version of the virtualizer which made excess renderer calls whereas the new version only calls the renderer once/index.

tomivirkki:
Fixes #891

### DIFF
--- a/frontend/demo/component/virtuallist/virtual-list-basic.ts
+++ b/frontend/demo/component/virtuallist/virtual-list-basic.ts
@@ -34,7 +34,6 @@ export class Example extends LitElement {
   @state()
   private people?: Person[];
 
-  @state()
   private expandedPeople: Set<Person> = new Set();
 
   async firstUpdated() {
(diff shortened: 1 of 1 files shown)


web-padawan (approved):
LGTM. I will also add virtual list to my `lit-vaadin-helpers` project and take this into account.

PR #894, opened 2021-11-04
chore: use live directive in virtual list example

tomivirkki:
Use live-directive as suggested by @web-padawan 

Follow-up from https://github.com/vaadin/docs/pull/892


web-padawan (approved):
LGTM. I think we could document `live` separately but let's not do it in scope of this PR.

PR #896, opened 2021-11-04
Simplify Fusion starting-a-project instructions

tarekoraby:
Fusion's starting-a-project instructions include an inaccurate flow-chart (one, for example, that mentions a non-existent Maven archetype for Fusion). The instructions also imply the existence of different stacks for Fusion, which is not accurate since only Spring Boot is supported.


articles/fusion/guide/start/index.asciidoc
```diff
@@ -6,10 +6,6 @@ order: 20
 
 = Starting a Project
 :wizard-name: Vaadin Start
-:toclevels: 2
-
-[.cards.quiet.large.hide-title]
-== Vaadin Start
 
 [.breakout]
 --
```
jouni, line 11: No need to use these open blocks. The `.breakout` block is only useful when inside a `.cards` section.

Remember to remove the ending `--` as well (github didn't allow me to make the suggestion there).
```suggestion
```

articles/fusion/guide/start/index.asciidoc
```diff
@@ -19,16 +15,7 @@ It allows you to configure and download a customized starting point for your app
 
 https://start.vaadin.com?preset=fusion[Open {wizard-name}, role="button primary water"]
 
-
-== Next Steps
-
+[.breakout]
+--
```
jouni, line 19:
```suggestion
```

articles/fusion/guide/start/index.asciidoc
```diff
@@ -19,16 +15,7 @@ It allows you to configure and download a customized starting point for your app
...
-
-.Ways to create and run a new project
-[.subtle]
-image::images/newproject-flowchart.svg[New project flowchart, width=100%]
-
-If you create a project outside an IDE, you need to import it in the IDE.
-During development, you can run the project with a Maven target, depending on the technology stack you are using.
-
-After you have the project in your IDE, you can go ahead and run or debug it.
+--
```
jouni, line 21:
```suggestion
```

articles/fusion/guide/start/index.asciidoc
tarekoraby: @jouni without the `.breakout`, this paragraph was placed on the same horizontal line as the button. It should be below it.

jouni: That was because of the `.cards` class that was previously used for the section these elements are in.

jouni (approved):

PR #900, opened 2021-11-08
docs: port radio button component docs to 14

sissbruecker:
## Description

Ports radio button docs to v14.

- asciidoc: commented out references to non-existing documentation, added TODO


articles/ds/components/radio-button/index.asciidoc
```diff
@@ -0,0 +1,216 @@
...
+////
+:component-name: Radio Button Group
+include::../_shared.asciidoc[tag=field-features]
+////
+
+== States
+
+=== Read-Only
+
+Use read-only when content needs to be accessible but not editable. Read-only elements cannot be edited, but they do participate in the tabbing order and can thus receive focus. The contents of a read-only input can be selected and copied.
```
sosa-vaadin, line 39: Write one sentence per line.
```suggestion
Use read-only when content needs to be accessible but not editable.
Read-only elements cannot be edited, but they do participate in the tabbing order and can thus receive focus.
The contents of a read-only input can be selected and copied.
```

articles/ds/components/radio-button/index.asciidoc
```diff
@@ -0,0 +1,216 @@
...
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/radiobutton/RadioButtonReadonly.java[render,tags=snippet,indent=0]
+----
+--
+
+=== Disabled
+
+Disable a field to mark it as currently unavailable. Disabled state is used for fields that are not editable and do not need to be readable. Disabled elements cannot be focused and may be inaccessible to assistive technologies like screen readers.
```
sosa-vaadin, line 57: Write one sentence per line.
```suggestion
Disable a field to mark it as currently unavailable.
Disabled state is used for fields that are not editable and do not need to be readable.
Disabled elements cannot be focused and may be inaccessible to assistive technologies like screen readers.
```

articles/ds/components/radio-button/index.asciidoc
```diff
@@ -0,0 +1,216 @@
...
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/radiobutton/RadioButtonGroupLabels.java[render,tags=snippet,indent=0]
+----
+--
+
+=== Custom Option
+
+To enable the user to enter a custom option instead of picking one from the list, use an “Other” radio button at the bottom of the list with an associated Text Field for entry. The field should be hidden or disabled until the “Other” option is selected.
```
sosa-vaadin, line 151: Write one sentence per line.
```suggestion
To enable the user to enter a custom option instead of picking one from the list, use an “Other” radio button at the bottom of the list with an associated Text Field for entry.
The field should be hidden or disabled until the “Other” option is selected.
```

articles/ds/components/radio-button/index.asciidoc
```diff
@@ -0,0 +1,216 @@
...
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/radiobutton/RadioButtonCustomOption.java[render,tags=snippet,indent=0]
+----
+--
+
+=== Default Value and Blank Option
+
+It is generally recommended to set the most common option as the default value for Radio Button Groups. Place the default option at the top of the list.
```
sosa-vaadin, line 169: Write one sentence per line.
```suggestion
It is generally recommended to set the most common option as the default value for Radio Button Groups.
Place the default option at the top of the list.
```

articles/ds/components/radio-button/index.asciidoc
```diff
@@ -0,0 +1,216 @@
...
+include::{root}/frontend/demo/component/radiobutton/radio-button-default-value.ts[render]
+----
+
+=== As Alternative to a Checkbox
+
+Two Radio Buttons can sometimes be a better alternative to a single Checkbox.
+
+If the Checkbox does not represent a simple yes/no choice, and its label cannot clearly communicate the meaning of its unchecked state, it is better to use a Radio Button Group with two options:
+
+// TODO split this into two examples, do and dont
```
sosa-vaadin, line 186: Any reason why not do it in this PR?

sosa-vaadin (changes requested):
There's only a few lines with multiple sentences which should be split into one line per sentence. Other than that, it looks good.

sissbruecker:
@sosa-vaadin In general I would prefer to keep the content untouched for the initial back-porting. In order to keep the docs in sync, it makes more sense to tackle this in `latest` first, and then back-port it.

sosa-vaadin (approved):

PR #901, opened 2021-11-08
fix: use .js extension for Lit decorator/directive imports

jouni:
Based on https://github.com/lit/lit/issues/1986, the `.js` is required by Lit.


jouni:
This should be picked to v21 as well. Not sure about v14.

web-padawan (approved):
LGTM, thanks!

PR #907, opened 2021-11-08
fix: add missing imports for select demos

sissbruecker:
## Description

Adds hidden Typescript module for importing web components, for the select empty selection demos. Fixes the demos not working when opening them in a new tab.


jouni:
One of the source files seems to have a space instead of a dash (`select-empty-selection caption.ts`): 
https://github.com/vaadin/docs/blob/b2c7e4cf3990eded7725efcae619b6d2d0e422bc/frontend/demo/component/select/select-empty-selection%20caption.ts

sissbruecker:
@jouni Thanks, fixed it.

tomivirkki (approved):
LGTM

PR #909, opened 2021-11-09
docs: port split layout component docs to 14

sissbruecker:
## Description

Ports split layout docs to v14.

- none


hdamr (approved):
I ran the docs locally and the examples worked fine with no issue.

PR #914, opened 2021-11-10
docs: port time picker component docs to 14

sissbruecker:
## Description

Ports time picker docs to v14.

- none


hdamr (approved):
All the examples were rendered correctly, the links also pointed to correct place.

PR #938, opened 2021-11-17
Vale config

jouni:
Add a separate Vale config for GitHub pull request reviews, which limits the checks to error level only.

When editing/authoring locally, you will get warnings and suggestions as well.

At the same time, copy all relevant style rules (from Microsoft, Google, and write-good style guides) over to the Vaadin styles, taking ownership of all the styles that we enforce.

The motivation is to reduce the noise in pull request reviews, because we have so many pre-existing issues in our documentation, that they usually distract from the reviews. This puts more responsibility on the authors, rather than reviewers to keep the content clean.

---

The change to [_maven.asciidoc](https://github.com/vaadin/docs/pull/938/files#diff-a5f2f8bfd1ddd1ce82b7dae21e772688ca57be2a362ad5f96bd4c7797d4ebcf9) is only made so that the linting check is triggered for that file, and we can see how this works in practice.


.github/workflows/lint.yml
```diff
@@ -7,11 +7,7 @@ jobs:
       - uses: actions/checkout@master
       - uses: errata-ai/vale-action@master
         with:
-          styles: |
-            https://github.com/errata-ai/Microsoft/releases/latest/download/Microsoft.zip
-            https://github.com/errata-ai/Google/releases/latest/download/Google.zip
-            https://github.com/errata-ai/write-good/releases/latest/download/write-good.zip
-          config: https://raw.githubusercontent.com/vaadin/docs/latest/.vale.ini
+          config: https://raw.githubusercontent.com/vaadin/docs/vale-config/.github/.vale-pr.ini
```
jouni, line 10: Update before merging:

```suggestion
          config: https://raw.githubusercontent.com/vaadin/docs/latest/.github/.vale-pr.ini
```

.github/.vale-pr.ini
```diff
@@ -0,0 +1,29 @@
...
+
+# Ignored HTML elements
+IgnoredScopes = code, tt, kbd
+
+# Rules for asciidoc files
+[*.{asciidoc,adoc}]
+
+BasedOnStyles = Vaadin, Vale
+
+MinAlertLevel = error
```
jouni, line 26: Hmm, odd. This doesn’t seem to have the desired effect.

jouni: FIxed

.github/.vale-pr.ini
```diff
@@ -0,0 +1,29 @@
...
+
+Vocab = Docs
+
+# Ignored CSS class names
+IgnoredClasses = filename, classname, methodname, interfacename, propertyname, guibutton, guilabel, custom-property
+
+# Ignored HTML elements
+IgnoredScopes = code, tt, kbd
+
+MinAlertLevel = error
```
jouni, line 21: Huh, still not working as expected here in GitHub. Works locally just fine.

jouni: Ah, here’s the reason (https://github.com/errata-ai/vale-action#config):

Screen Shot 2021-11-18 at 11 32 47

jouni: Seems like we need to explicitly disable all non-error rules manually.

.github/workflows/vale.yml
```diff
@@ -0,0 +1,14 @@
+name: lint
+on: [pull_request]
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: errata-ai/vale-action@master
+        with:
+          config: https://raw.githubusercontent.com/vaadin/docs/vale-config/.github/.vale-pr.ini
```
jouni, line 10:
```suggestion
          config: https://raw.githubusercontent.com/vaadin/docs/latest/.github/.vale-pr.ini
```

PR #945, opened 2021-11-23
fix: Fixes outdated package of VaadinServlet

mstahv:


anssit (approved):

jouni:
Remember to update the content labels when you do reviews 🤗

mstahv:
@jouni This should have been #bug 🤔 To tell you the truth, the contribution docs are so long that I didn't go through it for this trivial fix 😊

jouni:
Sure, the review labels are not the burden or contributors, but reviewers. If you are reviewing docs contributions, then you might want to read the instructions. But for simple stuff like this, not really necessary. But the review labels might prevent merging, and then you need to know what to do 🙂

PR #946, opened 2021-11-23
fix: fix links to Figma documents

sissbruecker:
## Description

Fixes links for Vaadin Charts + Vaadin Icons Figma documents, which were reversed.


jouni (approved):

jouni:
No idea why the labelling action decided to stamp this with "TypeScript". Need to see if that keeps happening more often.

PR #947, opened 2021-11-23
docs: port Figma docs to 14

sissbruecker:
## Description

- adds Figma page
- adds Material Theme demo link
- removes WIP hints for Design System from root index page

### DIFF
--- a/articles/ds/figma.asciidoc
+++ b/articles/ds/figma.asciidoc
@@ -1,0 +1,59 @@
+---
+title: Figma Libraries
+order: 150
+layout: index
+---
+
+= Figma Libraries
+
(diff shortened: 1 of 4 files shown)


jouni (approved):

jouni:
Should really figure out why Vale keeps reporting errors in all the wrong places.

jouni:
FYI, the reason for the incorrect Vale error annotations is because of the default `alt` texts Asciidoctor generates for images. It uses the image path the alt text (with some substitutions like dashes to spaces), and then Vale checks those strings, and incorrectly reports the error on the next occurrence of that string in the AsciiDoc source.

As a workaround, always provide an explicit alt text for images (for example `image::foobar[""]` for an empty string).

PR #950, opened 2021-11-24
Clarify and simplify Flow Get Started section, with a view to enhancing the onboarding experience for new users.

dougjewsbury:
Amendments to clarify and simplify Flow Get Started section, with a view to enhancing the onboarding experience for new users.


articles/flow/application/overview.asciidoc
```diff
@@ -22,41 +24,45 @@ The basic aspects of a Flow application are as follows:
...
-Applications often have a main view with a menu bar to show a selection of sub-views.
+Applications often have a main view with a menu bar that shows a selection of sub-views.
 When the user clicks on a menu item, the sub-view is displayed.
-The [classname]#AppLayout# component allows creating such views, supporting navigation and responsive layouts.
+With the [classname]#AppLayout# component, you can create such views, which support navigation and responsive layouts.
 
 <<resources#, Images and Icons>>::
 Images and icons are basic visual features of an application.
-Vaadin Design System comes with <<../../ds/foundation/icons#, two sets of icons>> that you can use in your application.
+The Vaadin Design System comes with <<../../ds/foundation/icons#, two sets of icons>> that you can use in your application.
```
jouni, line 50: I’m not entirely sure about the rules when "the" should be used. In this case, "Vaadin Design System" could be considered as a product name, so it is capitalized. And in those cases my assumption has been that "the" is not used.

articles/flow/application/overview.asciidoc
```diff
@@ -22,41 +24,45 @@ The basic aspects of a Flow application are as follows:
...
 
 <<styling#, Styling>>::
-To define the basic style of applications, Vaadin provides a default Lumo theme, and an alternative Material theme.
-You can customize styling with CSS style sheets.
+To define the basic style of applications, Vaadin provides a default _Lumo_ theme, and an alternative _Material_ theme.
```
jouni, line 53: Did we have a guideline about using italics when emphasizing names? I’m curious about the reason why you chose to use it here. I do think it makes sense, but I’m not sure I would know when to use it myself.

articles/flow/application/overview.asciidoc
```diff
@@ -86,11 +92,11 @@ public class GreetService implements Serializable {
...
 Context Dependency Injection (CDI) and Java EE::
 The Java Enterprise Edition (EE) includes many features for creating enterprise applications.
-CDI or _context dependency injection_ is the Java EE way to manage service objects and inject them into applications, such as Vaadin UIs.
-CDI requires a Java EE -enabled web container; the starter projects use Apache TomEE.
+CDI, or _context dependency injection_, is the Java EE way to manage service objects and inject them into applications, such as Vaadin UIs.
+CDI requires a Java EE-enabled web container; the starter projects use Apache TomEE.
 +
 The starter project includes an example service that handles business data and logic.
-It is injected in the main view, and can be injected in similar way to other views or elsewhere.
+It is injected into the main view, and can be injected in a similar way into other views or elsewhere.
```
jouni, line 99:
```suggestion
The service injected into the main view, and can be injected in a similar way into other views or elsewhere.
```

articles/flow/application/ui.asciidoc
```diff
@@ -57,16 +62,20 @@ include::{root}/frontend/demo/flow/application/ui/ui-menu-basic.ts[hidden]
 
 The [classname]#AppLayout# component gives a basic layout for the main view of an application.
```
jouni, line 63: Prefer using the component name rather than the Java class name, unless talking very explicitly about the Java API and how it works.

```suggestion
The App Layout component gives a basic layout for the main view of an application.
```

articles/flow/application/ui.asciidoc
```diff
@@ -57,16 +62,20 @@ include::{root}/frontend/demo/flow/application/ui/ui-menu-basic.ts[hidden]
...
 It has a menu area, as well as a content area controlled by the menu.
-It fully works with routing to provide navigation between content views.
+It provides full navigation functionality between content views using routing.
 
-The starter applications created with the https://start.vaadin.com[Vaadin Start] by default use the [classname]#AppLayout# to handle the main view.
+By default, starter applications created with https://start.vaadin.com[Vaadin Start] use [classname]#AppLayout# to handle the main view.
```
jouni, line 67:
```suggestion
By default, starter applications created with https://start.vaadin.com[Vaadin Start] use App Layout to handle the main view.
```

jouni (approved):
Very nice, thanks! Just a couple of small remarks, that I would want to clarify.

articles/flow/application/overview.asciidoc
dougjewsbury: I made this change rather casually, it's true. There's no guideline on this currently. The wording "a default Lumo theme" sounds a little strange, as if some other kinds of Lumo theme existed; similarly with "an alternative Material theme". Would it be appropriate to say "... the Lumo theme by default, with the Material theme as an alternative."? We could remove the italics, since there's no basis in the guidelines for using it.

articles/flow/application/overview.asciidoc
dougjewsbury: I would say the reason that "the" is used is because the phrase "Vaadin Design System" contains the common noun "system". In the same way, we talk about THE Java Development Kit, even though it, too, is a product name and, hence, a proper noun (as a phrase).

articles/flow/components/overview.asciidoc
```diff
@@ -6,12 +6,13 @@ layout: page
...
-<<{articles}/ds/overview#, Vaadin Design System>> includes a set of <<{articles}/ds/components#,components>> that you can use to build your UI.
+The <<{articles}/ds/overview#, Vaadin Design System>> includes a set of <<{articles}/ds/components#,components>> that you can use to build your UI.
 The components have a server-side Java API in addition to the TypeScript API for client-side development.
 
-You use the components by creating one and then adding it to a containing layout.
+You use a component by first creating it and then adding it to a containing layout.
 
-For example, to add a <<{articles}/ds/components/button#, Button component>>, create the object and add it to a layout (here the calling object extends a layout):
+For example, to add a <<{articles}/ds/components/button#, Button component>>, create the object and add it to a layout (here, the calling object extends a layout):
+// where does the calling object extend a layout?
```
jouni, line 15: This looks like a leftover comment. In case you are interested in the answer, it’s here: https://github.com/vaadin/docs/blob/latest/src/main/java/com/vaadin/demo/flow/application/events/EventsLambda.java#L9

Though, it actually extends a Div, which is not really a "layout" per se. But I believe what we try to say here, is that the calling object has the `add()` method.
```suggestion
```

articles/flow/components/overview.asciidoc
```diff
@@ -133,10 +133,10 @@ You can find additional prebuilt Web Components (with Java APIs) contributed by
 As an alternative to using the platform dependency, you can declare single components as dependencies.
 
 You should add both the Vaadin `bom` and the relevant Flow component package, for example `vaadin-button-flow`, to your project dependencies.
-The Vaadin `bom` fixes all vaadin-related dependencies to a tested combination, so that the individual components can be added safely.
-Without the BOM, some dependencies may change in the future, because of new releases, or because of their use of version ranges.
+The Vaadin `bom` binds all vaadin-related dependencies to a tested combination, so that the individual components can be added safely.
```
jouni, line 136:
```suggestion
The Vaadin `bom` binds all Vaadin-related dependencies to a tested combination, so that the individual components can be added safely.
```

jouni, line 136:
```suggestion
The Vaadin `bom` binds all Vaadin-related dependencies to a tested combination, so that the individual components can be added safely.
```

.github/styles/Vocab/Docs/accept.txt
```diff
@@ -44,6 +45,7 @@ LitElement
 Lumo
 macOS
 Maven
+(mM)emoiz(e|ed|es)
```
dougjewsbury, line 48:
```suggestion
[mM]emoiz(e|ed|es)
```

jouni (approved):

PR #980, opened 2021-12-12
chore: Upgrade Spring to 2.6.1

tarekoraby:


jouni:
I assume you tested this locally? The Netlify preview build doesn't run the Java server, so that needs to be verified locally.

tarekoraby:
`mvn clean install` runs fine locally.

tomivirkki:
> `mvn clean install` runs fine locally.

What should be tested is actually starting up the docs project locally (`mvn`) and then testing (with the docs viewer app) that the embedded Java examples etc work.

tarekoraby:
The project actually doesn't start with Spring Boot `2.6.1`. I believe there is a fix for that coming soon with Vaadin `22.0.2`. Let's wait for that.

tarekoraby:
Tested locally with Vaadin 22.0.2 + Spring 2.6.1, and everything seems to be running fine.

tomivirkki (approved):
Confirmed this works now and actually fixes the issue with the server currently not starting at all

PR #983, opened 2021-12-13
Add example for custom-field with whitespace

hdamr:
This PR adds a bit of explanation about using `custom-field` with components that don't have outer margin/padding (like RTE).


src/main/java/com/vaadin/demo/component/customfield/CoordinateField.java
```diff
@@ -0,0 +1,42 @@
...
+  private Input longitude;
+  private Input latitude;
+  private HorizontalLayout layout;
+
+  public CoordinateField() {
+    longitude = new Input();
+    longitude.setPlaceholder("Longitude");
+
+    latitude = new Input();
+    latitude.setPlaceholder("Latitude");
```
jouni, line 19: Missing `aria-label`s.

sissbruecker (approved):
Looks good. The PR should not be merged until the corresponding feature is merged. Marking the PR as draft until then.

Also we need to cherry-pick this for all targeted versions.

src/main/java/com/vaadin/demo/component/customfield/PaymentInformationField.java
```diff
@@ -27,6 +27,9 @@ public PaymentInformationField() {
     // Adds small amount of space between the components
     layout.getThemeList().add("spacing-s");
 
+    // Increases padding of field's label
+    getElement().getThemeList().add("whitespace");
```
jouni, line 31: Shouldn't there be a constant/enum for this variant?

hdamr: That can be discussed, but since we don't have constants for `spacing-s` and ... I thought we don't need one for this particular case either.

jouni: Okay. Just thought I’d mention, as this has been said to be a nuisance, that variants aren't available as enums. But not blocking this PR of course.

jouni:
@yuriy-fix, is this PR going somewhere soon?

PR #986, opened 2021-12-15
Update bundle loading sequence

TatuLund:
gwt-elemental is now needed to be activated separately as it is its own bundle and not static part of vaadin-shared, see

https://github.com/vaadin/framework/pull/12416


jouni:
I don't really know anything about this topic. I’d rather have someone familiar with the topic to review and accept this.

jouni:
Apparently no one else is going to review this, so I’ll merge this now.

PR #987, opened 2021-12-16
Update index heading to indicate the version number

jouni:


jouni:
This PR was only for testing the Amplitude service.

PR #990, opened 2021-12-17
Delete noisy AWS Amplify comments

tarekoraby:
Workaround copied from: https://github.com/aws-amplify/amplify-console/issues/2293#issuecomment-990709903

### DIFF
--- a/.github/workflows/delete-comments-from-aws-amplify.yml
+++ b/.github/workflows/delete-comments-from-aws-amplify.yml
@@ -1,0 +1,14 @@
+# .github/workflows/delete-comments-from-aws-amplify.yml
+name: Delete noisy AWS Amplify comments
+on:
+  issue_comment:
+    types: [created]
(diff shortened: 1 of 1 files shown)


jouni:
Seems to work as expected, right? But how do we get the preview URL then, if it’s not posted as a comment? I can't log in to the AWS console to check it.

tarekoraby:
Ah well, I didn't notice that we cannot access the preview in the same way as Amplify. I guess we could keep only one of the comments by the bot. But I don't know how to do that (if it's at all possible). Do we just close this PR without merging?

tarekoraby:
Maybe if there is a way to hide (i.e. minimize) the comments instead of deleting them, then that would work to minimize the noise.

tarekoraby:
I'm closing this PR for now as I couldn't easily discover a way to hide those noisy comments.

PR #991, opened 2021-12-17
fix: adjust chart docs to changes introduced with new theme feature

sosa-vaadin:
With the addition of Lumo and Material themes for charts, styles can now be handled within any chart separately instead of across the entire project. We need to update the wording in our docs to reflect this change.


jouni (approved):

jouni:
We should probably wait until the feature is released before we merge this.

sosa-vaadin:
A few more changes have to be made. Any mention of adding or including a default theme has to be removed since it is no longer necessary.Also, a new note about how to make sure general CSS rules only apply to charts with `styledMode` enabled.

articles/ds/components/charts/java-api/css-styling.asciidoc
```diff
@@ -59,12 +55,13 @@ custom-chart-styles.css
 }
 ----
 
+NOTE: If you want your CSS rules to only apply to charts with `styledMode` enabled, add an attribute selector like `:host(.first-chart)[styledMode=true]`.
```
jouni, line 58: I would assume that the `styledMode` attribute is not present when it’s false.

```suggestion
NOTE: If you want your CSS rules to only apply to charts with `styledMode` enabled, add an attribute selector like `:host(.first-chart)[styledMode]`.
```

PR #1024, opened 2022-01-13
Add Azure and AWS deployment instructions

tarekoraby:


jouni:
I’ll do some minor edits locally to this (so that I can see all Vale checks).

jouni:
> Shouldn’t all Vale issues be visible here in GH?

No, [most of them are actually disabled](https://github.com/vaadin/docs/blob/latest/.github/.vale-pr.ini) to avoid intimidating most contributors. We consider the majority of the checks to be useful only to the very eager authors.

PR #1050, opened 2022-02-01
add classic form-layout docs

sosa-vaadin:
Add mention of the classic component version of FormLayout.

### DIFF
--- a/articles/flow/upgrading/classic-components.asciidoc
+++ b/articles/flow/upgrading/classic-components.asciidoc
@@ -95,6 +95,19 @@ Related to the sizes, `VerticalLayout` and `HorizontalLayout` use Lumo variables
 
 |===
 
+
+=== FormLayout
+
+The classic `FormLayout` resembles, as close as possible, the Vaadin 7 or 8 version of `FormLayout`.
+It shows all components added to it in a table with two columns.
+The left column contains a component's caption (except for Checkbox).
+The components themselves appear stacked in the right column.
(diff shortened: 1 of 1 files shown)


articles/flow/upgrading/classic-components.asciidoc
```diff
@@ -95,6 +95,19 @@ Related to the sizes, `VerticalLayout` and `HorizontalLayout` use Lumo variables
 
 |===
 
+
+=== FormLayout
+
+The classic `FormLayout` resembles, as close as possible, the Vaadin 7 or 8 version of `FormLayout`.
```
dougjewsbury, line 101:
```suggestion
The classic `FormLayout` resembles, as closely as possible, the Vaadin 7 or 8 version of `FormLayout`.
```

dougjewsbury (commented):
Looks OK in terms of AsciiDoc.

jouni:
Note, that the "Flow / Upgrading" section has been removed since this branch was started. Upgrading instructions are now on the top level.  That said, I wonder if the Classic Components should be documented there, or some other place under Flow? @tarekoraby, do you have any thoughts about that?

Haprog (approved):

articles/flow/upgrading/classic-components.asciidoc
jouni: @sosa-vaadin, please implement this suggestion.

PR #1064, opened 2022-02-09
docs: add map sources and markers sections

sissbruecker:
## Description

Adds sections for map sources, markers, as well as the respective demos.


articles/ds/components/map/index.asciidoc
```diff
@@ -76,6 +76,99 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapViewport.java[ren
...
+Map provides various sources for loading map data in a specific format from a map service.
+Note that each source requires using a specific type of layer that can render data from that source.
+
+The following table lists the sources that can be used with Map, an example usage of each source can be found in the demo below:
+
+[cols="1,1,4"]
+|===
+| Source | Requires | Description
+
+| https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/component/map/configuration/source/OSMSource.html[OSM]
```
sissbruecker, line 94: 23.0.0.beta1 API docs are not published yet, so these links are broken for now

src/main/java/com/vaadin/demo/component/map/MapMarkers.java
```diff
@@ -0,0 +1,55 @@
...
+import com.vaadin.flow.component.map.configuration.Coordinate;
+import com.vaadin.flow.component.map.configuration.feature.MarkerFeature;
+import com.vaadin.flow.component.map.configuration.style.Icon;
+import com.vaadin.flow.server.StreamResource;
+
+@Route("map-markers")
+public class MapMarkers extends Div {
+
+    public MapMarkers() {
+        Span title = new Span("Map showing Vaadin office locations");
```
sissbruecker, line 17: Not very creative, let me know if you have better ideas 🤷

src/main/java/com/vaadin/demo/component/map/MapSources.java
```diff
@@ -0,0 +1,106 @@
...
+import com.vaadin.flow.component.radiobutton.RadioGroupVariant;
+
+import java.util.HashMap;
+import java.util.List;
+
+@Route("map-sources")
+public class MapSources extends VerticalLayout {
+    private static final String OSM_SOURCE = "OpenStreetMap";
+    private static final String XYZ_SOURCE = "XYZ (using satellite map via mapbox.com)";
+    private static final String WMS_SOURCE = "WMS (using topographical map via mundialis.de)";
```
sissbruecker, line 21: Could not find a well-known service that would host a full map (rather than overlay layers) with a WMS. Seems most use XYZ or vector tiles.

articles/ds/components/map/index.asciidoc
```diff
@@ -76,6 +76,99 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapViewport.java[ren
 ----
 --
 
+== Map Sources
+
+Web maps can be roughly categorized as being based either on tiled images, full images or vector data.
+Each category comes with its own set of standards and formats that define how the map data is structured and to be used.
+Map services are applications (typically servers or cloud services) that provide map data in one or more of these formats.
+
+Map provides various sources for loading map data in a specific format from a map service.
```
rolfsmeds, line 85: Should we be using the phrase _"The map component"_ rather than just _"Map"_ when referring to the component, to distinguish it from other uses in this same doc? (not just in reference to this particular line but other places where _"Map"_ is used to refer to the component)

rolfsmeds: Perhaps this is less of an issue in the middle of sentences where the capitalization might be sufficient to avoid misunderstandings, but in the beginning of sentences we don't have that distinction either.

articles/ds/components/map/index.asciidoc
```diff
@@ -76,6 +76,99 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapViewport.java[ren
...
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/map/MapSources.java[render,tags=snippet,indent=0,group=Java]
+----
+--
+
+== Markers
+
+Markers can be displayed on top of a map to indicate something of note.
```
rolfsmeds, line 142: would suggest the more common _"points of interest"_ rather than _"something of note"_

rolfsmeds (changes requested):
Very good overall, but see my comments for improvement suggestions.

articles/ds/components/map/index.asciidoc
sissbruecker: Makes sense, I also stumbled over this section. Changed this instance. Checked other places, I think it's not as problematic there, and it's consistent with other component docs.

articles/ds/components/map/index.asciidoc
sissbruecker: Sounds good, changed

articles/ds/components/map/index.asciidoc
```diff
@@ -76,6 +76,99 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapViewport.java[ren
...
+include::{root}/frontend/demo/component/map/map-markers.ts[preimport,hidden]
+----
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/map/MapMarkers.java[render,tags=snippet,indent=0,group=Java]
+----
+--
+
+.Performance Considerations
```
jouni, line 163: Heading formatting doesn’t apply to admonition titles.

```suggestion
.Performance considerations
```

articles/ds/components/map/index.asciidoc
```diff
@@ -76,6 +76,99 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapViewport.java[ren
...
+include::{root}/src/main/java/com/vaadin/demo/component/map/MapSources.java[render,tags=snippet,indent=0,group=Java]
+----
+--
+
+== Markers
+
+Markers can be displayed on top of a map to indicate something of note.
+That can be a point of interest, a vehicle, a building, or any other entity.
+A marker is defined by a coordinate and an icon.
+Icons can be configured either as a URL or a xref:{articles}/flow/advanced/dynamic-content#using-streamresource[StreamResource].
```
jouni, line 144: Even though both work, let’s keep the format of cross references consistent.

```suggestion
Icons can be configured either as a URL or a <<{articles}/flow/advanced/dynamic-content#using-streamresource,StreamResource>>.
```

DiegoCardoso (approved):
LGTM

src/main/java/com/vaadin/demo/component/map/Icons.java
```diff
@@ -0,0 +1,25 @@
+package com.vaadin.demo.component.map;
+
+import com.vaadin.flow.component.map.Assets;
+import com.vaadin.flow.component.map.configuration.style.Icon;
+
+public class Icons {
+    // DS publisher can not resolve default marker image, nor is there a way to host static images with a deterministic URL
+    // For now duplicate all required icons with base64 encoded URL
```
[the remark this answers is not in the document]
sissbruecker: Considering that we would need an alternative for the default marker anyway, I think it's good enough as it is. It's a super isolated workaround, and it's not hard to comprehend.

PR #1069, opened 2022-02-10
English-language improvements to Flow Configuration and Production documentation

dougjewsbury:
English-language improvements to Flow Configuration and Production documentation


.github/styles/Vocab/Docs/accept.txt
```diff
@@ -33,14 +33,16 @@ Gradle
...
+[hH]otpatch(ed|ing)?
+[hH]otswap(ped|ing)?
 Hilla
 i18n
-iterable
+[iI]terable
 Jandex
 Javadoc
 Karaf
+kbd
```
jouni, line 45: Instead of adding this to the vocabulary, try adding `:experimental:`  after the article heading. For example:

```asciidoc
= Deploying a Vaadin Application to Amazon Web Services
:experimental:
```

That enables so-called “experimental features” in the AsciiDoc compiler, one of which the `kbd` macro is: https://docs.asciidoctor.org/asciidoc/latest/macros/keyboard-macro/

The `experimantal` flag is automatically enabled in our documentation build, but when editing locally, you need to enable for each article separately. Or at least I’m not aware of a way to configure it globally in Atom.

dougjewsbury: Thanks, Jouni. I've done that, and removed "kbd" from accept.txt.

jouni (approved):

PR #1071, opened 2022-02-11
docs: map events section

sissbruecker:
## Description

Adds events section to the map component docs.

~Some of the features in here require using a snapshot.~
All features now should be in 23 beta2.

### DIFF
--- a/articles/ds/components/map/index.asciidoc
+++ b/articles/ds/components/map/index.asciidoc
@@ -186,10 +186,12 @@ The event contains details about the current viewport, such as the center, the z
 |https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/component/map/events/MapClickEvent.html[Map Click]
 |Triggered when the user clicks on the map.
 The event contains information about the clicked coordinates and other details of the mouse event.
+It also contains a list of all markers at that location in order to distinguish whether the click occurred on the map itself, or on a marker.
 
 |https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/component/map/events/MapFeatureClickEvent.html[Feature Click]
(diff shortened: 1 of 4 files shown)


articles/ds/components/map/index.asciidoc
```diff
@@ -168,6 +168,49 @@ For example, when adding multiple markers that use the same image, then it is su
...
+|https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/component/map/events/MapClickEvent.html[Map Click]
+|Triggered when the user clicks on the map.
+The event contains information about the clicked coordinates and other details of the mouse event.
+
+|https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/component/map/events/MapFeatureClickEvent.html[Feature Click]
+|Triggered when the user clicks on a marker.
+The event contains information about the clicked marker and other details of the mouse event.
+
+The term _feature_ refers to geographical features in a general sense.
+For now Map only supports markers as geographical features, so these are equivalent.
```
DiegoCardoso, line 195: Should it also mention that a click on the feature will trigger both `MapClickEvent` and `MapFeatureClickEvent`?

dougjewsbury (commented):
Looks good in terms of AsciiDoc.

articles/ds/components/map/index.asciidoc
sissbruecker: Maybe we should revise that behavior as part of https://github.com/vaadin/flow-components/pull/2650, let's discuss this in todays planning.

sissbruecker: Added the list of features at the clicked location to the click event in https://github.com/vaadin/flow-components/pull/2650, and added a paragraph to the map click event that this can be used to distinguish between map and feature clicks.

DiegoCardoso (approved):

PR #1097, opened 2022-02-21
Update FAQ

OlliTietavainenVaadin:
Added a Frequently Asked Question about Quarkus and JEE support

### DIFF
--- a/articles/faq.asciidoc
+++ b/articles/faq.asciidoc
@@ -64,4 +64,6 @@ You can still build hybrid applications with https://vaadin.com/flow[Vaadin Flow
 Hilla continues to be built and supported by Vaadin, just like Fusion.
 
 == Can I use Jakarta EE or Quarkus instead of Spring Boot?
-Currently neither Jakarta EE nor Quarkus support is on the roadmap, but we're aware that some people would be interested in those technologies as well. Let us know if you'd be interested in using either - if we know there's enough demand, that provides us guidance to adjust the priorities.
+Currently neither Jakarta EE nor Quarkus support is on the roadmap, but we are aware that some people would be interested in those technologies as well.
+Let us know if you would be interested in using either.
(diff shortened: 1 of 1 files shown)


articles/faq.asciidoc
```diff
@@ -62,3 +62,6 @@ Hilla used to be called Vaadin Fusion.
 It has since grown into its individual framework. 
 You can still build hybrid applications with https://vaadin.com/flow[Vaadin Flow^] and Hilla as before.
 Hilla continues to be built and supported by Vaadin, just like Fusion.
+
+== Can I use Jakarta EE or Quarkus instead of Spring Boot?
+Currently neither Jakarta EE nor Quarkus support is on the roadmap, but we're aware that some people would be interested in those technologies as well. Let us know if you'd be interested in using either - if we know there's enough demand, that provides us guidance to adjust the priorities.
```
knoobie, line 67: Somebody could link the feature request ticket for people to upvote (https://github.com/vaadin/hilla/issues/211)

I'm wondering; Should this ticket be split in Jakarta EE and another one for Quarkus? Now the thumbs up can be misinterpreted.

mrts: > I'm wondering; Should this ticket be split in Jakarta EE and another one for Quarkus? Now the thumbs up can be misinterpreted.

I would argue that both Jakarta EE and Quarkus can and should be supported with the same approach as they share many underlying technologies like CDI. Flow supports both with the Vaadin CDI Add-on. Hence I would vote for keeping just one ticket.

dougjewsbury, line 67:
```suggestion
Currently neither Jakarta EE nor Quarkus support is on the roadmap, but we are aware that some people would be interested in those technologies as well.
Let us know if you would be interested in using either.
If we know there is enough demand, that gives us some guidance to adjust the priorities.
```

marcushellberg (approved):

PR #1101, opened 2022-02-21
Add classname to download link in Flow Tutorial

tarekoraby:
Add a classname to use that to trigger a download event in Tag Manager.


tarekoraby:
@jouni is that the right way to add a classname to the link?

jouni (approved):

PR #1108, opened 2022-02-22
docs: add layers section for map component

sissbruecker:
## Description

Adds a section about layers to the map component docs.

### DIFF
--- a/articles/ds/components/map/index.asciidoc
+++ b/articles/ds/components/map/index.asciidoc
@@ -162,7 +162,15 @@ include::{root}/frontend/demo/component/map/map-layers.ts[preimport,hidden]
 
 [source,java]
 ----
-include::{root}/src/main/java/com/vaadin/demo/component/map/MapLayers.java[render,tags=snippet,indent=0,group=Java]
+include::{root}/src/main/java/com/vaadin/demo/component/map/MapLayers.java[render,tags=snippet1,indent=0,group=Java]
+
+...
+
+include::{root}/src/main/java/com/vaadin/demo/component/map/MapLayers.java[render,tags=snippet2,indent=0,group=Java]
+
+...
+
+include::{root}/src/main/java/com/vaadin/demo/component/map/MapLayers.java[render,tags=snippet3,indent=0,group=Java]
 ----
 
 [source,java]
--- a/articles/ds/components/map/index.asciidoc
+++ b/articles/ds/components/map/index.asciidoc
@@ -137,6 +137,41 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapSources.java[rend
 ----
 --
 
+== Layers
+
+The map component is pre-configured with a background layer, showing a map from the OpenStreetMap service by default, as well as a feature layer for quickly adding geographical features such as markers.
+It is possible to replace the background layer. Custom layers can be added to the map, for example to display overlays.
+
+By default, layers are rendered in the order that they were added to the map, with the first added layer rendered first, and the last added layer rendered last.
+For more fine-grained control of the layer rendering order, a `z-index` can be defined for each layer.
+Layers with a higher z-index are rendered on top of layers with a lower z-index.
+The exceptions are the background layer, which is always rendered first, and the feature layer, which is rendered with a z-index of `100` by default.
+
+The visibility of layers can be toggled, which can be an alternative to dynamically adding and removing them.
+Invisible layers are simply not rendered.
+For map overlays, a layer can be configured with a lowered opacity, so that the background layer is still semi-visible below the overlay layer.
+Note that some map services already use a lowered opacity in the map images that they provide, meaning that lowering the opacity of the layer is not needed in this case.
+
+The following example demonstrates how to replace the background layer, how to add several overlay layers to the map, and how to control their visibility:
+
+[.example]
+--
+[source,typescript]
+----
+include::{root}/frontend/demo/component/map/map-layers.ts[preimport,hidden]
+----
+
+[source,java]
+----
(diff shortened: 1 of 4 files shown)


articles/ds/components/map/index.asciidoc
```diff
@@ -137,6 +137,41 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapSources.java[rend
...
+The map component is pre-configured with a background layer, showing a map from the OpenStreetMap service by default, as well as a feature layer for quickly adding geographical features such as markers.
+It is possible to replace the background layer. Custom layers can be added to the map, for example to display overlays.
+
+By default, layers are rendered in the order that they were added to the map, with the first added layer rendered first, and the last added layer rendered last.
+For more fine-grained control of the layer rendering order, a `z-index` can be defined for each layer.
+Layers with a higher z-index are rendered on top of layers with a lower z-index.
+The exceptions are the background layer, which is always rendered first, and the feature layer, which is rendered with a z-index of `100` by default.
+
+The visibility of layers can be toggled, which can be an alternative to dynamically adding and removing them.
+Invisible layers are simply not rendered.
```
rolfsmeds, line 151: "Invisible layers are simply not rendered" feels a bit superfluous

articles/ds/components/map/index.asciidoc
```diff
@@ -137,6 +137,41 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapSources.java[rend
...
+For map overlays, a layer can be configured with a lowered opacity, so that the background layer is still semi-visible below the overlay layer.
+Note that some map services already use a lowered opacity in the map images that they provide, meaning that lowering the opacity of the layer is not needed in this case.
```
rolfsmeds, line 153: Propose to rephrase this simply as

> Note that map services may provide some layers with a lowered opacity by default.

articles/ds/components/map/index.asciidoc
```diff
@@ -137,6 +137,41 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapSources.java[rend
 ----
 --
 
+== Layers
+
+The map component is pre-configured with a background layer, showing a map from the OpenStreetMap service by default, as well as a feature layer for quickly adding geographical features such as markers.
+It is possible to replace the background layer. Custom layers can be added to the map, for example to display overlays.
```
rolfsmeds, line 143: Proposal to rephrase as:

> The default background layer can be replaced, and additional layers added on top of it, for example to display overlays.

rolfsmeds (changes requested):

articles/ds/components/map/index.asciidoc
```diff
@@ -137,6 +137,41 @@ include::{root}/src/main/java/com/vaadin/demo/component/map/MapSources.java[rend
...
+
+By default, layers are rendered in the order that they were added to the map, with the first added layer rendered first, and the last added layer rendered last.
+For more fine-grained control of the layer rendering order, a `z-index` can be defined for each layer.
```
dougjewsbury, line 146:
```suggestion
For finer-grained control of the layer rendering order, a `z-index` can be defined for each layer.
```

articles/ds/components/map/index.asciidoc
sissbruecker: Done

articles/ds/components/map/index.asciidoc
sissbruecker: Done

articles/ds/components/map/index.asciidoc
sissbruecker: Yeah, I wanted to hint at the fact that invisible layers do not impact performance, but I guess it's obvious. Removed.

articles/ds/components/map/index.asciidoc
sissbruecker: Simplified the wording a bit, but used `map data` instead of `layers`, as map services technically don't provide layers, and the map data term is consistent with a previous explanation in the document.

src/main/java/com/vaadin/demo/component/map/MapLayers.java
```diff
@@ -0,0 +1,131 @@
...
+        backgroundLayerGroup.addValueChangeListener(e -> {
+            LayerOption selectedOption = e.getValue();
+            XYZSource source = new XYZSource(new XYZSource.Options()
+                    .setUrl(selectedOption.getUrl())
+                    .setAttributions(List.of(selectedOption.getAttributions()))
+                    .setAttributionsCollapsible(false));
+            TileLayer layer = new TileLayer();
+            layer.setSource(source);
+            map.setBackgroundLayer(layer);
+        });
```
DiegoCardoso, line 84: I think the snippet can be more focused on the feature. Right now, I believe that there's more than needed to show feature API, like configuring the `RadioButtonGroup`. Maybe only the listener is enough. Same for the second radio button group to configure the layer visibility.

DiegoCardoso (changes requested):

src/main/java/com/vaadin/demo/component/map/MapLayers.java
sissbruecker: True, I split up the snippet into multiple ones. However I kept the section in the middle that shows how to add the custom overlay layers.

DiegoCardoso: Looks better, now 👍🏻

DiegoCardoso (approved):
LGTM

dougjewsbury (commented):
AsciiDoc looks good.

PR #1109, opened 2022-02-22
Replace Fusion docs with a landing page that points to Hilla docs

jouni:
Depends on a redirect rule that forwards all /docs/latest/fusion/* pages to /docs/latest/fusion


jouni:
Forgot to update the docs landing page, at least. Also, any cross references.

dougjewsbury (commented):
AsciiDoc looks good.

PR #1129, opened 2022-03-02
Feat/classic components

bennewi:
docs for Classic Components


articles/flow/upgrading/classic-components.asciidoc
```diff
@@ -0,0 +1,310 @@
...
+         <!-- Version can be omitted if using Vaadin version 22 and vaadin-bom -->
+        <version>2.0.0.beta1</version>
+    </dependency>
+</dependencies>
+----
++
+.Prerelease
+[NOTE]
+The pack is currently in prerelease state which means that you need to include the Vaadin prereleases Maven repository to your `<repositories>` listing.
+You can still use it on top of the latest stable version Vaadin 22.
```
knoobie, line 50: I would suggest to remove this note to reduce confusion and add the current release `23.0.0` (https://github.com/vaadin/classic-components/releases)

jouni (changes requested):
The path, `articles/flow/upgrading`, does not currently exist and does not specify an `index.adoc` file which would give that section a title and place it in the correct place in the Flow documentation navigation.

The `articles/flow/upgrading/overview.asciidoc` file is outdated and unnecessary, as all upgrading instructions have been moved to `articles/upgrading`.

At some point we discussed that perhaps the correct place for the Classic Components documentation is close to MPR docs. Those are now under `articles/tools/mpr`.

articles/flow/upgrading/classic-components.asciidoc
bennewi: yes we got the right numbers now

PR #1131, opened 2022-03-02
Store update config in URL

tarekoraby:
This will enable us to share the upgrade settings for specific versions and configurations more easily. It will also preserve the configuration upon navigating forward and then backward to the tool again.

### DIFF
--- a/articles/upgrading/index.adoc
+++ b/articles/upgrading/index.adoc
@@ -54,7 +54,8 @@ This is the description of the classnames:
 [class*=flow]:is(div, .openblock, h3), 
 [class*=fusion]:is(div, .openblock, h3), 
 [class*=spring]:is(div, .openblock, h3), 
-[class*=ts]:is(div, .openblock, h3)
+[class*=ts]:is(div, .openblock, h3),
+[class*=styling]:is(div, .openblock, h3)
 {
     display: none;
 }
--- a/articles/upgrading/index.adoc
+++ b/articles/upgrading/index.adoc
@@ -46,15 +46,15 @@ This is the description of the classnames:
 
 ++++
 <style>
-[class*=all]:is(div, .openblock), 
-[class*=v1]:is(div, .openblock),
-[class*=v2]:is(div, .openblock),
-[class*=v3]:is(div, .openblock),
-[class*=v4]:is(div, .openblock),
-[class*=flow]:is(div, .openblock), 
-[class*=fusion]:is(div, .openblock), 
-[class*=spring]:is(div, .openblock), 
-[class*=ts]:is(div, .openblock)
+[class*=all]:is(div, .openblock, h3), 
+[class*=v1]:is(div, .openblock, h3),
+[class*=v2]:is(div, .openblock, h3),
+[class*=v3]:is(div, .openblock, h3),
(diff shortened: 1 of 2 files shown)


jouni:
Seems to work fine. The only issue I notice is that if the user clicks on any of the heading anchor links (`#`), the query params / config if lost, unfortunately.

Screen Shot 2022-03-03 at 10 21 29

jouni:
We might be able to work around that by going through all the link href attributes, and rewriting them from `/upgrading-guide/#foobar` to just `#foobar`. But it looks like the current TS logic loses the hash from the URL, if such an anchor is clicked.

tarekoraby:
Do we even need the anchors on this page? Can't we just remove them all?

jouni:
Your call. Might be nice for some use cases, that we could link to specific instructions.

jouni:
But it’s better to remove them if we can't make them work now.

tarekoraby:
I removed the links using `[discrete]`. We shouldn't need those to share instructions as we can now do that via the URL. For example, by sharing something like: `docs/upgrading?from=22&to=23&isFlow=false&isFusion=true&isSpring=true&isTypeScript=true&isCustomStyling=false&isInstructionsDisplayed=true`

jouni:
I wonder what changed, but now when I navigate to the preview URL, https://pr-1131.dcmp4ajsd49kq.amplifyapp.com/, and click "Upgrading" in the menu, I see all the v22/23 styling update instructions automatically.

tarekoraby:
That's strange. I don't see that locally.

tarekoraby:
Now I see the same locally after restarting the server.

tarekoraby:
Fixed.

jouni:
> We shouldn't need those to share instructions as we can now do that via the URL. For example, by sharing something like:
> ```
> docs/upgrading?from=22&to=23&isFlow=false&isFusion=true&isSpring=true&isTypeScript=true&isCustomStyling=false&isInstructionsDisplayed=true
> ```

This doesn't work for me, it always reverts back to the default settings: https://pr-1131.dcmp4ajsd49kq.amplifyapp.com/upgrading?from=22&to=23&isFlow=false&isFusion=true&isSpring=true&isTypeScript=true&isCustomStyling=false&isInstructionsDisplayed=true

tarekoraby:
That's strange. It works locally. It's also working on Amplify if one changes the settings in the tool, then navigate elsewhere, then hit back.

tarekoraby:
Yeah, it seems to be an Amplify-related issue: https://github.com/aws-amplify/amplify-console/issues/97.

The workaround is to add a "/" before the query params. 
So this seems to work:
https://pr-1131.dcmp4ajsd49kq.amplifyapp.com/upgrading/?from=22&to=23&isFlow=false&isFusion=true&isSpring=true&isTypeScript=true&isCustomStyling=false&isInstructionsDisplayed=true

jouni (approved):
Let’s merge this and fix remaining issues in separate PRs.

tarekoraby:
I already added the logic so that a '/' is added when the tool is updating the URL. That fixed it in Amplify, and hopefully, this will fix it in production as well.

PR #1150, opened 2022-03-09
refactor: render basic Dialog example in iframe, instead of fake preview

jouni:
Remove the fake preview example, which is outdated today and could easily get out of date in the future as well.


articles/ds/components/dialog/index.asciidoc
```diff
@@ -12,33 +12,30 @@ Dialog is a small window that can be used to present information and user interf
...
+++++
+<style>
+.basic.example iframe {
+  height: 400px;
+}
+</style>
+++++
+
 == Modality
 A modal Dialog blocks the user from interacting with the rest of the user interface while the Dialog is open, as opposed to a non-modal Dialog which does not block interaction.
```
dougjewsbury, line 40:
```suggestion
A modal Dialog blocks the user from interacting with the rest of the user interface while the Dialog is open, as opposed to a non-modal Dialog, which does not block interaction.
```

PR #1152, opened 2022-03-11
Shifted all headings in Upgrade Guide up 1 level

rolfsmeds:
I.e. changed all H3s H2s, H4s to H3s and H5s to H4s, since the lower heading levels are difficult to distinguish from each other, making the hierarchy unclear and the guide harder to read, and because they were incorrectly levelled to begin with.


jouni (approved):
Waiting for the preview build, just to make sure everything looks okay.

tarekoraby (approved):

PR #1153, opened 2022-03-11
Upgrade Spring Boot to 2.6.4

tarekoraby:
NOT tested locally. @jouni is it enough if the cloud deployment works?


jouni:
Sorry for the delay. The preview build makes sure that at least the Java code compiles. If that’s enough, then it’s good to go.

jouni (approved):

PR #1188, opened 2022-03-17
Document necessary Spring security config changes when migrating to V23

### LINKED ISSUE #1184: Document necessary Spring security config changes when migrating to V23
The following need to be added to the list of ignored URLs:
```
"/offline-stub.html",
"/sw-runtime-resources-precache.js",
```

As a side mention, we should recommend using the built-in defaults from `VaadinWebSecurityConfigurerAdapter` as documented in https://vaadin.com/docs/latest/flow/integrations/spring/view-based-access-control

tarekoraby:
Fix #1184.


joheriks (approved):

joheriks:
Content looks fine, but there are conflicts to resolve.

PR #1191, opened 2022-03-18
Remove link to gitpod from quick start tutorial

tarekoraby:
Let's test removing the gitpod link for few weeks to see whether it has any positive or negative changes on the performance of this page. I speculate that the gitpod link might actually be distracting as users are not used to online IDEs in general, and on VS Code-like IDEs in particular.


dougjewsbury (commented):
AsciiDoc looks good.

jouni (approved):

PR #1192, opened 2022-03-18
Rename "Custom Serialization" to "Custom Type Conversion"

tarekoraby:
NOTE that I couldn't verify these changes locally as I'm still unable to run the docs-app there.


dougjewsbury (commented):
AsciiDoc looks good.

PR #1211, opened 2022-03-22
Styling for expandable section toggles

rolfsmeds:
Styling for the `<summary>` element (i.e. the toggle) of expandable sections:
image


dspublisher/theme/global.css
```diff
@@ -292,9 +292,22 @@ html[theme~='dark'] .card.card {
...
+  cursor: pointer;
+  font-size: var(--docs-font-size-m) !important;
+}
+
+[class^='ArticleLayout'] details:not([open]) > summary.title {
+  color: var(--docs-link-color);
+}
+
+[class^='ArticleLayout'] details[open] > summary.title {
+  color: var(--color-secondary-text);
```
jouni, line 312: Any reason why not use this property instead?

```suggestion
  color: var(--docs-secondary-text-color);
```

dspublisher/theme/global.css
```diff
@@ -292,9 +292,22 @@ html[theme~='dark'] .card.card {
...
+[class^='ArticleLayout'] .guilabel {
   font-family: var(--docs-font-family-ui);
   font-size: 0.875em;
   line-height: 1.25;
   font-weight: 600;
 }
+
+[class^='ArticleLayout'] details > summary.title {
+  cursor: pointer;
+  font-size: var(--docs-font-size-m) !important;
```
jouni, line 304: If we used some other class name than `title`, we could avoid this `!important`.

Personally, I might add the class on the `<details>` element instead, if we want to add a variant for it. 

That said, I would just style the raw elements, and let all of them look the same across the docs. I’m not aware of any other places where it is currently used than in [Styling components](https://vaadin.com/docs/latest/ds/customization/styling-components/#sub-components), and I think that could look the same as the new ones in the upgrade guide.

dspublisher/theme/global.css
rolfsmeds: Ah, didn't notice both existed.

dspublisher/theme/global.css
rolfsmeds: Anyhoo, removed the .title from selectors now that the prefixes make them specific enough without

jouni: Oh, right, so it seems. I didn't realize there's this `[%collapsible]` macro in AsciiDoc. I used raw HTML with a passthrough block before, which doesn't add the title class name.

jouni: I think we should add some nice styles for the collapsible block in DSP built-in styles.

rolfsmeds: True that. Made a ticket about it https://github.com/vaadin/docs-app/issues/328

tarekoraby (approved):

PR #1244, opened 2022-03-29
docs: remove fixed vaadin version from the upgrade guide

tomivirkki:
The version number in the asciidoc doesn't get updated automatically.

Another option would be to use `...change the Vaadin version to {moduleMavenVersion:com.vaadin:vaadin}` but then the version would automatically get bumped to the next major (24, sometime in the future) together with the docs project.

### DIFF
--- a/articles/v14-23-upgrade/essential-steps/instructions/_common.adoc
+++ b/articles/v14-23-upgrade/essential-steps/instructions/_common.adoc
@@ -3,8 +3,7 @@
 
 * Delete the `node_modules` folder and either lock file: `package-lock.json` (with npm) or `pnpm-lock.yaml` (with pnpm)
 
-* Edit the `pom.xml` file and change the Vaadin version to `23.0.2`.
-
+* Edit the `pom.xml` file and change the Vaadin version to the latest Vaadin 23 version available.
 
 == Update to Java 11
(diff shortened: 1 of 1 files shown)


jouni:
We could also ask @ZheSun88 to include this page in the automation that updates this information in other places. See this PR for example: https://github.com/vaadin/docs/pull/1240

tomivirkki:
> We could also ask @ZheSun88 to include this page in the automation that updates this information in other places

I guess it would have the same problem as using `{moduleMavenVersion:com.vaadin:vaadin}`: This 14 -> 23 upgrade guide could end up suggesting Vaadin 24

tarekoraby (approved):

tarekoraby:
Would it be possible for the docs to maintain a list of the latest Vaadin versions similar to the one used in the upgrade tool: https://github.com/vaadin/docs/blob/920e75b25191997818ce1b5b52f5e8800f73cf10/frontend/demo/upgrade-tool/upgrade-tool.ts#L16. This list is currently being updated by the vaadin-bot. 

If we have a list like this at the level of the docs-app (not just the upgrade tool), then we could perhaps use it in places like the current v14 > 23 upgrade pages.

PR #1245, opened 2022-03-29
Remove incorrect default text

Artur-:


articles/shared/guide/configuration/_index.asciidoc
```diff
@@ -125,7 +125,6 @@ Uses bytecode scanning, which increases application startup time.
 Configuration name for the parameter that determines whether Vaadin should automatically register servlets needed for the application to work.
 
 `disable-xsrf-protection`::
-The default is `true`.
```
tarekoraby, line 128: Why not set the value to `false` rather than removing the note altogether?

Artur-: None of the other options state the default value this way and the text below specifies what the default is

tarekoraby: Some do, some don't 
image

tarekoraby (approved):

articles/shared/guide/configuration/_index.asciidoc
tarekoraby: Anyways, it's your call since it's described a couple of sentences after.

PR #1246, opened 2022-03-29
Reference Vaadin 14 to 23 upgrade instructions from upgrade tool

tarekoraby:


tarekoraby:
I had to remove the `[discrete]` from the headers because otherwise, the structure of the document gets messed up, and the instructions are no longer displayed correctly according to the selected versions. Instead of `[discrete]`, the links are removed manually from the tool.

jouni:
Perhaps a proper fix would be to not prefix in-page links/anchors in DSP/docs-app. 

I’m not 100% sure, but I think Gatsby’s `<Link>` component prefixes in-page anchor links here: https://github.com/vaadin/docs-app/blob/master/src/components/ArticleContent.tsx#L89-L104

tarekoraby:
If we choose to fix it, then it shouldn’t be for the sake of the upgrade tool given that it will possibly retire soon.

jouni:
Right. In that case, let’s use this workaround you implemented.

PR #1256, opened 2022-04-01
Restructure upgrade instructions and delete upgrade tool

tarekoraby:
This PR:

1. Deletes the existing upgrade tool and upgrading instructions
2. Move the V14 to V23 instructions to be the main upgrade instructions available under /upgrade


articles/upgrading/essential-steps/index.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+This chapter contains the list of changes required to upgrade a Vaadin 14 application to Vaadin 23.
+It includes the essential instructions needed for the project to compile, run, and behave in Vaadin 23 as it did in Vaadin 14.
+
+The <<recommended-changes#,Recommended Changes>> chapter lists further changes that are recommended to conform with Vaadin 23 best practices and to avoid the usage of deprecated APIs and features.
+
+
+[NOTE,role=small]
+This guide provides instructions to upgrade from Vaadin 14 to 23.
+If you are upgrading from an earlier version, follow the https://vaadin.com/docs/v14/flow/upgrading/[upgrading instructions here] to bring your application to version 14 first, before continuing with this guide.
```
tarekoraby, line 16: Is there a recommended way to referecne v14 docs so that Vale doesn't complain?

tomivirkki: I think this issue can be fixed by adding `link:` prefix to the URL

tarekoraby: Vale is still complaining after adding `link:`

jouni: Vale checks for strings starting with https://vaadin.com/docs, and complains about those.

The only way currently to avoid that nag is to use a relative URL, so you need to "guess" the right amount of `../` to use. Which is not ideal. Otherwise just ignore the error.

jouni: If you want, you can try to do something with the check: https://github.com/vaadin/docs/blob/latest/.github/styles/Vaadin/CrossReference.yml

jouni:
Apparently this needs to be tested locally (preview build seems stalled/broken).

articles/upgrading/essential-steps/index.adoc
tomivirkki: Cross-references should probably not be used between two different docs sites since they're different SPAs

jouni: "Cross ref" is just a fancy term for a link. But yes, semantically it is incorrect to use them for linking between docs versions.

jouni (approved):
Would appreciate if someone else, who has worked on the guide, approved as well.

articles/upgrading/essential-steps/index.adoc
tarekoraby: Rverted back to using `link:`

tarekoraby:
This is how the upgrading chapter looks like now. 

image

rolfsmeds (approved):
Look good enough for now. Let's just get this merged already :)

tomivirkki (approved):

PR #1269, opened 2022-04-05
Update Spring Boot to 2.6.6

tarekoraby:


tarekoraby:
NOT tested locally.

jouni:
Any reason to believe why this would cause issues in production?

tarekoraby:
Not really. I'm just not 100% sure what are all the ways in which this should be tested. At least it seems to work fine on Amplify.

jouni (approved):

PR #1288, opened 2022-04-08
Added note about Z-index incompatibility

TatuLund:


jouni:
I tried to fix the Vale errors. I don't think it makes sense to convert all headings to use title case at this point. We can consider doing that when this page is under more thorough review.

articles/tools/mpr/configuration/limitations.asciidoc
```diff
@@ -6,6 +6,9 @@ layout: page
 
 = Limitations of MPR
 
+// Disable 
+pass:[<!-- vale Vaadin.Headings = NO -->]
+
 Using MPR in your project to port a legacy application to Vaadin Flow has some known limitations. This is the current list of limitations - keep in mind that it can change over time as new features are implemented.
```
dougjewsbury, line 12:
```suggestion
Using MPR in your project to port a legacy application to Vaadin Flow has some known limitations.
This is the current list of limitations.
Keep in mind that it may change over time as new features are implemented.
```

articles/tools/mpr/configuration/limitations.asciidoc
```diff
@@ -6,6 +6,9 @@ layout: page
...
 
 == It's possible to add a legacy component to a Flow layout, but not a Flow component in a legacy layout
```
dougjewsbury, line 14:
```suggestion
== It is possible to add a legacy component to a Flow layout, but not a Flow component to a legacy layout
```

articles/tools/mpr/configuration/limitations.asciidoc
```diff
@@ -89,4 +92,8 @@ Browser support for MPR applications is based on the running Vaadin version. For
 
 Since Vaadin Board 1.1.0 requires Polymer 2 and Vaadin 14 requires Polymer 3, Vaadin Board is not compatible with Vaadin 14 and cannot be used as a legacy component. It works if Vaadin 14 runs in compatibility mode.
```
dougjewsbury, line 93:
```suggestion
Since Vaadin Board 1.1.0 requires Polymer 2 and Vaadin 14 requires Polymer 3, Vaadin Board is not compatible with Vaadin 14 and cannot be used as a legacy component.
It works if Vaadin 14 is running in compatibility mode.
```

articles/tools/mpr/configuration/limitations.asciidoc
```diff
@@ -89,4 +92,8 @@ Browser support for MPR applications is based on the running Vaadin version. For
...
 
+== Z-indexes of the modern Vaadin Design System are not aligned with legacy
+
+Components using overlays / popups in Vaadin 7 and Vaadin 8 are organized in different z-indexes than similar components in the modern Vaadin Design System. Thus it can be possible that some combinations of them does not work. For example using such Vaadin 7 / 8 components with LegacyWrapper in Dialog may work un-expectly.
```
dougjewsbury, line 97:
```suggestion
Components using overlays / popups in Vaadin 7 and Vaadin 8 are organized in different z-indexes from similar components in the modern Vaadin Design System.
Thus it is possible that some combinations of them do not work.
For example, using such Vaadin 7 / 8 components with LegacyWrapper in Dialog may not work as expected.
```

jouni (approved):

PR #1298, opened 2022-04-11
Removed DS Upgrade Guide

rolfsmeds:


rolfsmeds:
> A redirect would be needed if we remove the page completely. Another option is to replace the content on this page with a link to the main upgrading guide.

Good point! Seems a bit silly to have a page just for the link, so I'd prefer a redirect. How do we do that?

tarekoraby:
I can create a PR for the redirect with vaadincom. Should it redirect to https://vaadin.com/docs/latest/upgrading/essential-steps?

tarekoraby (approved):

tomivirkki (approved):

PR #1299, opened 2022-04-11
Added note about CB and RB using Lumo properties instead of inherited styles

rolfsmeds:
(Not sure if this is worth mentioning in the upgrade guide, but it's one of few things that were in the DS guide but not in the new Upgrade Guide.)


dougjewsbury (commented):
AsciiDoc looks good.

tomivirkki (approved):

PR #1304, opened 2022-04-12
Optimize doc descriptions for search engines

samie:


articles/application/index.asciidoc
```diff
@@ -2,9 +2,10 @@
 title: Creating an Application
 order: 20
 section-nav: expanded
+page-title: "Creating a Hilla Application"
+description: "In Hilla application architecture, TypeScript views connect with server-side Java endpoints that can provide business data and logic."
```
tarekoraby, line 6:
```suggestion
description: "In Hilla applications, TypeScript views connect with server-side Java endpoints that can provide business data and logic."
```

jouni:
I’m a little torn with the term "Hilla Framework". I would not like to capitalize it, but it also looks a bit poor in things like "Hilla framework Security". 

I assume you want to add the "framework" word for better search matches. Otherwise it feels unnecessary.

tarekoraby (approved):

PR #1310, opened 2022-04-14
Fix Node installation page

tarekoraby:
The shared document is using an id that is already used in` articles/flow/configuration/node-js.asciidoc`. There is also one erroneous display of the filename 

image

### DIFF
--- a/articles/shared/guide/install/_node-js.asciidoc
+++ b/articles/shared/guide/install/_node-js.asciidoc
@@ -15,7 +15,6 @@ If Node.js is found globally, Vaadin validates that it is a supported version; i
 We recommend using the latest LTS version.
 A project-local installation will always take precedence over a global or `~/.vaadin` installation.
 
-[[node.installation]]
 You can force Node.js installation to the `~/.vaadin` folder via the `require.home.node` property.
 This property sets the Maven `requireHomeNodeExec` parameter value, so you can configure the Maven goal using `<requireHomeNodeExec>true</requireHomeNodeExec>`.
 To force node installation into the home directory in development mode, you should use the `vaadin.require.home.node` system property or the `require.home.node` web init parameter.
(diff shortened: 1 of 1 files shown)


articles/shared/guide/install/_node-js.asciidoc
```diff
@@ -74,6 +73,6 @@ If you are using https://travis-ci.org/[Travis] as a Continuous Integration serv
 . Specify the version via Travis configuration in [filename]#.travis.yml#
 . Install Node.js automatically through Vaadin
 
-Refer to https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions[Specifying Node.js versions] in the Travis documentation for how to specify the Node version via the [filename]#.travis.yml# file.
+Refer to https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions[Specifying Node.js versions] in the Travis documentation for how to specify the Node version via the [filename]`.travis.yml` file.
```
jouni, line 76: Actually, we decided that file/folder names should be marked using `#`. Maybe that is unnecessary complexity to authoring, but I somehow felt that file names could be rendered as plain text in GitHub, rather than monospace as proper application code.

tarekoraby: But at least in this case, using the "#" causes the styling to be messed up (see the screenshot above).

jouni, line 76: Actually, we decided that file/folder names should be marked using `#`. Maybe that is unnecessary complexity to authoring, but I somehow felt that file names could be rendered as plain text in GitHub, rather than monospace as proper application code. 

This should be the correct fix for the AsciiDoc formatting: 

```suggestion
Refer to link:https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions[Specifying Node.js versions] in the Travis documentation for how to specify the Node version via the [filename]#.travis.yml# file.

```

jouni: Sorry, I ended up moving my comment. Yeah, that can be fixed by using the explicit `link:` macro.

jouni (approved):

PR #1313, opened 2022-04-14
docs: fix broken charts cross-reference

web-padawan:
Looks like this warning caused AWS build to fail.

```
2022-04-14T06:52:50.076Z [WARNING]: error Broken cross reference [1m/ds/components/charts/basic-use.asciidoc[0m in ds/components/charts/configuration.asciidoc
```


tarekoraby (approved):

jouni:
I updated the cross ref syntax to be aligned with the others. We don't really use `xref:`, though sometimes it is needed if you need to define a class/role on the link.

PR #1315, opened 2022-04-14
Add note about the `--auth` option

tarekoraby:


tarekoraby:
@marcushellberg approved on slack.

PR #1341, opened 2022-04-27
refactor: update npm scripts to use the new dsp

tomivirkki:
This PR makes the dspublisher scripts use the NPM package release of DSP (instead of the Docker image).

Fixes https://github.com/vaadin/docs-app/issues/281
Fixes https://github.com/vaadin/docs-app/issues/331
Fixes https://github.com/vaadin/docs-app/issues/330
Closes https://github.com/vaadin/docs-app/issues/307
Fixes https://github.com/vaadin/docs-app/issues/278
Fixes https://github.com/vaadin/docs-app/issues/269


dspublisher/dspublisher-scripts.js
```diff
@@ -0,0 +1,294 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
tomivirkki, line 1: This file was added to hide the verbose output that comes from starting `docs` and `docs-app` in parallel in favor of a single progress bar + status message.

jouni:
> It also hides the DSP process output and displays a nice progress bar instead:

I assume warnings and errors will still be logged in addition to this progress bar?

dspublisher/dspublisher-scripts.js
```diff
@@ -0,0 +1,294 @@
...
+        // The current command was finished, run the next one
+        runCommand(commands, index + 1);
+      } else {
+        // All commands were run, finish
+        finish();
+      }
+    });
+
+    // TODO: --verbose option
+    process.stdout.on('data', (data) => {
```
tomivirkki, line 250: > I assume warnings and errors will still be logged in addition to this progress bar?

Errors do get logged, everything else gets hidden for now. Currently, the script runs a command, then listen to the output stream here, and reacts when the stream contains specific strings.

jouni: Warnings regarding broken cross references are the most useful at the moment, and I assume our own documentation authors rely on those. Sure, we can still run docs-app manually to see those, but regular users will have a difficult time tracking down those.

tomivirkki: The intention is to still add the verbose mode as suggested [here](https://github.com/vaadin/docs-app/issues/281#issuecomment-1062818834).

But I agree that the default mode should probably also show warnings.

jouni:
After showing `Ready. Open http://localhost:8000 in the browser.`, we could also show instructions how to stop the process.

articles/tools/dspublisher/devmode.asciidoc
```diff
@@ -5,15 +5,15 @@ order: 30
...
 
 [source,terminal]
 ----
 npm run dspublisher:start
 ----
 
-Note that the first run needs to download the Design System Publisher Docker image and install a lot of (Maven and npm) dependencies so it might take a good while (about ten minutes or so, depending on the network connection speed) to have it fully running.
-Following start-ups are quicker (about four minutes) as the dependencies are already available.
+Note that the first run needs to download the Design System Publisher package and install a lot of (Maven and npm) dependencies so it might take a good while to have it fully running.
+Following start-ups are quicker as the dependencies are already available.
```
dougjewsbury, line 16:
```suggestion
Subsequent start-ups are quicker, as the dependencies are already available.
```

jouni (approved):

jouni:
We should cherry-pick this to the `dsp` branch as well, right?

tomivirkki:
> We should cherry-pick this to the dsp branch as well, right?

That's right

PR #1375, opened 2022-05-10
feat: Add FormManager documentation

MatthewVaadin:
Adds a new page and code sample for the `FormManager` class in Collaboration Engine.


articles/ce/managers/form-manager.asciidoc
```diff
@@ -0,0 +1,33 @@
+---
+title: Form Manager
+order: 3
+layout: page
+---
+
+[[ce.form-manager]]
+= Handle Form Values and Field Highlighting with Your Own Logic
+
+Collaboration Engine includes a manager to set property values and editor highlighting in a form, and to react to changes in these.
```
heruan, line 10: I would say `field highlighting` since we use _field_ to describe the form components.

articles/ce/managers/form-manager.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+----
+include::{root}/src/main/java/com/vaadin/demo/ce/FormManagerExample.java[tags=snippet,indent=0]
+----
+
+<1> Creating a [classname]`FormManager` takes:
+* a component to bind the manager connection context to the UI,
+* an instance of [classname]`UserInfo` that represents the current local user, and
+* the ID of the topic to connect to.
+
+<2> The [methodname]`highlight()` method is used to mark the local user as editing a property.
```
heruan, line 23: Let's also explain what would happen to fields bound to the same property with a `CollaborationBinder`, e.g.

> […] a field component bound to the same property via a `CollaborationBinder` connected to the same topic will be highlighted and show the name of the user.

articles/ce/managers/form-manager.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+
+<3> The [methodname]`setHighlightHandler()` method sets a callback to handle a user editing a property.
```
heruan, line 25: Same here, e.g.

```suggestion
<3> The [methodname]`setHighlightHandler()` method sets a callback to handle a user editing a property, for example focusing a field component bound to the same property via a [classname]`CollaborationBinder` instance connected to the same topic.
```

articles/ce/managers/form-manager.asciidoc
```diff
@@ -0,0 +1,33 @@
...
+The property name and the `UserInfo` of the user are provided by the context passed as the argument to the callback.
+
+<4> The callback returns another callback, a registration, that is called when the same user stops editing a property.
+
+<5> The [methodname]`setValue()` method is used to set the value of a property.
```
heruan, line 30: Also here, something like

> the value will be set on all field components bound to the same property […]

heruan (changes requested):
Looks good! I would add a general introduction on how this is connected to the `CollaborationBinder` (with a link to its documentation), so that it's clear that when they're connected to the same topic the changes are reflected there and viceversa.

heruan (dismissed):

jouni:
I added the "Language not checked" label as that is the process we use for any new content introduced in the docs. I realize this was written by a native English speaker, so I doubt there are many issues.

heruan:
This needs next platform pre-release to remove the explicit dependency added to the POM (for the Java example to compile).

jouni:
@dougjewsbury, perhaps you want to review this when you do a full review of the whole CE section in docs, instead of separately now? Your choice.

dougjewsbury (commented):
AsciiDoc looks good.

heruan (approved):

PR #1379, opened 2022-05-12
docs: Added base documentation for Testbench UI Unit testing

mcollovati:
Part of vaadin/testbench#1330

PR #982, opened 2021-12-13
Update unit-and-integration-testing.adoc

mvysny:
Fixes #981

## Description

Please list all relevant dependencies in this section and provide summary of the change, motivation and context.

Fixes #981 

## Type of change

- [ ] Bugfix
- [x] Feature

## Checklist

- [ ] I have read the contribution guide: https://vaadin.com/docs/latest/guide/contributing/overview/
- [ ] I have added a description following the guideline.
- [ ] The issue is created in the corresponding repository and I have referenced it.
- [ ] I have added tests to ensure my change is effective and works as intended.
- [ ] New and existing tests are passing locally with my change.
- [ ] I have performed self-review and corrected misspellings.

#### Additional for `Feature` type of change

- [ ] Enhancement / new feature was discussed in a corresponding GitHub issue and Acceptance Criteria were created.

PR #1385, opened 2022-05-16
Update unit-and-integration-testing.adoc

mvysny:
Add link to Karibu-Testing


jouni:
See https://github.com/vaadin/docs/pull/982#issuecomment-1002681486

I’m still hesitant in promoting unofficial solutions in the official docs. Now that we are working on a similar official solution (as far as I know), I might wait for that instead, or at least add a note about that (Karibu-testing not being official and an official solution being in developed).

tarekoraby:
Doesn't #1379 already mention the Vaadin official offer that is based on Karibu?

mvysny:
@jouni we promote all sorts of third-party libraries; I think in the name of being helpful to the customers it's better to promote stuff than staying rigid (and unhelpful). Also the idea is validated and is getting inside Vaadin itself; I propose to merge this and change the reference to TestBench later on.

jouni (approved):

PR #1433, opened 2022-06-04
Remove extra spaces from Route template parameters

tarekoraby:
Remove what appears currently as spaces in the the route template parameters: 

image


jouni:
I’m not sure what the intention has been with the original syntax. I assume it has been to separate the “user defined parts” (identified, modifier, regex) from the rest of the syntax (`:`, `(`, `)`). I agree that it does seem very effective, rather quite messy.

jouni (approved):

PR #1434, opened 2022-06-06
fix: enhance dsp node.js version check

tomivirkki:
While DSP can be started with Node.js versions 17 and 18, the live Java examples refuse to work, most probably due to a [Webpack issue](https://stackoverflow.com/questions/69665222/node-js-17-0-1-gatsby-error-digital-envelope-routinesunsupported-err-os). Make the preconditions check require Node.js version between 14 and 16.


tarekoraby:
Wouldn't it be enough to show a warning rather than blocking people from using DSP completely?

tomivirkki:
> Wouldn't it be enough to show a warning rather than blocking people from using DSP completely?

Updated the PR to show a warning instead

dspublisher/dspublisher-scripts.js
```diff
@@ -51,12 +51,17 @@ async function checkPreConditions() {
 
     // Verify the Node.js version is supported
     const MINIMUM_NODE_VERSION = 14;
```
tarekoraby, line 53: This is not related to this PR, but I think Flow doesn't work with Node 14.

tarekoraby (approved):

dspublisher/dspublisher-scripts.js
tomivirkki: Flow automatically downloads and uses a Node.js version it needs if it's not available.

tomivirkki: Flow doesn't really have a requirement for even having Node.js installed at all. One benefit of not using Node 14 with DSP is that the first start would be a bit faster (Flow wouldn't need to download Node 16) but that's probably not a good enough reason to block Node.js 14 users from running DSP.

PR #1447, opened 2022-06-07
Simplify route-parameters instructions

### LINKED ISSUE #1410: Simplify instructions for including parameters into a route
Flow has three separate articles that describe three different ways in which one can include parameters into a route: [Route Templates](https://vaadin.com/docs/latest/flow/routing/templates), [Typed Parameters](https://vaadin.com/docs/latest/flow/routing/url-parameters), and [Query Parameters](https://vaadin.com/docs/latest/flow/routing/query-parameters). This, however, can be confusing to our users as we don't clarify the conditions and use cases in which someone should prefer one approach over the others.

To simplify things, let's implement @Legioth's [suggestion](https://vaadin.slack.com/archives/C6RAXJATF/p1653469160395679):
- As it is the easiest way and as it should work for most common use cases, let's recommend typed parameters as the main way to include parameters into a route. 
- Route Templates should be noted as an ultimate tool suitable for advanced use cases based on path segments.
- Query Parameters should be mentioned as suitable in those cases where it is necessary to have explicit name-value pairs or several optional parameters.

For typed parameters, we could consider de-emphasizing the wildcard and optional parameters since those are better covered using the more powerful Route Templates.

tarekoraby:
fix #1410, #1450, and #1452.


jouni:
To me these are the same page, and therefore redundant.

**Index**:
Screen Shot 2022-06-09 at 16 25 46

**Overview**:
Screen Shot 2022-06-09 at 16 25 56

Also, I dislike the ambiguous "Overview" page title, which is not very descriptive in the search results. 

Screen Shot 2022-06-09 at 16 27 19


This is one of the original reasons why I started to eliminate the overview pages from all the other sections, and before that unifying the page titles so that they say the section name instead of "Overview”.

Here’s an example from the v14 branch of the latter. Notice, that the navigation/breadcrumb label/title can still say "Overview", but the page H1 uses the section name.


Screen Shot 2022-06-09 at 16 29 59

tarekoraby:
@jouni thanks for the feedback! I made a new attempt at restructuring things.

@dmitrilc please take a look at the new structure to see if it makes sense.

dmitrilc:
The current state of the PR is acceptable to me, and has achieved its goal, which is to simplify the routing instructions. Further improvements can be added later.

I don't dislike the **Overview**-type sections (*sometimes called **Quick Start**, **Getting Started**, **The Basics**, **Common Tasks** etc* in Spring, Angular, Android, Swift, Typescript docs). I am approaching the Vaadin docs from the perspective of a non-Vaadin-expert web developer. Furthermore, I usually only care about quickly learning the basics to implement what I need. We can even make the **Overview** title more descriptive by calling it **Route Parameters Overview**, for example.
image

Following a similar pattern to the v14 doc also seem acceptable. 

While the **index** page and the **Overview** page are similar, the **index** page was not obviously accessible/clickable, especially when you are currently in a sub-topic within that index (*mouse hovering does not change appearance of direct parent*). 
Vaadin Index no Highlight

The two examples below show a different behavior.

https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-introduction
spring index with higlight

https://angular.io/guide/routing-overview
angular index with highlight

PR #1464, opened 2022-06-09
chore: Added a note with link for testing with spring

mcollovati:


articles/tools/testbench/ui-unit-testing/getting-started.asciidoc
```diff
@@ -13,6 +13,12 @@ layout: page
 To start unit testing your Vaadin application, create a class that extends [classname]`UIUnitTest` (for JUnit 5) or [classname]`UIUnit4Test` (for JUnit 4).
 The base class instantiates UI along with all the necessary Vaadin environment, which will be available on your test methods.
 
+.Testing with Spring
+[NOTE]
+This guide shows how to test UI in plain Java projects.
+For information about testing with Spring consult the <<{articles}/tools/testbench/ui-unit-testing/spring-tests#, specific documentation>>.
```
caalador, line 19:
```suggestion
For information about testing with Spring or Spring Boot, consult the <<{articles}/tools/testbench/ui-unit-testing/spring-tests#, Spring specific documentation>>.
```
Might be good to emphasize a bit more here with spring.

caalador (approved):

tarekoraby:
@mcollovati, @caalador: for future reference, since most of our users are actually using Spring Boot, we should try to prioritize SB-specific instructions (and we can then mention plain Java instructions in a separate guide).

For example, this is what we are doing now with the [Security Helpers instructions](https://vaadin.com/docs/latest/flow/security/enabling-security): 

image

PR #1468, opened 2022-06-13
Adding link to java.util.regex.Pattern class

### LINKED ISSUE #1455: Add external link for URL Template Regex implementation details
At https://vaadin.com/docs/latest/flow/routing/templates#route-parameters-matching-a-regex, regarding this sentence 

`The regex is compiled using java.util.regex.Pattern and should not contain segment delimiter sign /.`

, a direct link to https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/util/regex/Pattern.html should be added. Also, we need to make sure that the version of Java API linked matches the current version used by the latest annotation processors.

**Motivations**
1. It makes the class stand out more. This will improve readability for Windows users affected by https://github.com/vaadin/docs/issues/1430
2. It improves usability for developers because they can just directly click and be redirected instead of having to look up the **Pattern** class.

dmitrilc:
fixes #1455


dougjewsbury (commented):
AsciiDoc looks good.

tarekoraby:
Is it just me, or does it not look like a link now? (what do you think @jouni?)

image

PR #1469, opened 2022-06-13
Adding special exception for @Route derived route

### LINKED ISSUE #1456: Incorrect statement about @Route's derived route
The statement 

> If you omit the `@Route` annotation parameter, the route target is derived from the class name.

at https://vaadin.com/docs/latest/flow/routing/route

is not applicable to the last example used:

> MainView becomes ""

If the statement was correct, then **MainView** should have become `mainview`.

The java doc for `@Route` at https://vaadin.com/api/platform/23.0.11/com/vaadin/flow/router/Route.html states that 

> MainView or Main names will be mapped to root (value will be "")

dmitrilc:
fix #1456


articles/routing/route.adoc
```diff
@@ -34,7 +34,8 @@ public class SomePathComponent extends Div {
 
 Assuming your application is running from the root context, when the user navigates to `\http://example.com/some/path`, either by clicking a link in the application or entering the address in the address bar, the [classname]`SomePathComponent` component is shown on the page.
 
-If you omit the `@Route` annotation parameter, the route target is derived from the class name.
+If you omit the `@Route` annotation parameter, the route target is derived from the class name. The derived name will be in lower case and trailing *View* will be removed. Also, *MainView* or *Main* names will be mapped to *root* (_value will be_ *""*).
```
tarekoraby, line 37:
```suggestion
If you omit the `@Route` annotation parameter, the route target is derived from the class name.
The derived name will be in lower case and the trailing *View* will be removed.
Also, *MainView* or *Main* names will be mapped to *root* (_value will be_ *""*).
```

tarekoraby (approved):

PR #1480, opened 2022-06-14
Documenting the usage of return type for programmatic navigation

mstahv:
Also added instructions how to maintain URL parameters manually in the view (already valid and more relevant after the new API addition)

https://github.com/vaadin/flow/pull/13906


articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,69 @@ layout: page
...
+NativeButton editButton = new NativeButton(
+        "Edit user details");
+editButton.addClickListener(e ->
+        editButton.getUI().ifPresent(ui -> ui.navigate(
+                UserProfileEdit.class,
+                new RouteParameters("userID", "123")))
+);
+----
+
+[role="since:react-dom@v23.2"]
```
jouni, line 46: Define the Maven artifact ID (we are not interested in the version of the `react-dom` npm dependency), and use an uppercase `V` since that will appear as is in the production site (and we use that instead of lowercase `v`).

```suggestion
[role="since:com.vaadin:vaadin@V23.2"]
```

articles/routing/route-parameters.asciidoc
```diff
@@ -102,6 +102,62 @@ public class WildcardGreeting extends Div
 Note that more specific route always takes precedence over wildcard route.
 Note also that if no value was passed to the wildcard route parameter, then its value will be an empty String.
 
+== Maintaining Route Parameters manually
```
jouni, line 105:
```suggestion
== Maintaining Route Parameters Manually
```

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,69 @@ layout: page
...
+=== Configuring the target view
+
+The navigate methods where you give the target view class as a parameter returns the actual target view as [classname]`Optional`. This allows you to further configure the new view with native Java API. 
+
+The optional may be empty if the target view is not immediately available for some reason. This may happen for example because of security constraints or if the navigation was cancelled or posponed in navigation lifecycle events TODO link.
```
jouni, line 51: Not sure I understand the last part of this paragraph with the `TODO link` mention.

```suggestion
The optional may be empty if the target view is not immediately available for some reason.
This may happen, for example, because of security constraints or if the navigation was cancelled or postponed in navigation lifecycle events `TODO` link.
```

jouni (commented):
I’m not sure why the preview build has failed. Need to try this locally or ask Mikael to check the logs.

articles/routing/navigation.asciidoc
mstahv: Eh. I was wondering that that react-dom ment, just copy-pasting as I have no idea what this docs site does 😂 Still don't understand the version. Is that required to be a full maven version or something else? What is production site? (for further reference, I assume your suggestion is correct).

articles/routing/navigation.asciidoc
mstahv: I'll add a link to different part of docs there, forgot to do that there.

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
 
 = Navigating Between Routes
 
+Switching between routes/views can be initiated in two ways: programmatically using [methodname]`UI.navigate()` methods or using links. The navigate method is more agile for Java programmer, as the view change can be issued from anywhere in you code and you can interact with the target view. Using links on the other hand is the native web way of moving from view to another, allowing you for example to share the direct link to a specific view with your colleague and they work even if session has expired.
```
tarekoraby, line 9:
```suggestion
Switching between routes/views can be initiated in two ways: programmatically using [methodname]`UI.navigate()` methods or using links.
The navigate method is more agile for a Java programmer, as the view change can be issued from anywhere in your code and you can interact with the target view.
Using links, on the other hand, is the native web way of moving from one view to another, allowing you for example to share the direct link to a specific view with your colleague and they work even if the session has expired.
```

tarekoraby, line 9:
```suggestion
Switching between routes/views can be initiated in two ways: programmatically using [methodname]`UI.navigate()` methods or using links.
The navigate method is more agile for a Java programmer, as the view change can be issued from anywhere in your code and you can interact with the target view.
Using links, on the other hand, is the native web way of moving from one view to another, allowing you for example to share the direct link to a specific view with your colleague and they work even if the session has expired.
```

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+Compared to [classname]`String` versions, this avoids having to generate the route string manually.
+
+In the browser, navigate methods triggers a location update and the addition of a new history state entry, but does not issue a full page reload.
+
+*Example*: Navigation to the `company` route target when clicking a button:
+
+[source,java]
+----
+NativeButton button = new NativeButton(
+        "Navigate to company");
```
tarekoraby, line 26:
```suggestion
Button button = new Button("Navigate to company");
```

tarekoraby, line 26:
```suggestion
Button button = new Button("Navigate to company");
```

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+           ui.navigate("company"))
+);
+----
+
+*Example*: Navigation to the `user/123/edit` route target when clicking a button:
+
+[source,java]
+----
+NativeButton editButton = new NativeButton(
+        "Edit user details");
```
tarekoraby, line 38:
```suggestion
Button editButton = new Button("Edit user details");
```

tarekoraby, line 38:
```suggestion
Button editButton = new Button("Edit user details");
```

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+----
+NativeButton button = new NativeButton(
+        "Navigate to company");
+button.addClickListener(e ->
+     button.getUI().ifPresent(ui ->
+           ui.navigate("company"))
+);
+----
+
+*Example*: Navigation to the `user/123/edit` route target when clicking a button:
```
tarekoraby, line 33: Would the example be clearer if we use a simpler URL? How about `user-view/id1234`? If we do that, then also instead of using `UserProfileEdit.class`, we can use the more obvious `UserView.class`.

tarekoraby, line 33: Would the example be clearer if we use a simpler URL? How about `user-view/id1234`? If we do that, then also instead of using `UserProfileEdit.class`, we can use the more obvious `UserView.class`.

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+        editButton.getUI().ifPresent(ui -> ui.navigate(
+                UserProfileEdit.class,
+                new RouteParameters("userID", "123")))
+);
+----
+
+[role="since:com.vaadin:vaadin@V23.2"]
+=== Configuring the target view
+
+The navigate methods where you give the target view class as a parameter returns the actual target view as [classname]`Optional`. This allows you to further configure the new view with native Java API.
```
tarekoraby, line 49: > new view with native Java API
As a reader, I'm not sure what "native" is supposed to mean here.

tarekoraby, line 49: > new view with native Java API
As a reader, I'm not sure what "native" is supposed to mean here.

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+);
+----
+
+[role="since:com.vaadin:vaadin@V23.2"]
+=== Configuring the target view
+
+The navigate methods where you give the target view class as a parameter returns the actual target view as [classname]`Optional`. This allows you to further configure the new view with native Java API. 
+
+The return value may be empty if the target view is not immediately available for some reason.
+This may happen, for example, because of security constraints or if the navigation was cancelled or postponed in <<lifecycle#, navigation lifecycle events>>.
```
tarekoraby, line 52:
```suggestion
This may happen, for example, because of security constraints or if the navigation was canceled or postponed in a <<lifecycle#, navigation lifecycle events>>.
```

tarekoraby, line 52:
```suggestion
This may happen, for example, because of security constraints or if the navigation was canceled or postponed in a <<lifecycle#, navigation lifecycle events>>.
```

articles/routing/route-parameters.asciidoc
```diff
@@ -102,6 +102,62 @@ public class WildcardGreeting extends Div
 Note that more specific route always takes precedence over wildcard route.
 Note also that if no value was passed to the wildcard route parameter, then its value will be an empty String.
 
+== Maintaining Route Parameters Manually
```
tarekoraby, line 105: First, I think this is a separate topic from documenting programmatic navigation. So let's create a separate PR for it. 

That aside, I must say that I read this section three times, and I'm still struggling to understand the solution it's describing or even the problem that it's intended to solve. But as far as I can tell, this is an esoteric use case that should be covered in an "advanced section" somewhere (I don't think it's suitable for people who are just starting to learn about how to add parameters to a Vaadin route). I suggest either creating a new article for that under `routing/additional-guides` or adding it to the History API guide.

tarekoraby (changes requested):

articles/routing/navigation.asciidoc
mstahv: Was about to make other improvement in an other commit, but suits me...

articles/routing/navigation.asciidoc
mstahv: Yes, I had a note about this part as well. Getting to advanced example, when explaining the basics. But I'd favour doing this (and others) change separately, as this patch is related to new functionality and supporting areas.

articles/routing/route-parameters.asciidoc
mstahv: The example could indeed be better. Think of a search view where you want to maintain the deep linking URL in the browser. You don't want to use our somewhat clumsy routing/navigating API to navigate to a new search view with new parameters. Will be much better with this approach (or with query parameters).

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
 = Navigating Between Routes
 
+Switching between routes/views can be initiated in two ways: programmatically using [methodname]`UI.navigate()` methods or using links.
+The navigate method is more agile for a Java programmer, as the view change can be issued from anywhere in your code and you can interact with the target view.
+Using links, on the other hand, is the native web way of moving from one view to another, allowing you for example to share the direct link to a specific view with your colleague and they work even if the session has expired.
+
+
+== Server-Side Navigation
+
+You can trigger navigation from the server side code using various [methodname]`UI.navigate()` methods.
```
dougjewsbury, line 16:
```suggestion
You can trigger navigation from the server-side code using various [methodname]`UI.navigate()` methods.
```

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+
+You should primarily use [methodname]`UI.navigate(Class<? extends Component> navigationTarget)` or [methodname]`navigate(Class<? extends C> navigationTarget, RouteParameters parameters)` methods, where you pass the [classname]`Class` of the target view as a parameter.
+Compared to [classname]`String` versions, this avoids having to generate the route string manually.
+
+In the browser, navigate methods triggers a location update and the addition of a new history state entry, but does not issue a full page reload.
```
dougjewsbury, line 21:
```suggestion
In the browser, the [methodname]`navigate()` method triggers a location update and the addition of a new history state entry, but does not issue a full page reload.
```

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+        editButton.getUI().ifPresent(ui -> ui.navigate(
+                UserProfileEdit.class,
+                new RouteParameters("userID", "123")))
+);
+----
+
+[role="since:com.vaadin:vaadin@V23.2"]
+=== Configuring the target view
+
+The navigate methods where you give the target view class as a parameter returns the actual target view as [classname]`Optional`. This allows you to further configure the new view with its Java API.
```
dougjewsbury, line 49:
```suggestion
The [methodname]`navigate()` method where you give the target view class as a parameter returns the actual target view as [classname]`Optional`.
This allows you to further configure the new view with its Java API. 
```

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,70 @@ layout: page
...
+[source,java]
+----
+new Button("Edit " + user.getName(), event -> {
+    ui.navigate(UserEditor.class)
+            .ifPresent(editor -> editor.editUser(user));
+})
+----
+
+Note, that with the above code Vaadin don't have the data to maintain deep link to edit the very same user. 
+In many use cases this is fine, but if you expect to share links to the target view, the target view must maintain enough details in the URL and add de-serialise the parameters, as described in <<route-parameters#, Route Parameters>>.
```
dougjewsbury, line 66:
```suggestion
Note that, with this code, Vaadin does not have the data to maintain a deep link to edit the very same user. 
In many use cases, this is fine, but if you expect to share links to the target view, the target view must maintain enough details in the URL and deserialize the parameters, as described in <<route-parameters#, Route Parameters>>.
```

articles/routing/route-parameters.asciidoc
```diff
@@ -102,6 +102,62 @@ public class WildcardGreeting extends Div
...
+
+In certain cases, developer might want to maintain the parameters in the browser URL. 
+Like if the edited entity in a form changes via instance method instead of via passing URL parameters.
+In this case developer can use the History object to maintain the RouteParameters.
```
dougjewsbury, line 109:
```suggestion
In certain cases, a developer might want to maintain the parameters in the browser URL, such as if the edited entity in a form changes via an instance method instead of via passing URL parameters.
In this case, the developer can use the [classname]`History` object to maintain the [classname]`RouteParameters`.
```

articles/routing/route-parameters.asciidoc
```diff
@@ -102,6 +102,62 @@ public class WildcardGreeting extends Div
...
+
+*Example*: A view that is updating the deep linking parameters based on a parameter given in a public API. The view also implements HasUrlParameter to de-serialise the parameter in case a user arrives directly to the page:
```
dougjewsbury, line 110:
```suggestion
*Example*: A view that updates the deep linking parameters based on a parameter given in a public API.
The view also implements [interfacename]`HasUrlParameter` to deserialize the parameter in case a user arrives at the page directly:
```

articles/routing/route-parameters.asciidoc
mstahv: Changed the title and added a bit more context why using Router for everything isn't the best way to make the world a better place.

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,21 @@ order: 30
...
 
-The router takes care of serving content when the user navigates within an application.
-It includes support for nested routes, access to route parameters and more.
+The user <<navigation#,navigates between views>> in your application by interacting with UI elements such as buttons, links, and menus.
+Vaadin maintains the proper deep linking URL in your end user's browser when using the framework's built-in mechanisms for navigation.
 
-The user navigates by interacting with UI elements such as buttons and menus.
-When this occurs, the interaction needs to be mapped to a particular route.
+Routing and navigation should not be considered as the primary mechanism to interact with different parts of your UI, but you should simply define pure Java APIs instead.
+In fact, if deep linking is not needed, even a large Vaadin application can ignore routing and navigation for simplicity and only route the entry view to the root URL of the application.
```
tarekoraby, line 22: The message that these two sentences attempt to convey could be clarified more. Do you here mean that user interaction in Vaadin applications is often better achieved by dynamically adding and removing components/views (instead of routing between different views)?

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,21 @@ order: 30
...
+However, Routing and Navigation APIs in Vaadin help you achieve a decent level of deep linking relatively easily. Thus, we suggest routing at least the top-level views to different URLs and then utilizing more advanced features for views where you need a better deep linking.
```
tarekoraby, line 23: > and then utilizing more advanced features for views where you need a better deep linking

It's not clear here what "more advanced features" is referring to.

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,21 @@ order: 30
...
+
+In most non-trivial applications, views also need to interact with each other and share data. When initiating navigation, the previous view can <<navigation#passing-data-using-route-parameters,submit trivial data>> as route or query parameters to the new view. The easiest, safest, and performant way to share data between views is to <<navigation#interacting-directly-with-the-target-view,call their Java API directly>>. Doing this, however, will not automatically maintain the URL in the browser. The actual UI state can also lose parity with the URL shown in the browser by internal interaction within the view or, for example, by a message arriving through another thread. In these cases, you should <<route-parameters#maintaining-route-parameters-without-navigation,manually maintain the URL>> if deep linking is required.
```
tarekoraby, line 25:
```suggestion
In most non-trivial applications, views also need to interact with each other and share data. When initiating navigation, the previous view can <<navigation#passing-data-using-route-parameters,submit trivial data>> as route or query parameters to the new view. The easiest, safest, and performant way to share data between views is to <<navigation#interacting-directly-with-the-target-view,call their Java API directly>>.
```

The caveat that starts with _"Doing this, however, will not automatically ...."_  is many steps ahead (it assumes the reader has already a good understanding of the Java API for navigation). I think it's better to remove this since the same caveat (with the link to `route-parameters#maintaining-route-parameters-without-navigation`) is already mentioned in the linked Java API for navigation.

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,77 @@ layout: page
...
+
+[source,java]
+----
+new Button("Edit " + user.getName(), event -> {
+    ui.navigate(UserEditor.class)
+            .ifPresent(editor -> editor.editUser(user));
+})
+----
+
+Note that, with this code, Vaadin does not have the data to maintain a deep link to edit the very same user.
```
tarekoraby, line 53: > Vaadin does not have the data to maintain a deep link to edit the very same user

Do you mean that "Vaadin would not automatically update the URL to include information about the edited user"? 

Also, does this sentence presuppose that the target view accepts parameters? If so, then this should be stated.

articles/routing/navigation.asciidoc
```diff
@@ -6,12 +6,77 @@ layout: page
...
+new Button("Edit " + user.getName(), event -> {
+    ui.navigate(UserEditor.class)
+            .ifPresent(editor -> editor.editUser(user));
+})
+----
+
+Note that, with this code, Vaadin does not have the data to maintain a deep link to edit the very same user. 
+In many use cases, this is fine, but if you expect to share links to the target view, the target view must maintain enough details in the URL and deserialize the parameters, as described in <<route-parameters#maintaining-route-parameters-without-navigation, Maintaining Route Parameters without Navigation>>, or use Route Parameters to pass the data.
+
+Note, in previous versions, a similar mechanism has been widely used, but the reference of the target view has been fetched using dependency injection, searched from the [classname]`UI` internals, or picked up using application-specific helpers.
```
tarekoraby, line 56: I'm not sure which previous versions are being referred to here. But in any case, is this information useful and actionable for our existing users? If so, then let's mention it in the upgrade guide and keep the main guides optimized for our new users. That is, I think this sentence should be removed from this guide.

articles/routing/navigation.asciidoc
```diff
@@ -63,7 +51,25 @@ new Button("Edit " + user.getName(), event -> {
...
-In many use cases, this is fine, but if you expect to share links to the target view, the target view must maintain enough details in the URL and deserialize the parameters, as described in <<route-parameters#, Route Parameters>>.
+In many use cases, this is fine, but if you expect to share links to the target view, the target view must maintain enough details in the URL and deserialize the parameters, as described in <<route-parameters#maintaining-route-parameters-without-navigation, Maintaining Route Parameters without Navigation>>, or use Route Parameters to pass the data.
+
+Note, in previous versions, a similar mechanism has been widely used, but the reference of the target view has been fetched using dependency injection, searched from the [classname]`UI` internals, or picked up using application-specific helpers.
+
+=== Passing Data using Route Parameters
+
+If your target view implements HasRouteParameters, you can submit trivial data to the target view as RouteParameters. Compared to directly interacting with the target views Java API, this method automatically maintains the URL for deep linking. However, you can't pass more complex objects as route parameters as such.
```
tarekoraby, line 60: > automatically maintains the URL for deep linking

To my mind, the term "deep linking" is not widely used. I would just get rid of it wherever possible and explain it in plain English. Here for example I think it would be clearer to say something like: _Compared to directly interacting with the target views' Java API, this method automatically updates the URL to include the Route Parameter._

jouni: “Deep linking” is fairly common language in my opinion. For example, Apple is using that in their developer-focused communication.

tarekoraby: Let's keep it then.

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
 
 = Routing and Navigation
 
-Vaadin provides the [classname]`Router` class to structure the navigation of your web application into logical parts.
+Views in Vaadin are larger components that present logical parts of your UI.
+They are typically composed of other smaller components and help you structure and re-use your UI code.
```
dougjewsbury, line 12:
```suggestion
They are typically composed of other smaller components and help you structure and reuse your UI code.
```

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
...
-The router takes care of serving content when the user navigates within an application.
-It includes support for nested routes, access to route parameters and more.
+A <<route#,route>> in Vaadin lingo means a view mapped to a specific URL. 
+URLs can also contain <<route-parameters#,Route Parameters>> or <<additional-guides/query-parameters#,Query Parameters>> that views can utilize to configure their internal state further.
+
+The user <<navigation#,navigates between views>> in your application by interacting with UI elements such as buttons, links, and menus.
+Vaadin maintains the proper deep linking URL in your end user's browser when using the framework's built-in mechanisms for navigation.
+
+Routing and navigation is not intended to be the primary mechanism to interact with different parts of your UI or to maintain the state of the application.
+Directly changing the state of components and adding/removing them from the UI using Java API is easier and more efficient.
```
dougjewsbury, line 23:
```suggestion
Directly changing the state of components and adding/removing them from the UI using Java APIs is easier and more efficient.
```

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
...
+In fact, if deep linking is not needed, even a large Vaadin application can ignore routing and navigation APIs for simplicity and only route the entry view to the root URL of the application.
+However, routing and navigation APIs in Vaadin help you achieve a decent level of deep linking relatively easily. 
+Thus, we suggest routing at least the top-level views to different URLs and then utilizing more advanced features like parameters and <<layout#,nested routes>> for views where you need a better deep linking.
```
dougjewsbury, line 26:
```suggestion
Thus, we suggest routing at least the top-level views to different URLs and then utilizing more advanced features like parameters and <<layout#,nested routes>> for views where you need better deep linking.
```

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
...
+
+
+Sometimes a proper deep linking requires developer to manually <<updating-url-parameters#,maintain the URL parameters>>.
```
dougjewsbury, line 29:
```suggestion
Sometimes proper deep linking requires the developer to manually <<updating-url-parameters#,maintain the URL parameters>>.
```

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
...
+For example, the UI state can lose parity with the URL shown in the browser by user interaction within the view, like changing search filters, or by a message arriving through another thread.
+Also, other views can be <<navigation#interacting-directly-with-the-target-view,passing data directly>> using Java API, instead of using URL parameters.
```
dougjewsbury, line 31:
```suggestion
Also, other views can be <<navigation#interacting-directly-with-the-target-view,passing data directly>> using Java APIs, instead of using URL parameters.
```

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
...
+In these cases the routing mechanism can't update the URL automatically.
```
dougjewsbury, line 32:
```suggestion
In these cases, the routing mechanism cannot update the URL automatically.
```

articles/routing/navigation.asciidoc
```diff
@@ -1,17 +1,83 @@
...
+----
+new Button("Edit " + user.getName(), event -> {
+    ui.navigate(UserEditor.class)
+            .ifPresent(editor -> editor.editUser(user));
+})
+----
+
+Note that, with this code, Vaadin does not have the data to maintain a deep link to edit the same entity.
+For the active user of the application, a view to edit the selected entity is opened, but they cannot share a direct link to edit the same entity with their colleague.
+If deep linking is required, the target view must <<updating-url-parameters#,maintain enough details about the UI state in the URL>> or developer must use Route Parameters to pass the data instead of the direct Java API.
```
dougjewsbury, line 55:
```suggestion
If deep linking is required, the target view must <<updating-url-parameters#,maintain enough details about the UI state in the URL>>, or the developer must use RouteParameters to pass the data, instead of the direct Java API.
```

articles/routing/navigation.asciidoc
```diff
@@ -1,17 +1,83 @@
...
+
+=== Passing Data using Route Parameters
+
+If your target view implements HasRouteParameters, you can submit trivial data to the target view as RouteParameters.
```
dougjewsbury, line 59:
```suggestion
If your target view implements [interfacename]`HasRouteParameters`, you can submit trivial data to the target view as [classname]`RouteParameters`. 
```

articles/routing/navigation.asciidoc
```diff
@@ -1,17 +1,83 @@
...
+
+Note that, with this code, Vaadin does not have the data to maintain a deep link to edit the same entity.
+For the active user of the application, a view to edit the selected entity is opened, but they cannot share a direct link to edit the same entity with their colleague.
+If deep linking is required, the target view must <<updating-url-parameters#,maintain enough details about the UI state in the URL>> or developer must use Route Parameters to pass the data instead of the direct Java API.
+
+=== Passing Data using Route Parameters
+
+If your target view implements HasRouteParameters, you can submit trivial data to the target view as RouteParameters. 
+Compared to directly interacting with the target views Java API, this method automatically maintains the URL for deep linking. 
+However, you can't pass more complex objects as route parameters as such and you always need to consider data safety parameters that end up to URLs.
```
dougjewsbury, line 61:
```suggestion
However, you cannot pass more complex objects as route parameters as such, and you always need to consider data safety parameters that end up in URLs.
```

articles/routing/updating-url-parameters.asciidoc
```diff
@@ -0,0 +1,67 @@
+---
+title: Updating URL Parameters without Navigation
+description: To preserve deep linking developer must sometimes maintain the Route parameters and Query parameters manually
```
dougjewsbury, line 3:
```suggestion
description: To preserve deep linking, the developer sometimes needs to maintain the Route parameters and Query parameters manually
```

articles/routing/updating-url-parameters.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+order: 25
+layout: page
+---
+
+== Updating URL Parameters without Navigation
+
+Route parameters, Route templates and Query Parameter are most often defined in a link pointing to a specific view with the desired state and when navigating to view the Router calls the view's [methodname]`setParameter` method.
```
dougjewsbury, line 10:
```suggestion
Route parameters, Route templates and Query parameters are most often defined in a link pointing to a specific view with the desired state, and when navigating to a view, the Router calls the view's [methodname]`setParameter()` method.
```

articles/routing/updating-url-parameters.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+In many cases, the view itself should also maintain the route and query parameters so that the user can share the URL from the browser's location bar meaningfully.
+
+A good example of such a use case could be a search view that maintains the search parameters in the URL but doesn't want to build its UI logic around the Router usage.
```
dougjewsbury, line 13:
```suggestion
A good example of such a use case could be a search view that maintains the search parameters in the URL but does not want to build its UI logic around the Router usage.
```

articles/routing/updating-url-parameters.asciidoc
```diff
@@ -0,0 +1,67 @@
...
+Another case would be passing a more complex object to a view via Java API instead of serializing the object somehow into the URL and passing the value via Router.
```
dougjewsbury, line 14:
```suggestion
Another case would be passing a more complex object to a view via a Java API instead of serializing the object somehow into the URL and passing the value via Router.
```

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
 
 = Routing and Navigation
 
-Vaadin provides the [classname]`Router` class to structure the navigation of your web application into logical parts.
+Views in Vaadin are larger components that present logical parts of your UI.
+They are typically composed of other smaller components and help you structure and re-use your UI code.
+Routing in Vaadin means mapping views into specific URLs and maintaining a meaningful https://en.wikipedia.org/wiki/Deep_linking[deep linking].
```
dougjewsbury, line 13:
```suggestion
Routing in Vaadin means mapping views into specific URLs and maintaining meaningful https://en.wikipedia.org/wiki/Deep_linking[deep linking].
```

articles/routing/index.asciidoc
```diff
@@ -8,13 +8,29 @@ order: 30
...
+Deep linking allows your users (by copy-pasting the current URL from their browser) and other applications to share direct links to a specific view and state in your Vaadin UI.
 
-The router takes care of serving content when the user navigates within an application.
-It includes support for nested routes, access to route parameters and more.
+A <<route#,route>> in Vaadin lingo means a view mapped to a specific URL. 
+URLs can also contain <<route-parameters#,Route Parameters>> or <<additional-guides/query-parameters#,Query Parameters>> that views can utilize to configure their internal state further.
```
dougjewsbury, line 17:
```suggestion
URLs can also contain <<route-parameters#,route parameters>> or <<additional-guides/query-parameters#,query parameters>> that views can utilize to configure their internal state further.
```

mstahv:
@tarekoraby I guess we are ready now with this?

tarekoraby (approved):

PR #1489, opened 2022-06-16
Move Icons under /components

tarekoraby:
Most users probably treat Icons as just another thing (or component) that they can add to the UI. It would be more natural to have icons there.


jouni:
I went ahead and combined the Lumo and Vaadin icons pages together, and moved the icon sizing information to the Lumo size & space page.

jouni (approved):

PR #1492, opened 2022-06-16
Improving routing exceptions doc

dmitrilc:


articles/routing/exceptions.asciidoc
```diff
@@ -39,8 +39,10 @@ public class RouteNotFoundError extends Component
 
 * This returns a `404` HTTP response and displays the text specified in the parameter to [methodname]`setText()`.
 
-The exception matching order is first by exception cause and then by exception super type.
+Exceptions are matched in the order below:
 
+1. By exception cause.
+2. By exception super type.
```
jouni, line 45: No need for explicit number. Asciidoc will generate a `<ol>` from this syntax.
```suggestion
. By exception cause.
. By exception super type.
```

articles/routing/exceptions.asciidoc
```diff
@@ -16,7 +16,7 @@ Exception targets generally work in the same way as regular navigation targets,
 Errors in navigation are resolved to a target that is based on the exception type thrown during navigation.
 
 At startup, all classes implementing the [interfacename]`HasErrorParameter<T extends Exception>` interface are collected for use as exception targets during navigation.
-Example classes include [classname]`RouteNotFoundError` for [classname]`NotFoundException`.
+Example classes from Vaadin include [classname]`RouteNotFoundError` for [classname]`NotFoundException`.
```
tarekoraby, line 19:
```suggestion
An example of such a class is the [classname]`RouteNotFoundError`.
This class is included by default in the framework and is used to resolve errors related to the router's `NotFoundException`.
```

tarekoraby (approved):

PR #1500, opened 2022-06-20
Remove “since” class for old Vaadin versions

tarekoraby:
There is no need to have "since" notice for old Vaadin versions like this one: 

image


jouni:
The original idea with the versioning badges was that we would not need to remove them, and that they’d “fade away” once they are no longer relevant for users.

Initially, there was some logic that removed/hid them after a couple of major versions, but that got changed at some point to keep them visible forever, just with reduced visual weight.

Here’s the relevant piece of code in DSP: https://github.com/vaadin/docs-app/blob/master/src-node/modules.js#L120-L137

I’m not sure we can figure out a heuristic when to completely hide “since” badges. The best I can think of is “after two major versions”. Or alternatively, make these even more subtle, some kind of meta data of the page (similar to “last updated”).

jouni:
I think we should think about how we continue using these badges with our new release model, after each minor version update and what happens when a new major is released. Perhaps part of a bigger plan how docs versions are managed.

tarekoraby:
For branch Latest. which is today V23, I feel that we should keep all those "since" tags that could be relevant for people who could be on 23.0.0.

PR #1504, opened 2022-06-20
Update dynamic.asciidoc

dmitrilc:
Fixes errors and improve readability

### DIFF
--- a/articles/routing/dynamic.asciidoc
+++ b/articles/routing/dynamic.asciidoc
@@ -7,41 +7,51 @@ layout: page
 
 = Registering Routes Dynamically
 
-In addition to registering routes and route templates using the <<index#,`@Route` annotation>>, you can add and remove routes dynamically at runtime.
+In addition to registering routes and route templates using the <<index#, `@Route`>> annotation, you can add and remove routes dynamically at runtime. 
 This is useful, for example, when a route should be added or removed based on changed business data or application configuration at startup.
 
-The [classname]`RouteConfiguration` class allows you to configure routes to a specific scope.
-You can configure routes to:
+The [classname]`RouteConfiguration` class can be used to limit route access to:
 
 * all users using the *application* scope, or
 * only certain active users using the *session* scope.
 
-You can access the scope using the [methodname]`forSessionScope()` and [methodname]`forApplicationScope()` static methods.
-All components with an `@Route` annotation are added to the application scope.
+You can access the scope using the static methods [methodname]`forSessionScope()` and [methodname]`forApplicationScope()` from the [classname]`RouteConfiguration` class.
+
+[NOTE]
+All components annotated with `@Route` are added to the application scope at startup, unless the <<Dynamic Registration of @Route Annotated Classes, `registerAtStartup`>> element has been set to `false`.
 
 == Configuring User-Specific Routes
 
 You can add and remove routes for certain users; for example, based on their access rights.
 
-*Example*: Adding two views for currently active users.
+*Example*: Adding *admin* and *home* views for users with an active session.
 
 [source,java]
 ----
-RouteConfiguration.forSessionScope().setRoute("admin",
-        AdminView.class);
+RouteConfiguration.forSessionScope()
+        .setRoute(
+                "admin", //path
+                AdminView.class //navigation target
+         );
 
 // parent layouts can be given as a vargargs parameter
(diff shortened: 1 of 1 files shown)


articles/routing/dynamic.asciidoc
```diff
@@ -7,41 +7,51 @@ layout: page
 
 = Registering Routes Dynamically
 
-In addition to registering routes and route templates using the <<index#,`@Route` annotation>>, you can add and remove routes dynamically at runtime.
+In addition to registering routes and route templates using the <<index#, `@Route`>> annotation, you can add and remove routes dynamically at runtime.
```
tarekoraby, line 10:
```suggestion
In addition to registering routes and route templates using the <<index#, `@Route`>> annotation, you can add and remove routes dynamically at runtime.
```

articles/routing/dynamic.asciidoc
```diff
@@ -7,41 +7,51 @@ layout: page
...
+        .setRoute(
+                "home", //path
+                HomeView.class, //navigation target
+                MainLayout.class //one or more parents
+        );
 ----
 
-A route set for the user can override a route with the same path from the application scope.
-This means that any statically registered `@Route` can be overridden for a specific user, if necessary.
+Any registered `@Route` in the application scope can be overridden for a specific user in the session scope.
```
tarekoraby, line 46: I feel that this suggestion hides a couple of important terms in the original formulation. First, in the original sentence, it was clear that we are speaking about "routes with the same path". Also, the original formulation implies that an application-scoped route could be defined dynamically or through the `@Route` annotation (whereas in the reformulated suggestion, only the `@Route` way is mentioned).

articles/routing/dynamic.asciidoc
```diff
@@ -65,6 +75,13 @@ configuration.removeRoute(AdminView.class);
 configuration.removeRoute("admin");
 ----
 
+[NOTE]
+Removing a route in the session scope that had previously overridden a route in the application scope makes the application-scoped route accessible once again.
+
+[NOTE]
+When dynamically registering a route, any annotations on classes *are ignored*, except when the method used contains `Annotated`; for example, [methodname]`setAnnotatedRoute()`.
+See <<Dynamic Registration of @Route Annotated Classes>> below for more.
```
dougjewsbury, line 83:
```suggestion
See <<Dynamic Registration of @Route Annotated Classes>> later in this article for more.
```

articles/routing/dynamic.asciidoc
```diff
@@ -7,41 +7,51 @@ layout: page
...
-This means that any statically registered `@Route` can be overridden for a specific user, if necessary.
+Any registered `@Route` in the application scope can be overridden for a specific user in the session scope.
 
 The routes in the session scope are accessible for the current user only for as
 long as the session is valid.
-When the session is invalidated by the user logging out, the session-scoped routes are no longer available automatically.
-It is not necessary to specifically remove these routes.
+
+When the session is invalidated by the user logging out, the session-scoped routes are no longer available.
+It is unnecessary to manually remove these routes.
```
tarekoraby, line 52:
```suggestion
When the session is invalidated by the user logging out, the session-scoped routes are no longer available.
It is unnecessary to manually remove these routes.
```

articles/routing/dynamic.asciidoc
dmitrilc: Hi Tarek, thanks for the comment.

The problem with the original text was that it felt like reading the same sentence repeated twice.

> First, in the original sentence, it was clear that we are speaking about "routes with the same path".

I felt that "_routes with the same path_" is redundant when the word "_override_" was already being used. Whenever I read "_override_", I would automatically assume that you must be overriding something that is the same from the parent or wider scope. Including "_with the same path_" is also correct, albeit a little redundant.

For example:
In the sentence, "_Please override the method `public void doSomething()` with the same method name_", the fragment "_with the same method name_" is redundant because the overriding method must already have the same name.

> Also, the original formulation implies that an application-scoped route could be defined dynamically or through the `@Route` annotation

The original sentence said "This means that any **statically** registered `@Route`", so it definitely didn't read like both `dynamic` **and** `static`. 

> whereas in the reformulated suggestion, only the `@Route` way is mentioned

The reformulated sentence actually omitted both `dynamic` and `static`, so it can cover both cases.

tarekoraby (approved):

PR #1520, opened 2022-06-27
Add "Testing" as a top-level topic of navigation

### LINKED ISSUE #1466: Add "Testing" as a top-level topic of navigation
Since this is one of the most common tasks, testing should have its own top-level topic in the navigation. Such topic should probably include TestBench as well an alternative tutorial that doesn't require paid subscription to run.

tarekoraby:
Fix #1466


jouni (commented):
I would, again, want to remove the explicit **Overview** page and move that content to the index page. Probably needs some content rewrite, and maybe a new page for the “A Typical Test” section (or maybe that could just be part of “Getting Started”).

This notice at the end of the Overview-page is not needed:
Screen Shot 2022-06-28 at 11 40 30
 
Remove the “Product page” link, it is IMO not needed. Same for the example project, it is already linked from the “Getting Started” page. Move the API links into the index page `page-links`, instead of having them in the sidebar nav:
Screen Shot 2022-06-28 at 11 41 07

jouni:
This is out of scope of this PR, but just a note: the amount of admonitions on most of these docs pages is quite excessive, and would benefit from cleanup. Also, the page titles could perhaps be simplified a bit, so that they are not so long in the sidebar nav.

tarekoraby:
Done.

> This is out of scope of this PR, but just a note: the amount of admonitions on most of these docs pages is quite excessive, and would benefit from cleanup. Also, the page titles could perhaps be simplified a bit, so that they are not so long in the sidebar nav.

I think we can look into doing some major cleanup in connection with the upcoming reorganization of TestBench material into two equal approaches:

- End-to-end testing
- UI unit testing

tarekoraby:
Is this one ready to be merged?

jouni:
Sorry for the delay. I’ll make one minor update, then it’s good to go IMO.

jouni (approved):

PR #1537, opened 2022-07-04
Create Selenium guide

dmitrilc:


dmitrilc:
Here is my example test https://github.com/dmitrilc/vaadin-selenium-example

articles/testing/selenium.adoc
```diff
@@ -0,0 +1,272 @@
...
+----
+<1> If you are using `WebDriverManager`, you will need to setup your driver for the test.
+<2> Instantiates the [classname]`ChromeDriver`, which will be used to interact with the Chrome web browser.
+<3> Directs the [classname]`ChromeDriver` to load the URL.
+<4> Because Flow produces a single-page frontend, Selenium is not aware of DOM manipulation after the initial page has been loaded, so we had to use Selenium's https://www.selenium.dev/documentation/webdriver/waits/#fluentwait[FluentWait] API to wait until the DOM has been _compiled_.
+<5> After we are sure that the page title is available, we retrieve the page title.
+<6> We check whether the page title is "Hello World".
+<7> It is important that we close the browser session after the test.
+
+== Verify the URL After Navigation Action
```
tarekoraby, line 116: To me, the most important point that I, as the reader, have learned from the test included in this section is that I can:
. Find elements on the page 
. and, interact with those elements, say, via clicking

I think the title should reflect the most important aspects of this section (which, to my mind, is not the very specific use case of verifying the URL after navigation)

articles/testing/selenium.adoc
```diff
@@ -0,0 +1,272 @@
...
+  assertEquals("http://localhost:8080/about", url);
+  
+  //Ends the browser session
+  driver.quit();
+}
+----
+<1> We wait for the "About" page to load first before attempting to get the URL. This reduces flakiness.
+<2> We use the convenient method to get the full current URL.
+
+== Add User to Master Detail Template
```
tarekoraby, line 162: Here again, the title (and the first couple of sentences) of this section could be rewritten to clarify the most important learning points that people will gain after reading this section. Only a very small subset of developers will have a very specific interest in creating a test that adds a user to a master-detail template that is generated from start.vaadin.com. However, many more readers would benefit from learning how to implement a more advanced testing scenario that involves finding and interacting with different elements.

I suggest, thus, renaming this section's title to something like "Advanced Testing Scenario" (and mention in the first couple of sentences in the section that this example demonstrates a more complex use case that involves using APIs like `By.id` and `By.xpath`.

articles/testing/selenium.adoc
```diff
@@ -0,0 +1,272 @@
...
+  
+  //Ends the browser session
+  driver.quit();
+}
+----
+<1> Finds the first field in the form to input First Name.
+<2> We must click on the field to simulate real behavior of an end user.
+<3> Sending key strokes to this field.
+<4> We use xpath expression to find the element.
+<5> Finally, we test whether all of the information in the table cells match our original data.
```
tarekoraby, line 272: Let's finish this page with a link to Selenium docs so that people can read more if they want to.

tarekoraby (changes requested):

articles/testing/selenium.adoc
```diff
@@ -0,0 +1,272 @@
...
+. Add the WebDriverManager dependency to your project’s `pom.xml` file.
++
+.`pom.xml`
+[source,xml]
+----
+<!-- Web Driver Management Software -->
+<dependency>
+   <groupId>io.github.bonigarcia</groupId>
+   <artifactId>webdrivermanager</artifactId>
+   <version>5.1.1</version>
```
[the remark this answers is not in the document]
tarekoraby: Also the latest version is `5.2.1`: https://search.maven.org/artifact/io.github.bonigarcia/webdrivermanager. Let's use that.

articles/testing/selenium-testing.adoc
```diff
@@ -0,0 +1,350 @@
...
+== Advanced Selenium Test
+
+The test below demonstrates how to use advanced [classname]`By` matchers such as `id()` and `xpath()`.
+This test assumes a Master-Detail view of the kind that could be generated from https://start.vaadin.com/[Vaadin Start].
+
+.`ExampleIT.java`
+[source,java]
+----
+@Test
+public void addUser(){
```
tarekoraby, line 246: @dmitrilc this test doesn't pass in the repo you created. Can you please check?

dmitrilc: @tarekoraby I was able to reproduce it on an old Windows machine with the error below.

```
ElementClickIntercepted element click intercepted

element click intercepted: Element <vaadin-grid-sorter path="col3" aria-label="Sort by ...">Phone</vaadin-grid-sorter> is not clickable at point (788, 83). Other element would receive 
the click: <div class="editor">...</div>
```

I was not able to reproduce it on my MBP M1 16" screen because the field `Phone` is still in view, so Selenium was able to click on it. On the 15" 1080p Windows laptop screen though, the field `Phone` was not in view, so Selenium couldn't click on it.

Because screen size, resolution, and scaling settings can affect the tests, I will make the browser maximized just for that test, which should reduce flakiness.

dmitrilc: Can you update your project from my repo and run the test again? Thanks.

articles/testing/selenium-testing.adoc
```diff
@@ -0,0 +1,353 @@
...
+
+To use Selenium in a Vaadin app, you need to perform the following steps:
+
+. Add the Selenium library dependency to your project's `pom.xml` file.
++
+.`pom.xml`
+[source,xml]
+----
+<!-- Selenium Library -->
+<dependency>
```
alvarezguille, line 41: Latest version now is 4.4.
But for start (spring-boot) projects the version shouldn't be explicit as it's defined in `selenium.version` property in pom, that is used in the bom of spring-boot-starter-parent.
If we add 
```
<dependency>
    <groupId>org.seleniumhq.selenium</groupId>
    <artifactId>selenium-java</artifactId>
    <version>4.3.0</version>
</dependency>
```

We end up with inconsistent versions in dependency:tree which might lead to issues in the future

>[INFO] +- org.seleniumhq.selenium:selenium-java:jar:4.3.0:compile
[INFO] |  +- org.seleniumhq.selenium:selenium-api:jar:4.2.1:compile
[INFO] |  +- org.seleniumhq.selenium:selenium-chrome-driver:jar:4.2.1:compile

User should add the following instead
```
<dependency>
    <groupId>org.seleniumhq.selenium</groupId>
    <artifactId>selenium-java</artifactId>
</dependency>
```

articles/testing/selenium-testing.adoc
```diff
@@ -0,0 +1,353 @@
...
+
+public class ExampleIT {
+
+   @Test
+   public void checkTitle() {
+       WebDriverManager.chromedriver().setup(); // <1>
+       var driver = new ChromeDriver(); // <2>
+       driver.get("http://localhost:8080/hello"); // <3>
+       
+       new FluentWait<WebDriver>(driver)
```
alvarezguille, line 118: Steps 4-7 can be shortened to
```
    new WebDriverWait(driver, Duration.ofSeconds(30), Duration.ofSeconds(1))
        .until(ExpectedConditions.titleIs("Hello World"));
```
I would use the title to make sure that it loaded but make the test about asserting something different.
WebDriverWait is a specialized type of FluentWait that makes things simpler, especially with combined with ExpectedConditions

articles/testing/selenium-testing.adoc
```diff
@@ -0,0 +1,353 @@
...
+
+== Create Your First Selenium Test
+
+As an example, the code snippet below tests whether the page title is “Hello World”. 
+
+.`ExampleIT.java`
+[source,java]
+----
+import io.github.bonigarcia.wdm.WebDriverManager;
+import org.junit.Test;
```
alvarezguille, line 101: >import org.junit.Test;
import org.junit.jupiter.api.Assertions;

Mixing junit 4 and junit 5 imports can result in unexpected outcome (no test run, no assert executed, ...)

articles/testing/selenium-testing.adoc
```diff
@@ -0,0 +1,353 @@
...
+
+  //Have to explicitly wait because it takes time for compiled html to load
+  new FluentWait<WebDriver>(driver)
+          .withTimeout(Duration.ofSeconds(30))
+          .pollingEvery(Duration.ofSeconds(1))
+          .ignoring(NoSuchElementException.class)
+          .until(_driver -> !_driver.getTitle().isEmpty());
+
+  //Clicks on the About button
+  driver.findElement(By.linkText("About")) // <1>
```
alvarezguille, line 213: `driver.findElement(By.linkText("About"))` is not compatible with the new AppNavItem used in start projects, this selector failed when using the the guide in a new project
>org.openqa.selenium.NoSuchElementException: no such element: Unable to locate element: {"method":"link text","selector":"About"}

articles/testing/selenium-testing.adoc
```diff
@@ -0,0 +1,353 @@
...
+          .ignoring(NoSuchElementException.class)
+          .until(_driver -> driver.getTitle().equals("About")); // <3>
+
+  var url = driver.getCurrentUrl(); // <4>
+
+  //Checks whether the url matches
+  assertEquals("http://localhost:8080/about", url);
+  
+  //Ends the browser session
+  driver.quit();
```
alvarezguille, line 228: Having `driver.quit();` as part of the test flow will leave browser windows open in case that a test fail. 
This could be added to a finally clause. Or ideally handle the creation and end of the session in setup and teardown methods that are executed before and after each test.

alvarezguille (changes requested):
Went through the guide and tested it locally, in general it looks good, I still added some comments that I hope are helpful.
Some of them might be a bit opinionated and not require a change like the `WebDriverWait` one.
But others should be addressed so that the guide stays valid.

articles/testing/selenium-testing.adoc
tarekoraby: Yes, it works now with the screen resize. Thanks!

articles/testing/selenium-testing.adoc
dmitrilc: This is an interesting case because you again readability on the `until` condition and gets to skip the `ignoring` call, but loses the information on which duration is timeout and which is polling.

This method still comes out with one extra benefit, so I will be switching to it. Thanks for bringing it up.

I will see what I can do about asserting something else as well.

articles/testing/selenium-testing.adoc
dmitrilc: I somehow prefer `try/finally` over `before/after` for the doc because the tests can appear as individual examples. And we are not really sharing the whole repo with the readers anyway. Though in real code, I would use before/after.

dmitrilc:
@tarekoraby I have resolved all of Alvarez's concerns. You can update your repo and see whether all is good on your side.

articles/testing/selenium-testing.adoc
```diff
@@ -41,11 +41,10 @@ To use Selenium in a Vaadin app, you need to perform the following steps:
 <dependency>
    <groupId>org.seleniumhq.selenium</groupId>
    <artifactId>selenium-java</artifactId>
-   <version>4.3.0</version>
 </dependency>
 ----
 +
-The latest version of Selenium can be found https://search.maven.org/artifact/org.seleniumhq.selenium/selenium-java[here].
+We can skip the version number here because it has already been declared by the `<selenium.version>` property in the `pom.xml` file.
```
tarekoraby, line 47:
```suggestion
In a https://start.vaadin.com/[Vaadin Start] generated project, we can skip the version number here because it has already been declared by the `<selenium.version>` property in the `pom.xml` file.
You can find the latest version of Selenium https://search.maven.org/artifact/org.seleniumhq.selenium/selenium-java[here].
```

We should clarify that the version number is not needed in a Start-generated project because not all projects are generated from Vaadin Start. It would also still be useful to include a link to the latest version (even for a Start generated project) so that people can update it if needed.

articles/testing/selenium-testing.adoc
```diff
@@ -234,12 +239,12 @@ public void routeSwitch(){
 This reduces flakiness.
 <4> We use the convenient method to get the full current URL.
 
-== Advanced Selenium Test
+== Long Selenium Test
```
tarekoraby, line 242: Why "Long"? Why would the user be interested in reading something that is long? What will they gain?

articles/testing/selenium-testing.adoc
```diff
@@ -0,0 +1,354 @@
...
+<3> We wait for the "About" page to load first before attempting to get the URL.
+This reduces flakiness.
+<4> We use the convenient method to get the full current URL.
+
+== Long Selenium Test
+
+The test below demonstrates how a long Selenium test might look like.
+This test assumes a Master-Detail view of the kind that could be generated from https://start.vaadin.com/[Vaadin Start].
+
+.`SimpleTestIT.java`
```
tarekoraby, line 247: Would it be better to remove the word "Test" from the class name since it is implicit in the "IT" posfix?

tarekoraby (approved):

alvarezguille (approved):
Content looks good 👍

PR #1558, opened 2022-07-13
docs: document the need to set display CSS class

web-padawan:
Fixes https://github.com/vaadin/flow-components/issues/3034

Screenshot 2022-07-13 at 14 59 05


articles/components/ds-resources/foundation/utility-classes.asciidoc
```diff
@@ -418,6 +418,10 @@ pass:[<!-- vale Vale.Spelling = YES -->]
 === Flexbox & Grid
 Classes for flexbox and grid layouts.
 
+.Remember to set `display`
+[NOTE,role=small]
+The following classes require setting appropriate <<#display,`Display`>> CSS class on the element.
```
sissbruecker, line 423:
```suggestion
The following classes require setting either the `flex` or `grid` CSS class on the element in order to work. See <<#display,Display classes>>.
```
Maybe we should explicitly mention `flex` and `grid` instead

PR #1580, opened 2022-08-02
Fix passing data via Route Parameters example

tarekoraby:
The [Passing Data Using Route Parameters](https://vaadin.com/docs/latest/routing/navigation/#passing-data-using-route-parameters) section says that:
"_If your target view implements `HasRouteParameters`, you can submit trivial data to the target view as route parameters_"

However, the code example below uses `UI.navigate(Class<? extends Component> navigationTarget, RouteParameters parameters)` in which the `RouteParameters` argument is actually a [Route Template](https://vaadin.com/docs/latest/routing/additional-guides/route-templates) (not a target implementing `HasRouteParameters` as the description says).


mstahv:
What is the fix here? Using more trivial example instead of having "route parameter" in the middle of the url?

Lower on the same page there is also one such example. At least there we should maybe have the more trivial one first and then the "parameter in the middle" case. Or move all that to "route templates" part 🤔

Di you remove "However, you cannot pass more complex objects as route parameters as such, and you always need to consider data safety parameters that end up in URLs." on purpose or do you consider that to be obvious for all users? I wouldn't be surprised to see people trying to use POJOs or even more complex objects as parameters...

tarekoraby:
@mstahv I updated the PR description to clarify the fix. 

> Di you remove "However, you cannot pass more complex objects as route parameters as such, and you always need to consider data safety parameters that end up in URLs." on purpose or do you consider that to be obvious for all users?

Now I don't remember why I removed that 🤔 I'll revert this removal.

mshabarov (approved):

mstahv (approved):

Issue #1587, opened 2022-08-04
Add information about new RouterLink constructor methods


MarcinVaadin:
vaadin/docs follow up of https://github.com/vaadin/flow/issues/14143

RouterLink has new constructor methods introduced which can cause ambiguous compilation errors in specific cases after moving to flow 23.2.

Additional information on docs and release notes should be presented to the users.

PR #1588, opened 2022-08-04
docs: added note about new RouterLink constructors (#1587)

MarcinVaadin:
Documentation update related to https://github.com/vaadin/flow/issues/14143


articles/routing/navigation.asciidoc
```diff
@@ -142,6 +142,10 @@ public class GreetingComponent extends Div
 }
 ----
 
+[NOTE]
+Since Flow 23.2 new constructor signatures have been introduced. Using ambiguous constructor calls will cause build errors. For example `new RouterLink(null, viewClass)` can refer to both `public RouterLink(String text, Class<? extends Component> navigationTarget)` and `public RouterLink(Router router,
+Class<? extends Component> navigationTarget)`.
+
```
jouni, line 148: Always define a title for admonitions. It’s usually quite easy by promoting the most important information from the admonition to be the title.

Break sentences on new lines. This makes editing slightly easier (to move lines) and also more easily indicates if you have really long sentences (which are usually harder for the reader to understand).

I think a `CAUTION` admonition is in order (vs a more discrete `NOTE`), as this warns about build errors.

Avoid version specific text if possible, to keep the documentation texts “evergreen”.

```suggestion
.Using ambiguous constructor calls will cause build errors
[CAUTION]
For example, [methodname]`new RouterLink(null, viewClass)` can refer to both [methodname]`public RouterLink(String text, Class<? extends Component> navigationTarget)` and [methodname]`public RouterLink(Router router,
Class<? extends Component> navigationTarget)`.

```

MarcinVaadin: Yes, thanks for remarks, suggestion committed.

articles/routing/navigation.asciidoc
```diff
@@ -142,6 +142,11 @@ public class GreetingComponent extends Div
 }
 ----
 
+.Using ambiguous constructor calls will cause build errors
```
mshabarov, line 145: The idea behind implemented feature was to provide a constructor without text.
Thus, I propose to just mention this possibility here in the docs: developers don't need to call `RouterLink(String text, Class<? extends Component> navigationTarget)` and pass `""` or `null` there, but instead call the constructor with only class as a parameter. I don't think we have to tell readers about ambiguous constructors.

mshabarov (changes requested):

articles/routing/navigation.asciidoc
MarcinVaadin: > I don't think we have to tell readers about ambiguous constructors. 

so maybe that docs change could be dropped?

mshabarov: Yes.
Only mention that the text parameter is optional and there is a constructor available without it.

MarcinVaadin: How about:
image

MarcinVaadin:
Update simplified to:
Screenshot 2022-08-09 at 08 24 32

articles/routing/navigation.asciidoc
```diff
@@ -142,6 +142,9 @@ public class GreetingComponent extends Div
 }
 ----
 
+[NOTE]
+Create [classname]`RouterLink` without text using [methodname]`new RouterLink(viewClass)`.
```
mshabarov, line 146:
```suggestion
Use [methodname]`new RouterLink(viewClass)` if you want to embed other elements into [classname]`RouterLink` without text, for example images or icons.
```

mshabarov (approved):

mshabarov:
Need to resolve conflicts with master

articles/routing/navigation.asciidoc
```diff
@@ -142,10 +142,8 @@ public class GreetingComponent extends Div
 }
 ----
 
-.Using ambiguous constructor calls will cause build errors
-[CAUTION]
-For example, [methodname]`new RouterLink(null, viewClass)` can refer to both [methodname]`public RouterLink(String text, Class<? extends Component> navigationTarget)` and [methodname]`public RouterLink(Router router,
-Class<? extends Component> navigationTarget)`.
+[NOTE]
+Create [classname]`RouterLink` without text using [methodname]`new RouterLink(viewClass)`.
```
jouni: I don't think there’s a need to use an admonition here. The sentence works just fine as part of the regular flow of content.

```suggestion
You can create a [classname]`RouterLink` without providing an explicit label using the [methodname]`new RouterLink(viewClass)` method.
```

articles/routing/navigation.asciidoc
```diff
@@ -142,6 +142,9 @@ public class GreetingComponent extends Div
 }
 ----
 
+[NOTE]
+Use [methodname]`new RouterLink(viewClass)` if you want to embed other elements into [classname]`RouterLink` without text, for example images or icons.
```
tarekoraby, line 146: We need to indicate to the reader that this feature is only available from V23.2. @jouni is there a way to attach the since tag to the note?

(Btw, I wonder if this is better as a "TIP", not a "NOTE").

tarekoraby (changes requested):

articles/routing/navigation.asciidoc
jouni: > is there a way to attach the since tag to the note?

`[since:com.vaadin:vaadin@V23.2]#Some sentence here that only applies for V23.2#`

jouni: > (Btw, I wonder if this is better as a "TIP", not a "NOTE").

Like I posted before, I don't think this needs to be an admonition at all.

tarekoraby, line 146:
```suggestion
[since:com.vaadin:vaadin@V23.2]#Use [methodname]`new RouterLink(viewClass)` if you want to embed other elements into [classname]`RouterLink` without text, for example images or icons.#

*Example*: Using [classname]`RouterLink` with an icon instead of text.
[source,java]
----
void buildMenu() {
    Icon vaadinIcon = new Icon(VaadinIcon.HOME);
    RouterLink link = new RouterLink(HomeView.class);
    link.add(vaadinIcon);
    menu.add(link);
}

@Route(value = "")
public class HomeView extends Component {
}
----
```

tarekoraby: Agreed.

articles/routing/navigation.asciidoc
tarekoraby: I like my suggestion below better as it includes an example and the since tag.

mshabarov (approved):

articles/routing/navigation.asciidoc
MarcinVaadin: Second review comments applied

articles/routing/navigation.asciidoc
```diff
@@ -160,17 +176,6 @@ Anchor anchor = new Anchor("/hello", "Go to /hello route");
...
-This means that unlike in earlier versions, clicking on a standard link does not cause a full page reload to happen by default.
-If you want a full page reload to happen, for example when navigating to a page that is not implemented using Vaadin, you can add `router-ignore` <<../create-ui/element-api/properties-attributes/#about-attributes, attribute>>; for example, `<a router-ignore href="company">Go to the company page</a>`.
-This can be done from the Java API as follows:
-
-[source, java]
-----
-Anchor anchor = new Anchor("/hello", "Go to /hello route");
-anchor.getElement().setAttribute("router-ignore", "");
-----
-
```
tarekoraby, line 173: @MarcinVaadin @mshabarov why were these lines removed?

MarcinVaadin: There was a conflict during merge, maybe I've resolved it wrongly, let me check.

mshabarov: When I was checked, they were removed and moved few lines below...Let's fix it in another PR

MarcinVaadin: Revert PR created https://github.com/vaadin/docs/pull/1609

PR #1604, opened 2022-08-09
Fix typos and clarifying lazy data binding

dmitrilc:
This PR attempts to improve clarity of the lazy data binding documentation.


dmitrilc:
This is a work in progress.

dmitrilc:
Next step: explains how the component creates the Query and which method does it call when it needs more data

articles/binding-data/data-provider.asciidoc
```diff
@@ -28,8 +28,8 @@ grid.setItems(
 ----
 
 All listing components in Vaadin have a number of overloaded [methodname]`setItems()` methods to define the items to display.
-Items can be basic objects, such as strings or numbers, or plain old Java objects (POJOs), such as data transfer objects (DTOs) or Java Persistence API (JPA) entities, from your domain model.
-The easiest way is to provide a [classname]`List` of objects to be shown in such a component.
+Items can be basic objects, such as strings, numbers, plain old Java objects (POJOs), data transfer objects (DTOs) or Java Persistence API (JPA) entities.
```
tarekoraby, line 31: I don't think that this reformulation is correct. The original text was more accurate as it says that items can be either:
- Basic objects (such as string or numbers)
- Or, POJOs (such as DTOs and JPA entities)

With the proposed reformulation, the text will (wrongly in my opinion) say that items can only be _basic objects_, and that examples of basic objects include:
- strings
- numbers
- POJOs (which to my mind, are not _basic objects_)
- DTOs (also not basic objects, and they are a subset POJOs)
- JPA entities (also better described as POJOs, not basic objects).

articles/binding-data/data-provider.asciidoc
```diff
@@ -143,15 +144,20 @@ For example, to bind data lazily to a `Grid`:
...
 ----
 <1> To create a lazy binding, use an overloaded version of the [methodname]`setItems()` method that uses a callback instead of passing data directly to the component.
 <2> Typically, you call your service layer from the callback, as is done here.
-<3> Use the query object's parameters to limit the data you pass from the back end to the component.
-<4> The callbacks return the data as a [classname]`Stream`.
-In this example, the back end returns a [classname]`List`, so we need to convert it to a [classname]`Stream`.
+<3> Use the query object's properties to limit the data you pass from the back end to the component.
+<4> The link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html#getOffset()[_offset_] refers to the first index of the item to fetch.
+<5> The link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html#getLimit()[_limit_] refers to the number of items to fetch.
+<6> The properties of a [classname]`Query` object are determined by the component when it creates the [classname]`Query` object.
```
tarekoraby, line 159: As these four points are very much related, I suggest combining them into just one point with four sentences.

articles/binding-data/data-provider.asciidoc
```diff
@@ -143,15 +144,20 @@ For example, to bind data lazily to a `Grid`:
...
+<7> The back end returns a [classname]`List`, so we need to convert it to a [classname]`Stream`.
```
tarekoraby, line 160:
```suggestion
<7> In this example, it is assumed that the back end returns a [classname]`List`, so we need to convert it to a [classname]`Stream`.
```

Backends do not always have to return `List`. Our language should clarify that this is just an example.

tarekoraby (changes requested):

articles/binding-data/data-provider.asciidoc
dmitrilc: You have a good point, and your comment just gave me an idea. I will change it to

> Items can be basic objects (such as string or numbers), or POJOs (such as DTOs and JPA entities).

instead.

articles/binding-data/data-provider.asciidoc
dmitrilc: One point with four sentences would be too hard to follow. I prefer them as separate points instead.

tarekoraby: If we don't combine the four points, then we have to do something about callouts number 3 and 6. Callout 3 is placed next to `fetchPersons(` but it's not really describing anything specific about this method. Callout 6 is weirdly placed next to the closing bracket.

Let's also keep in mind that we don't have to mention every detail in the callouts. It is also okay to mention some advanced notes and caveats in the text after/before the code snippet.

dmitrilc: I see your points now. 

I intended callout 3 to signify to developers that they should only fetch what they need. Perhaps the language was not obvious enough. What do you think of the version below?

```suggestion
<3> When fetching for more data, you should utilize the [classname]`Query`'s properties to litmit the amount of data to fetch.
<4> The link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html#getOffset()[_offset_] refers to the first index of the item to fetch.
<5> The link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html#getLimit()[_limit_] refers to the number of items to fetch.
```

You have a good point about callout 6. What do you think about rephrasing it a little and moving it to right after the sentence
> When more data is needed, the component passes a link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html[[classname]`Query`] object as a parameter to your callback methods, where you can check what part of the data set is needed.

This is how it would look like after
> When more data is needed, the component passes a link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html[[classname]`Query`] object as a parameter to your callback methods, where you can check what part of the data set is needed. 
The component will set the properties of the [classname]`Query` when it creates the [classname]`Query` object.

tarekoraby: > <3> When fetching for more data, you should utilize the [classname]`Query`'s properties to litmit the amount of data to fetch.

I like the wording as it clarifies the meaning, but then isn't this sentence better placed next to ` query.getLimit()` (which is currently point number 5)?

> You have a good point about callout 6. What do you think about rephrasing it a little and moving it to right after the sentence

Yes, I think your suggestion makes sense.

dmitrilc: Thank you for the reminder. I totally forgot `Query.getLimit()` literally mean **limit**.

dmitrilc: I have committed the changes. Can you please check?

articles/binding-data/data-provider.asciidoc
```diff
@@ -135,7 +135,8 @@ grid.setItems(personRepository.findAll());
 Using callback methods is a more advanced way to bind data to components.
 This way, only the required portion of the data is loaded from your back end to the server memory.
 This approach is harder to implement and provides fewer features out of the box, but can save a lot of resources on the back end and the UI server.
-The component passes a query object as a parameter to your callback methods, where you can check what part of the data set is needed.
+
+When more data is needed, the component passes a link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html[[classname]`Query`] object as a parameter to your callback methods, where you can check what part of the data set is needed. The component will set the properties of the [classname]Query when it creates the [classname]Query object.
 
 Currently, only `Grid` and `ComboBox` properly support lazy data binding.
```
tarekoraby, line 141:
```suggestion

Currently, only `Grid` and `ComboBox` properly support lazy data binding.

Lazy data binding works as follows:

. The user performs an action that requires the component to display more data.
For example, the user might be scrolling down the list of items in a `Grid` component.
. The component automatically detects that more data is needed, and it passes a link:https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/data/provider/Query.html[`Query`] object as a parameter to your callback methods.
This object contains the necessary information about the data that should be displayed next to the user.
. Your callback methods use this [classname]`Query` object to fetch only the required data (usually from the back end) and return it to the component, which displays it automatically once the data is available.
```

articles/binding-data/data-provider.asciidoc
tarekoraby: It works now, thanks.
I added another suggestion to expand a bit on how the whole lazy data binding process works.

tarekoraby (approved):

PR #1609, opened 2022-08-10
docs: revert of removed fragment

MarcinVaadin:


articles/routing/navigation.asciidoc
```diff
@@ -176,6 +176,17 @@ Anchor anchor = new Anchor("/hello", "Go to /hello route");
 anchor.getElement().setAttribute("target", "_blank");
 ----
 
+Note that Vaadin Router intercepts all anchor elements since Vaadin 15.
+This means that unlike in earlier versions, clicking on a standard link does not cause a full page reload to happen by default.
```
jouni, line 180: No need to mention differences between older versions. Focus on explaining how things work right now in this version.

```suggestion
Vaadin router intercepts all anchor navigations, and clicking on a standard link does not cause a full page reload to happen by default.
```

jouni: I’m a bit unsure about “Vaadin Router” being a proper product name. But it looks like we don't really use it in the docs, so keeping it lowercase here would be consistent.

mshabarov (approved):

articles/routing/navigation.asciidoc
```diff
@@ -176,6 +176,16 @@ Anchor anchor = new Anchor("/hello", "Go to /hello route");
 anchor.getElement().setAttribute("target", "_blank");
 ----
 
+Vaadin router intercepts all anchor navigations, and clicking on a standard link does not cause a full page reload to happen by default.
```
tarekoraby, line 179:
```suggestion
Vaadin router intercepts all instances of anchor navigation, and clicking on a standard link does not cause a full page reload to happen by default.
```

PR #1586, opened 2022-08-03
Add note about external icon sets

tarekoraby:
Related to #721.

PR #1612, opened 2022-08-10
feat: add iconset generator

### LINKED ISSUE #721: Improve Icon documentation for other sets
Currently the Icon documentation mentions "You can use icons from Vaadin icons **or any other set.**" (https://vaadin.com/docs/latest/flow/components/icon)

But it's not really clear how this is done. 

Expectation:

Describe how to include some other icon sets like bootstrap or font awesome.

jouni:
Allow users to generate `<vaadin-iconset>` definitions by “uploading” a folder of `.svg` icons. The files are not actually uploaded anywhere, the generator works completely in the browser.

The generator code is of prototype quality, but it does what it needs to do. No proper error handling yet, so there’s no helpful error messages if someone tries to upload something else than `.svg` files.

Closes #721


tarekoraby:
Note that #1586 is aiming to add a caveat about the complexity of adding external icon sets. Is such caveat still valid in light of this iconset generator? Should we have the caveat for non-SVG iconsets?

jouni:
Good point. 

Yeah, I think this generator makes the use of SVG icons easy. Font icons are still more complicated to use, and the Start templates offer an example for that.

If you think it’s okay, I could move the caveat text to this PR instead.

jouni:
Would be nice to get the preview build fixed…

tarekoraby:
> If you think it’s okay, I could move the caveat text to this PR instead.

@jouni please do that.

tarekoraby:
This PR should fix https://github.com/vaadin/docs/issues/721.

articles/components/icons/index.asciidoc
```diff
@@ -81,7 +81,33 @@ return html`
 
 The icons are rendered as an inline `<svg>` element inside the shadow root of a `<vaadin-icon>` element, allowing you to style them using CSS.
 
+== Creating a Custom Icon Set
```
tarekoraby, line 84: A nitpick, but is it "Creating" or something like "Using" or "Importing" a custom icon set?

jouni: Right, “creating” might indeed give the wrong impression. How about:

```suggestion
== Using a Third-Party Icon Set
```

articles/components/icons/index.asciidoc
```diff
@@ -81,7 +81,33 @@ return html`
...
+
+A collection of SVG files can be converted to the `<vaadin-iconset>` format, which is “consumed” by the `<vaadin-icon>`/`Icon` component.
+You can take a look at the https://github.com/vaadin/web-components/blob/master/packages/icons/vaadin-iconset.js[implementation of the Vaadin icon set] for reference.
+
+Optionally, you can create a Java class/enum as a type-safe API for your Flow applications.
+
+=== Icon Set Generator
+
+This tool converts a folder of SVG files into `<vaadin-iconset>` and corresponding Java enum files.
+You can then download them and place them in your project.
```
tarekoraby, line 94: This sounds a bit abstract. I personally have some experience creating addons for another icon set. Yet, reading this doesn't make it clear to me where to place those files, nor how to use them in my project. Could we clarify this with some specific instructions?

jouni: I try to explain it here: https://github.com/vaadin/docs/pull/1612/files#diff-6173ef28a9a9f1df3042809fc52b9384cd19405135f48a6300b035974795e448R290-R292

jouni: For using the icons, perhaps there could be more instructions, but I thought the earlier content on the page would suffice.

jouni:
Finally we have the preview build 🥳

articles/components/icons/index.asciidoc
```diff
@@ -81,7 +81,33 @@ return html`
...
+
+[source,typescript,role="render-only embedded"]
+----
+include::{root}/frontend/demo/component/icons/iconset-generator.ts[render, persistent]
+----
+
+=== Other Icon Formats
+
+Third-party icon sets come in many different formats, and there is not a single integration method that works for every format.
```
dougjewsbury, line 103:
```suggestion
Third-party icon sets come in many different formats, and there is no single integration method that works for every format.
```

articles/components/icons/index.asciidoc
jouni: I tried to improve this a little bit in https://github.com/vaadin/docs/pull/1612/commits/2054809864f518358b2d99ace1863dbbb6a2a09a

tarekoraby:
This PR looks good to me. But maybe someone who knows a bit more than me about icon sets can be a better judge.

jouni:
Summoning the great @web-padawan for the task – can you spot some grand issues with this? Please ignore the poor code quality, all the `// @ts-ignore` and such. Could surely be better, but I didn't think it would be worth the effort to make this much cleaner. Might make it easier to maintain, but it smells a bit like premature optimization.

frontend/demo/component/icons/iconset-generator.ts
```diff
@@ -0,0 +1,433 @@
...
+  }
+
+  handleDragLeave() {
+    this.classList.remove('drag-active');
+  }
+
+  protected firstUpdated() {
+    const input = this.shadowRoot?.querySelector('.drop input');
+    // @ts-ignore
+    input.addEventListener('change', this.handleFiles.bind(this));
```
web-padawan, line 209: Why don't we use `@change=${this.handleFiles}` listener in the template?

frontend/demo/component/icons/iconset-generator.ts
```diff
@@ -0,0 +1,433 @@
...
+  }
+
+  async handleFiles() {
```
web-padawan, line 212: We could probably use `event.target` here to get a reference to the `<input>` element.

frontend/demo/component/icons/iconset-generator.ts
```diff
@@ -0,0 +1,433 @@
...
+    const dropzone = this.shadowRoot?.querySelector('.drop input') as HTMLInputElement;
+    const nameInput = this.shadowRoot?.querySelector('.name') as HTMLInputElement;
```
web-padawan, line 214:
```suggestion
    const nameInput = this.renderRoot.querySelector('.name') as HTMLInputElement;
```

web-padawan (changes requested):
There are way too many `@ts-ignore` comments. We could just use JS instead.  
Let me try to fix at least some of them before I can approve this PR.

jouni:
> There are way too many @ts-ignore comments. We could just use JS instead.

I implemented this generator in JavaScript initially, and then copied it to the docs project. But DSP has this limitation that only `.ts` or `.java` files can be “rendered”, so I had to use TypeScript: https://github.com/vaadin/docs-app/blob/master/src-node/asciidoc.js#L156-L164

frontend/demo/component/icons/iconset-generator.ts
jouni: I had trouble getting browsers to accept files via drag'n'drop on the file input, and I believe this had an accept on it. Not 100% sure.

web-padawan: Tested and it works for me in Chrome, Firefox and Safari.

web-padawan (approved):

PR #1623, opened 2022-08-15
Leave /docs prefix in discussion url

sashamihaha:
We're going to create notifications for new threads and we use urls from discussion component to generate links. I noticed that you drop /docs prefix from url parameter in the integrated discussion component and this causes invalid links. To avoid dirty workarounds on our side I think it's easier just to leave the prefix.


jouni (approved):
Fine by me. I thought that it should be dropped, but there is no reason from the docs site’s point of view to drop it.

PR #1626, opened 2022-08-16
chore: Fix typo in charts link computation

alvarezguille:
Accidentally copy-pasted `moduleNpmVersion:vaadin-crud` from another module.
Fixed to `moduleNpmVersion:vaadin-charts`
Both packages are released with same version so link was fine anyway.


jouni (approved):
How nice of me to not even spot that in the previous PR review 🙈

PR #1627, opened 2022-08-16
Add note about role of items' identities in data providers

### LINKED ISSUE #1496: Document Lombok @EqualsAndHashCode problems with Grid
A frequent confusion is caused by Lombok's default @EqualsAndHashCode in the context of Grid. This should be covered by the documentation with an example.

From https://stackoverflow.com/a/72646870/3608089 : 
>  By default [Lombok] generates equals and hashcode that use all the properties of the entity. This does not work with Vaadin's data model, as it expects the beans to be invariant, i.e. changing value of properties like confirmationPhone should not affect the beans identity. You can solve this either 1. defining in Lomboks annotation that uses only the id or userId for the equals and hashcode. Or 2. Extend DataProvider and override getId method so that uses either one.

> Example of Lombok use below
```java
@Entity
@EqualsAndHashCode(onlyExplicitlyIncluded = true)
@Table(name = "forms", schema = "click", catalog = "")
public class FormEntity {

    @EqualsAndHashCode.Include private long id;
    private String userId;
    private String requestorType;
    private String confirmationPhone;
    private Timestamp submittedDate;
    private int status;
    private Timestamp approvedDate;
    private String comments;
    private String organization;
    private PatientEntity patientEntity;
    ...
```

dmitrilc:
Also advises proper `hashCode()` and `equals()`

Fixes #1496


tarekoraby:
This is generally good, but as I noted in this [comment](https://github.com/vaadin/docs/issues/1496#issuecomment-1208101890),  we need to document both:

1. The importance of having a proper implementation of equals and hashcode for many Vaadin components, like the grid, to behave properly. If those are not properly implemented, then the Vaadin component will not be able to distinguish between the different items added to it.
2. The special consideration related to the use of Lombok.

Currently, the PR only does the second point.

articles/binding-data/_lombok-warning.adoc
```diff
@@ -0,0 +1,4 @@
+Vaadin's data model requires that beans' identities stay the same even when _non-identifying_ properties have changed.
+
+CAUTION: If you rely on [methodname]`equals()` and [methodname]`hashCode()` implementations that include all properties, such as in the case of Lombok's [classname]`@EqualsAndHashCode`, read
+<<{articles}/binding-data/data-provider/#Beware-of-Lombok’s-@EqualsAndHashCode, this>>.
```
tarekoraby, line 4:
```suggestion
Note that to function properly, components such as the `Grid` and `ComboBox` require that the assigned items have identifiers that are _stable_ and _unique_.
This usually requires a proper implementation of the [methodname]`hashCode()` and [methodname]`equals()` methods as described in this <<data-binding.data-provider.item-identifiers>> section.
```

articles/binding-data/data-provider.asciidoc
```diff
@@ -108,6 +110,8 @@ Check the component examples for more details on configuring the display of  lis
 The easiest way to pass data to the listing component is to use an array or [classname]`List`.
 You can create these easily yourself or pass values directly from your service layer.
 
+
+
```
tarekoraby, line 114:
```suggestion
```

tarekoraby:
Also, there is no need to mention the special case of overriding `getId()`. At least not in this PR. (I'm not sure if there are some edge cases that would require it, but let's add it anyway in a separate PR if and when we identify those edge cases).

articles/binding-data/data-provider.asciidoc
```diff
@@ -558,6 +562,46 @@ public class PersonDataProvider implements CallbackDataProvider.FetchCallback<Pe
...
+class PersonDataProvider extends CallbackDataProvider<SamplePerson,Void> {
+    
+    ...
+
+    @Override
+    public Object getId(SamplePerson item) {
+        return item.getId();
+    }
+}
+----
```
tarekoraby, line 604:
```suggestion
[[data-binding.data-provider.item-identifiers]]
== Ensuring Items Identities are Stable and Unique

When you bind items to a component, the identities of those items will be essential for the component to work properly.
For example, if you bind a list of `Person` objects to a `Grid`, the `Grid` will rely on the identities of the `Person` objects for various operations, such as for highlighting the selected rows and updating the data in an edited row.

**It is, therefore, important that the identities of the items are stable and unique.**
To ensure that, your POJO class should implement [methodname]`hashCode()` and [methodname]`equals()` methods based on the invariant properties of the POJO.
For example, the [classname]`Person` class in the example below have a `final long id` property that is used in the [methodname]`hashCode()` and [methodname]`equals()` methods as the identity of the [classname]`Person` object.
This way even if the [classname]`Person` object has a property such as its `phoneNumber` modified, the identity of the [classname]`Person` object will not change, and a component like the `Grid` will still be able to identify the object correctly.

[source,java]
----
class Person {
    private final long id;
    private String name;
    private String phoneNumber;

    public Person(long id, String name, String phoneNumber) {
        this.id = id;
        this.name = name;
        this.phoneNumber = phoneNumber;
    }

    // Getters and setters omitted for brevity

    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        Person person = (Person) o;
        return id == person.id;
    }

    @Override
    public int hashCode() {
        return Objects.hash(id);
    }
}
----

If you are using the Lombok library to generate [methodname]`hashCode()` and [methodname]`equals()` methods, you need to ensure that only the invariant properties are being used.
You can do that by setting `onlyExplicitlyIncluded = true` for [annotationname]`@EqualsAndHashCode`, and then annotate the entity's `id` property with [annotationname]`@EqualsAndHashCode.Include`.

[source,java]
----
@EqualsAndHashCode(onlyExplicitlyIncluded = true)
class Person {
    @EqualsAndHashCode.Include
    private final long id;
    private String name;
    private String phoneNumber;

    public Person(long id, String name, String phoneNumber) {
        this.id = id;
        this.name = name;
        this.phoneNumber = phoneNumber;
    }

    // Getters and setters omitted for brevity
}
----
```

articles/binding-data/_lombok-warning.adoc
dmitrilc: The new version looks fine. I like the _stable_ and _ unique_ adjectives used here. It definitely sounds better than my version.

Are you sure that we can skip making `@EqualsAndHashCode` obvious here? This version forces the reader to go to another page/section to realize the problem with Lombok.

tarekoraby: I think that's fine. The potential Lombok pitfall is only one of many caveats and considerations related to the binding of items to components. We cannot list every possible caveat in the introduction.

articles/binding-data/data-provider.asciidoc
```diff
@@ -558,6 +560,68 @@ public class PersonDataProvider implements CallbackDataProvider.FetchCallback<Pe
 personGrid.setItems(dataProvider);
 ----
 
+[[data-binding.data-provider.item-identifiers]]
+== Ensuring Items Identities are Stable and Unique
```
dougjewsbury, line 564:
```suggestion
== Ensuring That Item Identities Are Stable and Unique
```

articles/binding-data/data-provider.asciidoc
```diff
@@ -558,6 +560,68 @@ public class PersonDataProvider implements CallbackDataProvider.FetchCallback<Pe
...
+
+When you bind items to a component, the identities of those items will be essential for the component to work properly.
+For example, if you bind a list of `Person` objects to a `Grid`, the `Grid` will rely on the identities of the `Person` objects for various operations, such as for highlighting the selected rows and updating the data in an edited row.
+
+**It is, therefore, important that the identities of the items are stable and unique.**
+To ensure that, your POJO class should implement [methodname]`hashCode()` and [methodname]`equals()` methods based on the invariant properties of the POJO.
+For example, the [classname]`Person` class in the example below have a `final long id` property that is used in the [methodname]`hashCode()` and [methodname]`equals()` methods as the identity of the [classname]`Person` object.
```
dougjewsbury, line 571:
```suggestion
For example, the [classname]`Person` class in the following example has a `final long id` property that is used in the [methodname]`hashCode()` and [methodname]`equals()` methods as the identity of the [classname]`Person` object.
```

articles/binding-data/data-provider.asciidoc
```diff
@@ -558,6 +560,68 @@ public class PersonDataProvider implements CallbackDataProvider.FetchCallback<Pe
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id);
+    }
+}
+----
+
+If you are using the Lombok library to generate [methodname]`hashCode()` and [methodname]`equals()` methods, you need to ensure that only the invariant properties are being used.
+You can do that by setting `onlyExplicitlyIncluded = true` for [annotationname]`@EqualsAndHashCode`, and then annotate the entity's `id` property with [annotationname]`@EqualsAndHashCode.Include`.
```
dougjewsbury, line 605:
```suggestion
You can do this by setting `onlyExplicitlyIncluded = true` for [annotationname]`@EqualsAndHashCode`, and then annotating the entity's `id` property with [annotationname]`@EqualsAndHashCode.Include`.
```

articles/binding-data/data-provider.asciidoc
```diff
@@ -558,6 +560,68 @@ public class PersonDataProvider implements CallbackDataProvider.FetchCallback<Pe
...
 ----
 
+[[data-binding.data-provider.item-identifiers]]
+== Ensuring That Item Identities Are Stable and Unique
+
+When you bind items to a component, the identities of those items will be essential for the component to work properly.
+For example, if you bind a list of `Person` objects to a `Grid`, the `Grid` will rely on the identities of the `Person` objects for various operations, such as for highlighting the selected rows and updating the data in an edited row.
+
+**It is, therefore, important that the identities of the items are stable and unique.**
+To ensure that, your POJO class should implement [methodname]`hashCode()` and [methodname]`equals()` methods based on the invariant properties of the POJO.
```
OlliTietavainenVaadin, line 570: Maybe "Bean class" instead of "POJO class"?

tarekoraby: Is it actually a requirement for the items to be beans? Isn't it possible to use POJOs with custom getters and setters?

OlliTietavainenVaadin: True, it's not strictly necessary. "Java class" might be more appropriate still.

tarekoraby, line 570:
```suggestion
To ensure that, your Java class should implement [methodname]`hashCode()` and [methodname]`equals()` methods based on the invariant properties of the class.
```

tarekoraby: Changed.

OlliTietavainenVaadin (approved):

PR #1638, opened 2022-08-22
Add notice to header about viewing old docs version

tarekoraby:
The styles of the notice are copied from the v8 branch End of Life notice.


jouni (approved):
Not the ideal implementation, but let’s see if I can find the time to improve it later.

Issue #1632, opened 2022-08-18
Use new VaadinWebSecurity for Vaadin Spring Security


mshabarov:
Update [Securing Spring Boot Applications](https://vaadin.com/docs/latest/security/enabling-security#security-configuration-class) article with a new `VaadinWebSecurity` API introduced as a replacement for deprecated `WebSecurityConfigurerAdapter` [here](https://github.com/vaadin/flow/pull/14303):
- code example section https://vaadin.com/docs/latest/security/enabling-security/#security-configuration-class
- note that `VaadinWebSecurityConfigurerAdapter` can still be used but would be excluded in the later versions
- prepare a branch in https://github.com/vaadin/flow-crm-tutorial/blob/v23/src/main/java/com/example/application/security/SecurityConfig.java to update the code examples there as well.

tarekoraby:
When we change the docs to use `VaadinWebSecurity`, let's add a small note mentioning the old `VaadinWebSecurityConfigurerAdapter` for 23.1 users. We can remove this note after 23.1 is no longer supported.

PR #1653, opened 2022-08-25
docs: ConfigurerAdapter to component-based security configuration

MarcinVaadin:
Information about new component-based security configuration instead of ConfigurerAdapter

Fixes: #1632


articles/security/enabling-security.adoc
```diff
@@ -194,28 +199,12 @@ public class SecurityConfiguration
...
-    @Override
-    public void configure(WebSecurity web) throws Exception {
-        // Configure your static resources with public access here:
-        web.ignoring().antMatchers(
-                "/images/**"
-        );
-
-        // Delegating the ignoring configuration for Vaadin's
-        // related static resources to the super class:
-        super.configure(web); // <3>
```
mshabarov, line 208: I propose to keep the `configure(WebSecurity web)` method in the code snippet to highlight the opportunity to override it and explain to devs for what it is. And you kept the point <3> below the code snippet.

articles/security/enabling-security.adoc
```diff
@@ -245,6 +234,13 @@ Obviously, the implementation of the [methodname]`userDetailsService()` method i
 In a real-world application, you can change the Spring Security configuration to use an authentication provider for LDAP, JAAS, and other real-world sources.
 https://dzone.com/articles/spring-security-authentication[Read more about Spring Security authentication providers].
 
+.Component-based security configuration
+[NOTE]
+Spring Security 5.7.0-M2 deprecates the `WebSecurityConfigurerAdapter` and encourages users to move towards a component-based security configuration.
+Vaadin supports both approaches, although it is recommended to use component-based security configuration as in `SecurityConfiguration` example above.
+https://spring.io/blog/2022/02/21/spring-security-without-the-websecurityconfigureradapter[Read more about updating from WebSecurityConfigurerAdapter to component-based security configuration].
+`VaadinWebSecurityConfigurerAdapter` is still available for Vaadin 23.1 users but has been marked as deprecated and will be removed in the future.
```
mshabarov, line 242:
```suggestion
`VaadinWebSecurityConfigurerAdapter` is still available for Vaadin 23.2 users but has been marked as deprecated and will be removed in the future.

```

articles/security/enabling-security.adoc
```diff
@@ -245,6 +234,13 @@ Obviously, the implementation of the [methodname]`userDetailsService()` method i
 In a real-world application, you can change the Spring Security configuration to use an authentication provider for LDAP, JAAS, and other real-world sources.
 https://dzone.com/articles/spring-security-authentication[Read more about Spring Security authentication providers].
 
+.Component-based security configuration
+[NOTE]
+Spring Security 5.7.0-M2 deprecates the `WebSecurityConfigurerAdapter` and encourages users to move towards a component-based security configuration.
```
mshabarov, line 239:
```suggestion
Spring Security 5.7.0 deprecates the `WebSecurityConfigurerAdapter` and encourages users to move towards a component-based security configuration.

```

mshabarov (changes requested):

articles/security/enabling-security.adoc
```diff
@@ -245,6 +234,13 @@ Obviously, the implementation of the [methodname]`userDetailsService()` method i
 In a real-world application, you can change the Spring Security configuration to use an authentication provider for LDAP, JAAS, and other real-world sources.
 https://dzone.com/articles/spring-security-authentication[Read more about Spring Security authentication providers].
 
+.Component-based security configuration
```
mshabarov, line 237: I would also suggest to not use a note for this paragraph. Just place it before `This is how the view-based access control mechanism ...`.
Probably start the paragraph with ``VaadinWebSecurityConfigurerAdapter` is still available ...` is better, to make readers relax that they can still use the old class.

articles/security/enabling-security.adoc
mshabarov: For example, this user mentions migration guide for custom security implementation https://github.com/vaadin/flow/issues/14427#issue-1350521477 even.

MarcinVaadin: TBH we don't use that method anymore as marking static resources using `ignoring()` is not proper way. Static resources have been moved to default `HttpSecurity` filter. I can leave that method empty:
```
/**
     * Allows customization of WebSecurity configuration.
     */
    @Override
    public void configure(WebSecurity web) throws Exception {
        // Vaadin by default does not customize WebSecurity.
        super.configure(web);
    }
```
But for me it pointless...

I don't fully understand your remark about point <3> - it is next to calling super.configure(HttpSecurity) as it was previously and there is matching explanation below.

articles/security/enabling-security.adoc
MarcinVaadin: I would reorganize those paragraphs like:
image

articles/security/enabling-security.adoc
mshabarov: Vaadin abstract class doesn't use it, indeed. But it doesn't mean that the extension classes wouldn't use `WebSecurity` API for their purposes. I would add this method as you showed in your comment above.

About <3> - my only concern was that you deleted the <3> next to the code line, but kept it in the description block.

MarcinVaadin: There are 2 occurrences of <3>, there is one more in `configure(HttpSecurity)`

mshabarov: Ah, okay, I didn't notice that.

mshabarov (changes requested):
Looks good to me, but I set "request changes" to merge it after V23.2 GA is out.

mshabarov (approved):

PR #1663, opened 2022-08-29
Add mysql-connector-java version

tarekoraby:
Without the version explicitly set, the project doesn't run with an error: `[Cannot load driver class: com.mysql.jdbc.Driver Spring](https://stackoverflow.com/questions/33123985/cannot-load-driver-class-com-mysql-jdbc-driver-spring)`.


tarekoraby:
Actually, the error is probably generated if H2 dependency is there in the pom.xml. It's not caused by the missing MySQL dependency version.

PR #1716, opened 2022-09-15
refactor: Testing section

jouni:
Reintroduce the previous “UI unit testing” terminology, while introduce the term “end-to-end testing” for the previous browser-based testing.

Relocate the articles in correspondingly named folders, `testing/ui-unit` and `testing/end-to-end`. Keep the old folders and pages available (but they only contain a link to the new location) so that we don't need to create redirects. We can hopefully remove the old pages eventually (and perhaps create the redirects then).


articles/testing/ui-unit/index.adoc
```diff
@@ -0,0 +1,55 @@
...
+If the server-side bootstrap fails and throws an exception, the test method fails with the same exception.
+No need to go hunting for exceptions in a log located somewhere on a CI server.
+* No need to write a massive set of Page Objects.
+You are already on the server, and you have access to the actual Java components that are already providing you with high-level APIs, exactly as Page Objects do.
+* Full access to the database.
+You are on the server side, so you can access the database from your tests in the same way your business logic accesses the database.
+You can run a bunch of SQL statements to restore the database to a known state before every test.
+Even better, you can run the test in a transaction, then roll back after the test to perform a fast database revert to a known state.
+
+With this technique, you can run 600 UI tests in 7 seconds, as opposed to 1 to 2 hours with a Selenium-based approach.
```
tarekoraby, line 42:
```suggestion
* *Fast*: browser-less tests are typically 100&times; faster than end-to-end tests and run in 5-60 milliseconds, depending on their complexity.
* Reliable: no arbitrary sleeps are needed, as the test is executed on the server side and it can wait until the request is fully processed.
No random failures because of incompatibility between the Selenium drivers used in end-to-end testing and the browser.
* Headless: the tests run headless, as there is no browser.
No need to set up the screen in your CI environment.
* Robust: the test runs in the same JVM as the server-side components.
If the server-side bootstrap fails and throws an exception, the test method fails with the same exception.
No need to go hunting for exceptions in a log located somewhere on a CI server.
* No need to write a massive set of Page Objects.
You are already on the server, and you have access to the actual Java components that are already providing you with high-level APIs, exactly as Page Objects do.
* Full access to the database.
You are on the server side, so you can access the database from your tests in the same way your business logic accesses the database.
You can run a bunch of SQL statements to restore the database to a known state before every test.
Even better, you can run the test in a transaction, then roll back after the test to perform a fast database revert to a known state.

With this technique, you can run 600 UI tests in 7 seconds, as opposed to 1 to 2 hours with the end-to-end testing approach.
```
I suggest using "end-to-end testing" instead of "Selenium based" as people might not immediately realize that e2e is actually based on selenium.

articles/testing/end-to-end/index.adoc
```diff
@@ -0,0 +1,35 @@
+---
+title: End-to-End Testing
+order: 20
```
tarekoraby, line 3:
```suggestion
description: TestBench end-to-end testing enables you to simulate and verify real-world user interaction with your application
order: 20
```

articles/testing/end-to-end/index.adoc
```diff
@@ -0,0 +1,35 @@
...
+* Control one or several browser instances from Java, both desktop and mobile browsers
+* A powerful and robust way to describe your tests, so that they do not break with application changes
+* A high-level API for finding the component you want to interact with
+* Vaadin Component API for easy interaction with all <<{articles}/components#, Vaadin components>> and HTML elements
+* Automatic screen comparison, highlighting differences
+* Assertion-based UI state validation
+* Easily run tests in parallel
+* Test grid support to speed up tests by running in parallel on multiple browsers on selected operating systems
+* Support for JUnit and other testing frameworks
+* All features available in Selenium
```
tarekoraby, line 35:
```suggestion
TestBench end-to-end testing simulates a user of your application, performs the tasks specified using Java code, and verifies that the expected actions take place in the application.

TestBench end-to-end testing approach can also visually inspect your application and detect unintentionally introduced changes, and verify that the application is visually OK in all the browsers you are testing with.
TestBench also includes special support for other Vaadin products, making testing easy and robust compared to generic web testing solutions.

Although it is not the main purpose of TestBench, you can also use the end-to-end testing approach to automate mundane tasks such as filling out forms.

== Features

The main features of end-to-end testing are:

* Control one or several browser instances from Java, both desktop and mobile browsers
* A powerful and robust way to describe your tests, so that they do not break with application changes
* A high-level API for finding the component you want to interact with
* Vaadin Component API for easy interaction with all <<{articles}/components#, Vaadin components>> and HTML elements
* Automatic screen comparison, highlighting differences
* Assertion-based UI state validation
* Easily run tests in parallel
* Test grid support to speed up tests by running in parallel on multiple browsers on selected operating systems
* Support for JUnit and other testing frameworks
* All features available in Selenium
```

tarekoraby (approved):

PR #1734, opened 2022-09-21
chore: added documentation for CdiComponent annotation

mcollovati:
CdiComponent stereotype is needed to let the container scan and manage pseudo-scope annotated components


articles/integrations/cdi/contexts.asciidoc
```diff
@@ -36,6 +36,10 @@ Injection of a pseudo-scoped bean creates a direct reference to the object, but
 ** Injecting into a larger scope binds the instance from the currently active smaller scope, and ignores changes in the smaller scope.
 For example, after being injected into a session scope, a `@UIScoped` bean will point to the same instance (even if its [classname]`UI` is closed), regardless of the current [classname]`UI`.
 
+Pseudo scopes are not bean defining annotation, so, in implicit bean archives, `@UIScoped` and `@RouteScoped` components needs to be marked with the `@CdiComponent` stereotype annotation, that lets the container scan and manage the instances of the annotated types.
```
dougjewsbury, line 39:
```suggestion
Pseudo scopes are not bean-defining annotations, so, in implicit bean archives, [annotationname]`@UIScoped` and [annotationname]`@RouteScoped` components need to be marked with the [annotationname]`@CdiComponent` stereotype annotation, which lets the container scan and manage the instances of the annotated types.
```

jouni:
I’m wondering why both “Pseudo scopes” and “pseudo-scopes” are used. I assume the latter is the correct one to use (hyphenated).

mcollovati:
On the specifications, it is always 'pseudo-scope`

mshabarov (approved):

PR #1740, opened 2022-09-22
docs: observability service configuration instructions

sissbruecker:
Add instructions on how to configure service attributes

### DIFF
--- a/articles/tools/observability/advanced/configuration.asciidoc
+++ b/articles/tools/observability/advanced/configuration.asciidoc
@@ -16,16 +16,78 @@ Service attributes are as follows:
 
 * `service.name` is an attribute that is used to distinguish a service by name.
 +
-In the custom distribution this defaults to `vaadin`
+The default value is `vaadin`.
+This is the only attribute with a default value, as it is the only one required.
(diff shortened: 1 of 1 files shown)


articles/tools/observability/advanced/configuration.asciidoc
```diff
@@ -16,16 +16,78 @@ Service attributes are as follows:
...
+The default value is `vaadin`.
+This is the only attribute with a default value, as it is the only one required.
 * `service.namespace` helps to distinguish a group of services
 * `service.instance.id` helps to distinguish instances of the same service that exist at the same time
 +
 Must be unique for each instance of the same `service.namespace`, `service.name` pair
 * `service.version` Semantic Versioning string of the service version
 
-Only `service.name` has a default value as it is the only one required.
+The `service.name` attribute is configured using the `otel.service.name` property, either in the `agent.properties` configuration file, as a system property, or as environment variable.
```
sissbruecker, line 27: Originally I planned to just document the `otel.resource.attributes` property, but it seems the service name from there is always overwritten by the instrumentation default. So I added separate instructions for name, and the remaining attribute.

caalador (approved):

PR #1751, opened 2022-09-23
refactor: use correct TS types for change event

web-padawan:
## Description

Fixed `change` events to use correct types. Also fixed one example that was broken due to wrong type.

### DIFF
--- a/frontend/demo/component/custom-field/custom-field-native-input.ts
+++ b/frontend/demo/component/custom-field/custom-field-native-input.ts
@@ -2,7 +2,7 @@ import 'Frontend/demo/init'; // hidden-source-line
 import { html, LitElement } from 'lit';
 import { customElement, state } from 'lit/decorators.js';
 import '@vaadin/custom-field';
-import type { CustomFieldValueChangedEvent } from '@vaadin/custom-field';
+import type { CustomFieldChangeEvent } from '@vaadin/custom-field';
 import '@vaadin/horizontal-layout';
 import { applyTheme } from 'Frontend/generated/theme';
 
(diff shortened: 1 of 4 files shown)


frontend/demo/component/custom-field/custom-field-native-input.ts
```diff
@@ -24,7 +24,7 @@ export class Example extends LitElement {
       <vaadin-custom-field
         label="Payment information"
         theme="whitespace"
-        @change="${(e: CustomFieldValueChangedEvent) => (this.customFieldValue = e.detail.value)}"
+        @change="${(e: CustomFieldChangeEvent) => (this.customFieldValue = e.target.value || '')}"
```
web-padawan, line 27: This example wasn't broken because `vaadin-custom-field` fires `change` event with `detail`, which is different from all other components. Not sure it's a good idea to use it though, so I changed the type.

frontend/demo/component/datepicker/date-picker-custom-functions.ts
```diff
@@ -53,7 +53,7 @@ export class Example extends LitElement {
         label="Select a date:"
         value="${this.selectedDateValue}"
         helper-text="Date picker configured to use ISO 8601 format"
-        @change="${(e: DatePickerValueChangedEvent) => (this.selectedDateValue = e.detail.value)}"
+        @change="${(e: DatePickerChangeEvent) => (this.selectedDateValue = e.target.value)}"
```
web-padawan, line 56: This example was broken: there is no `detail` object, so the following error was thrown:

Screenshot 2022-09-23 at 14 23 55

tomivirkki (approved):

PR #1775, opened 2022-09-29
chore: added feature flags documentation

### LINKED ISSUE #1711: Add documentation for the ability to set feature flags also from system properties
The documentation should cover the property name pattern and obviously the fact that it has higher priority over the properties file. PR that added the feature: vaadin/flow#14408

mcollovati:
Closes #1711


articles/configuration/feature-flags.adoc
```diff
@@ -0,0 +1,73 @@
...
+|===
+|Feature name
+|Description
+|Status
+
+|enforceFieldValidation
+|Enforce client / constraint / binder validation
+|Experimental
+
+|spreadsheetComponent
```
caalador, line 50: @jouni does this need to be added to `.github/styles/Vocab/Docs/accept.txt` or can it be ignored some other way as it is a configuration?

articles/tools/ce/advanced/backend-api.asciidoc
```diff
@@ -153,12 +153,23 @@ The [methodname]`addMembershipListener` should return a [classname]`Registration
 
 == Feature Flag
 
-In order to use the Backend API, it must be explicitly enabled with a feature flag:
+In order to use the Backend API, it must be explicitly enabled with a feature flag.
```
caalador, line 156:
```suggestion
To use the Backend API, it must be explicitly enabled with a feature flag.
```

articles/configuration/feature-flags.adoc
jouni: I would suggest to use code/monospace formatting for these, as they are code more than plain text.

jouni: accept.txt should only contain names or words that are used in regular English. It currently contains some words that perhaps should not be there, but let’s not grow that list.

jouni: I can take care of cleaning this up.

mcollovati: Thanks, @jouni

jouni (approved):

PR #1785, opened 2022-10-03
chore: fix ESLint warnings, use type imports

web-padawan:
## Description

- Fixed some ESLint warnings detected by using config from web components repo,
- Updated most of components imports to use `import type` where it makes sense.


frontend/demo/foundation/icons-preview.ts
```diff
@@ -117,8 +117,8 @@ export class IconsPreview extends HTMLElement {
 
     this.innerHTML = html;
 
-    const search = this.querySelector('input');
-    search?.addEventListener('input', () => {
+    const search = this.querySelector('input') as HTMLInputElement;
```
tomivirkki, line 120: nit: prefer `querySelector<HTMLInputElement>('input')` over type casting

tomivirkki (approved):

frontend/demo/foundation/icons-preview.ts
web-padawan: Thanks, fixed.

tomivirkki: Ah, TS is smart enough to determine the type from the query 👍  Then `search?.addEventListener...` would be enough

web-padawan: Actually it was already using `search?.addEventListener` so I reverted this change as unnecessary.

PR #1789, opened 2022-10-04
Add an article describing the new Hilla multi-module engine

### LINKED ISSUE #1353: Hilla: using the multi-module engine
While the multi-module generator is not a part of the Hilla platform yet, there are some explicit steps necessary for users to switch from the old generator for the new one.

Add an article, that explains:
- The multi-module generator features and differences with the old one
- How to use the multi-module generator
- The necessary generator plugin configuration in `pom.xml` in brief

cromoteca:
This article explains how to enable and use the new multi-module engine and illustrates the most important advantages.

Closes #1353 

- [x] The example project is in my GitHub, it should probably be transferred to Vaadin.


articles/advanced/multimodule.asciidoc
```diff
@@ -0,0 +1,224 @@
+---
+title: Multi-Module Engine
+description: Using a Hilla Engine — a pluggable set of Java bytecode parser & TypeScript AST generator.
+order: 110
+layout: page
+---
+
+[role="since:com.vaadin:vaadin@V1.2"]
+= Multi-Module Engine
```
jouni, line 9: The previous syntax does not work in AsciiDoc (setting a role/class on the main page heading). You need to wrap the inline text with a span that has a role/class.

```suggestion
= [since:com.vaadin:vaadin@V1.2]#Multi-Module Engine#
```

platosha (approved):
Content-wise looks great! Thanks!

PR #1790, opened 2022-10-04
In-Depth Course, Backend Page: Add missing import in the initial class snippet

starg09:
UUID import wasn't added when changed in the final fixes prior to launching Hilla: https://github.com/vaadin/docs/commit/3bc1fb572767d956bdacd3aaaf472f0e5d84a6bd


jouni:
Thanks!

PR #1793, opened 2022-10-05
chore: remove not necessary font-family style

web-padawan:
## Description

Removed `font-family` styles from `ProgressBar` examples as it gets applied by the application theme:

Screenshot 2022-10-05 at 13 51 55


jouni (approved):
Could use utility classes as well, instead of inline styles.

PR #1802, opened 2022-10-06
Changed to use JUnit 5

mstahv:
related pr for the example project: https://github.com/vaadin/flow-crm-tutorial/pull/99

Let's still agree if we want both junit 4 & 5 for the project before merging both.


mstahv:
I tried to find a solution to use Junit 5 to control the TestBench test in the tutorial today, but didn't get any good solution done. So I guess it is either to accept both JUnit 4 & 5 usage in the same app or to postpone this change until TestBench gets updated.

MarcinVaadin:
Testbench 9 with JUnit5 updates pushed to https://github.com/vaadin/docs/pull/1865

jouni:
@mstahv, are you still working on this PR? Anything missing from it, or can we merge it?

mstahv:
Let's merge it. JUnit 5 is mainstream, so let's use it where possible. The only downside is that people will notice that they can't really use JUnit 5 with testbench yet. Hopefully we could still get TestBench for V24 to support JUnit 5, V24 will have a long life ahead still.... (cc: @mshabarov)

jouni (approved):

jouni:
> The only downside is that people will notice that they can't really use JUnit 5 with testbench yet

That should be covered by https://github.com/vaadin/docs/pull/1865

PR #1816, opened 2022-10-10
Edits to Closed PR 1414 (backend-api file).

russelljtdyer:


articles/tools/ce/advanced/backend-api.asciidoc
```diff
@@ -10,77 +10,77 @@ layout: page
 
 .Experimental feature
 [IMPORTANT]
-The Collaboration Engine backend for clustering support is an experimental feature, which means that its behavior, API, and look and feel might still change.
-You must enable it with a <<Feature Flag>>.
+The Collaboration Engine backend for clustering support is an experimental feature. This means that its behavior, API, as well as its look and feel might change.
+You may enable it with a <<Feature Flag>>.
 
-Using Collaboration Engine out of the box in an application running in a clustered environment would result in users being able to collaborate only with others connected to the same application instance, meaning, on the same node.
+Using Collaboration Engine with its default settings in an application running in a clustered environment would result in users being able to collaborate only with others connected to the same application instance (i.e., on the same node).
```
jouni, line 16: We try to avoid using the latin abbreviations `e.g.` and `i.e.`, as those are not necessarily easy for non-native English readers to understand.

And parenthesis are often unnecessary. I don't mean to be strict about these, but I think it’s always good to evaluate the need for them.

I wonder if there’s a better way of explaining the relationship between an application instance and a server node?

```suggestion
Using Collaboration Engine with its default settings in an application running in a clustered environment would result in users being able to collaborate only with others connected to the same application instance, that is, on the same node.
```

articles/tools/ce/advanced/backend-api.asciidoc
```diff
@@ -10,77 +10,77 @@ layout: page
...
-To implement subscriptions to events, you add an item listener to the Hazelcast list.
-The subscriber receives all past and future events for this Event Log.
-A newly added subscriber should initially receive all previous events in the log based on their original order, so that it can catch up with the latest state.
-New events should be delivered (in order) only after all previous events have been delivered.
-It is not allowed to invoke the [methodname]`subscribe` method again until the previous subscriber has been removed.
+To implement subscriptions to events, you would add an item listener to the Hazelcast list.
+The subscriber will receive all past and future events for this Event Log.
+A newly added subscriber should receive initially all of the previous events in the log based on their original order. This allows it to catch up to the latest state.
+New events should be delivered, in order, only after all previous events have been delivered.
+The [methodname]`subscribe` method may not be invoked again until the previous subscriber has been removed.
```
jouni, line 51: We try to avoid the future tense (will, won’t). Is it actually needed here?

I think “would” and “wouldn’t” fall into similar category. Can those be replaced with “can” and “can’t”?

russelljtdyer: I think it works better here.

jouni:
Surpassed by #1846

PR #1845, opened 2022-10-25
Better examples for calling JS methods

samie:


articles/create-ui/element-api/client-server-rpc.asciidoc
```diff
@@ -63,11 +68,12 @@ The supported types are `String`, `Integer`, `Double`, `Boolean` and `Element`.
 [source,java]
 ----
 public void complete() {
-    getElement().executeJs("this.complete($0)", true);
+    getElement().executeJs("MyModule.complete($0)", true);
 }
 ----
 
-It's also possible to call the [methodname]`executeJs()` method to access methods and fields of a Web Component.
+It's also possible to call the [methodname]`executeJs()` method to access methods and fields of a Web Component using [methodname]`this.myFieldName`.
```
jouni, line 75: Is this a method call? If yes, then it should have the closing brackets `()`.

```suggestion
It's also possible to call the [methodname]`executeJs()` method to access methods and fields of a Web Component using `this.myFieldName`.
```

articles/create-ui/element-api/client-server-rpc.asciidoc
```diff
@@ -63,11 +68,12 @@ The supported types are `String`, `Integer`, `Double`, `Boolean` and `Element`.
...
+If you do not need the element to be initialized before, you can use [methodname]`UI.getCurrent().getPage().executeJs` instead.
```
jouni, line 76:
```suggestion
If the element doesn't need to be initialized first, you can use [methodname]`UI.getCurrent().getPage().executeJs()` instead.
```

jouni (commented):
Thanks for the update!

jouni:
I’ll leave the subject matter review for @Legioth.

Legioth (approved):

PR #1853, opened 2022-10-28
Update DSP branch from latest

manolo:


jouni (approved):

jouni:
@tomivirkki and @rolfsmeds: do you think it’s safe to merge this PR? I assume it will immediately start affecting any new installations of DSP. Is there something we should consider, like perhaps combine this PR with the DSP 2.1 release which is currently in alpha?

tomivirkki:
I think the intention of this PR is mainly to [test the script](https://vaadin.slack.com/archives/C046R3PPJLR/p1667390182341429). @manolo? The `dsp` branch should ideally only be updated from `latest` when it's running a stable Vaadin version.

manolo:
@tomivirkki  is right, this was a draft for testing the script, will open a new one when latest is ready to be picked

PR #1859, opened 2022-11-02
docs: add spreadsheet component documentation

tomivirkki:
Adds documentation for the `Spreadsheet` component.
- Migrate existing documentation from https://vaadin.com/docs/v8/spreadsheet
- Add live examples


jouni:
For future reference, the `[discussion-id]`...` ` should be a random UUID/GUID string. You can get one from https://www.uuidgenerator.net, for example

rolfsmeds (approved):

Issue #1848, opened 2022-10-27
Documentation for JUnit 5 support in TestBench end-to-end testing


mshabarov:
JUnit 5 support was included into Vaadin 23 for Browser-less testing feature (aka UI Unit testing).
Since Vaadin 24 the JUnit 5 support is going to be included into TestBench end-to-end testing part.
TestBench chapter https://vaadin.com/docs/latest/testing/end-to-end should be updated with the following:
- which TestBench maven artifacts/dependencies to use for JUnit 5
- which is used by default, 4 or 5
- rules->extension brief overview
- changes in annotations: "Test" can be used, but to be able to run the test method with different browser capabilities "EndToEndTest" annotation is used.
- parallel run with JUnit 5 native API
- retry on failures with Surefire plugin
- base test class: it's optional and contains util methods
- Flow test util helper classes.

PR #1865, opened 2022-11-04
docs: Add JUnit5 examples for TestBench9

MarcinVaadin:
Added JUnit5 examples to present new features of Testbench 9 that will be available with Vaadin 24.

Fixes https://github.com/vaadin/docs/issues/1848


articles/testing/end-to-end/bdd.adoc
```diff
@@ -63,6 +65,47 @@ public class CalculatorSteps extends TestBenchTestCase {
     }
 }
 ----
+[source,java]
+----
+<source-info group="JUnit5"></source-info>
+public class CalculatorSteps implements HasElementQuery {
```
tomivirkki, line 71: Where possible, prefer writing code in a Java/ts/js file and then including it in the asciidoc file

MarcinVaadin: All examples in Testing/* were just inline code snippets, should I refactor everything or is it just a notice for future changes?

tomivirkki: No need to refactor now IMO, but just a general recommendation. Inline code snippets tend to age easily since they never break when the project is compiled.

jouni: Yes, it would be preferable to keep code under the `src` and `frontend` folders when possible. Sometimes it isn't. I assume we could include test classes in the project as well without issues.

All the existing TestBench examples are obviously not extracted like that, so it’s convenient to continue the same approach. But perhaps we should start at some point to move to the better approach, so that as much of the example code as possible is compiled.

jouni: An example where it's not possible to avoid writing examples inside the asciidoc content:

Screenshot 2022-11-04 at 15 07 33

I assume it’s not possible to have both dependencies at once.

articles/testing/selenium.adoc
```diff
@@ -346,5 +346,42 @@ The `xpath` can be generated by the https://www.selenium.dev/selenium-ide/[Selen
 
 For more usage scenarios, you can see the official https://www.selenium.dev/documentation/webdriver/elements/[Selenium doc].
 
+== Selenium-Jupiter extension
+
+https://bonigarcia.dev/selenium-jupiter/[Selenium-Jupiter] is a JUnit5 extension that can be used to initialize, run and manage browser based tests.
+
+Combining TestBench with Selenium-Jupiter is straightforward, like in example:
```
mshabarov, line 353:
```suggestion
Combining TestBench with Selenium-Jupiter is shown in the following example:
```

mshabarov (changes requested):
We also need to put a general note in the first chapter of end-to-end part that since Vaadin 24 both JUnit 4 and JUnit 5 test framework versions are supported. Here maybe https://vaadin.com/docs/latest/testing/end-to-end/#features

articles/testing/end-to-end/bdd.adoc
MarcinVaadin: It is technically possible to use both in same time, but what do mean by "pulled from the pom.xml"?

jouni: > what do mean by "pulled from the pom.xml"?

Tag the region in the pom.xml file like this:

```xml
<!-- tag::junit5-dependency[] -→
<dependency>
  ...
</dependency>
<!-- end::junit5-dependency[] -->
```

The include that in the docs page like this:

```asciidoc
[source,xml]
----
include::{root}/pom.xml[tag=junit5-dependency,indent=0,group="JUnit 5"]
----
```

mshabarov (approved):

articles/testing/index.adoc
```diff
@@ -74,6 +74,8 @@ Moreover, *end-to-end testing enables you to test the client-side functionality*
 
 These tests are impossible with the UI unit testing approach because UI unit tests run on the server side without initializing the browser.
 
+Since Vaadin 24 both JUnit 4 and JUnit5 test frameworks are supported on end-to-end tests.
```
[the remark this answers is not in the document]
MarcinVaadin: I've unified naming everywhere to "JUnit 4" and "JUnit 5".

MarcinVaadin:
There will be updates related to testbench update.

articles/testing/index.adoc
```diff
@@ -74,6 +74,8 @@ Moreover, *end-to-end testing enables you to test the client-side functionality*
 
 These tests are impossible with the UI unit testing approach because UI unit tests run on the server side without initializing the browser.
 
+Since Vaadin 24 both JUnit 4 and JUnit 5 test frameworks are supported on end-to-end tests.
```
jouni, line 77:
```suggestion
[since:com.vaadin:vaadin@V24]#Both JUnit 4 and JUnit 5 test frameworks are supported on end-to-end tests#.
```

articles/testing/selenium.adoc
```diff
@@ -346,5 +346,42 @@ The `xpath` can be generated by the https://www.selenium.dev/selenium-ide/[Selen
 
 For more usage scenarios, you can see the official https://www.selenium.dev/documentation/webdriver/elements/[Selenium doc].
 
+== Selenium-Jupiter extension
```
jouni, line 349:
```suggestion
== Selenium-Jupiter Extension
```

mshabarov (approved):

vursen (commented):
A few minor suggestions.

articles/testing/end-to-end/multiple-browsers.adoc
```diff
@@ -18,44 +18,47 @@ When running on a hub, you don't need a local WebDriver installed. The WebDriver
...
 
-[NOTE]
 When running tests in parallel, you need to ensure that the tests are independent and don't affect each other in any way.
 
-[NOTE]
 You can also handle creation and destruction of the WebDriver manually when running on a hub if you don't want to use `ParallelTest` for one reason or another.
 In this case, you should create a `RemoteWebdriver` with the correct hub URL and set suitable `DesiredCapabilities` on the driver.
 
+==== Notes for JUnit 5
+If your test class already creates `WebDriver` implementing `DriverSupplier` you can remove it in favor of automatic driver initialization by TestBench.
```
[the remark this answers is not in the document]
MarcinVaadin: I've omitted that one and cannot apply it now :-/

jouni:
Actually, I’ll move Russell’s edits to another branch, so this PR stays focused on the original scope.

PR #1874, opened 2022-11-08
Create a tutorial for Azure Kit

MatthewVaadin:
Create a getting started tutorial that covers the basics, including:

- Adding the Azure Kit dependency
- Configuring an application
- Deploying the application and Hazelcast to a local cluster
- Scaling the application
- Rolling out a new version of the application

### DIFF
--- a/articles/tools/azure/getting-started.asciidoc
+++ b/articles/tools/azure/getting-started.asciidoc
@@ -1,24 +1,39 @@
 ---
 title: Getting Started
-description: Step-by-step guide showing how to add Azure Kit to your application.
+description: Step-by-step guide showing how to add Azure Kit
+  to your application.
 order: 10
 ---
 
 = Getting Started with Azure Kit
 :sectnums:
 
-This guide covers the following topics:
+Azure Kit provides high availability for Vaadin applications
+ running in a Kubernetes cluster (including Azure).
 
-- Downloading Azure Kit by ...
-// compared to SSO Kit
-- Setting up TEXT;
-// What do they need to set up to get started?
-- Running a Vaadin Flow application with Azure Kit.
+This tutorial guides you through setting up and deploying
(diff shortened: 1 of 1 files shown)


articles/tools/azure/getting-started.asciidoc
```diff
@@ -29,58 +44,434 @@ To get started you first need to add Azure Kit as a dependency to your applicati
 </dependency>
 ----
 
+Next, you need to set some properties to configure Azure
+ Kit. These properties can be added to the application
+ configuration file, where you can define the packages that
```
mcollovati, line 49: The `vaadin.serialization.include-packages` property does not define which packages should be serialized (all objects in HTTP session will be serialized, regardless of the package), but which packages to consider when inspecting objects for transient fields.
Indeed, the name of the property is misleading.

MatthewVaadin: Ah, thanks for the clarification.

articles/tools/azure/getting-started.asciidoc
```diff
@@ -30,57 +45,431 @@ To get started you first need to add Azure Kit as a dependency to your applicati
...
-1. Build the application JAR:
-[source,terminal]
+Next, you need to add the following to the application
+configuration file:
+
+.application.properties
+[source,properties]
 ----
-# mvn clean package -Pproduction,redis
+vaadin.azure.hazelcast.kubernetes.service-name=hazelcast-service
```
mcollovati, line 72: I would add the suggestion to set `vaadin.serialization.include-packages` to limit transient inspection. 
Otherwise, there may be exceptions because of reflective access to classes from not opened modules.
This happens for example with Java 17 when trying to inspect classes from `java.util` package

mcollovati (approved):

PR #1876, opened 2022-11-08
Simplified docs as simplifying the default Dockerfile coming view s.v.c

mstahv:
Related PR: https://github.com/vaadin/start/pull/2099#pullrequestreview-1172060312


jouni (approved):
Thanks! I'll try to do some small adjustments to help with maintenance, and then merge. But not blocking because of those. Some minor language edits are also needed.

PR #1889, opened 2022-11-10
docs: replace hasControls with stepButtonsVisible

sissbruecker:
Replaces usages of the deprecated `hasControls` property with the `stepButtonsVisible` property, and updates references to the buttons in the asciidoc.

Needs to wait for the next 23.3 alpha release before being merged.

Part of https://github.com/vaadin/flow-components/issues/1106

### DIFF
--- a/articles/components/number-field/index.asciidoc
+++ b/articles/components/number-field/index.asciidoc
@@ -8,12 +8,12 @@ page-links:
 = Number Field
 
 // tag::description[]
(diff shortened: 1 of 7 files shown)


jouni:
Very minimal English content changes, so no need for a language check here.

jouni:
`latest` is using 23.3.0 now, so I believe this can be merged.

PR #1899, opened 2022-11-17
Changing tutorial to use Azure to host the app

mstahv:
Opening draft PR about Azure based deployment (replacing Heroku), don't merge yet!


mstahv:
@roryp Some early drafting of Azure based hosting. 

Just today discovered the `up` command. Would it make sense to use that here? We could get rid of explaining Docker builds and to make the deployment look much simpler. Used this on top of the project and it seemed to work 🤷‍♂️

    az containerapp up -n mytestcontainerapp --source .

And this as suggested by the CLI output to enable ingress:

    az containerapp ingress enable -n mytestcontainerapp -g flowcrmspringapp --type external --target-port 8080 --transport auto

roryp:
> @roryp Some early drafting of Azure based hosting.
> 
> Just today discovered the `up` command. Would it make sense to use that here? We could get rid of explaining Docker builds and to make the deployment look much simpler. Used this on top of the project and it seemed to work 🤷‍♂️
> 
> ```
> az containerapp up -n mytestcontainerapp --source .
> ```
> 
> And this as suggested by the CLI output to enable ingress:
> 
> ```
> az containerapp ingress enable -n mytestcontainerapp -g flowcrmspringapp --type external --target-port 8080 --transport auto
> ```

yes - I use the UP command all the time.
Does the ingress work?

mstahv:
👌 Let's use that then. I'll remove most of what I wrote, but this is like with code, deleting it feels good 😎 Azure subscription and installed CLI is then pretty much only things that is needed. And as the Dockerfile is in the project already, this gets pretty much to the same level as with Heroku before then, or even simpler as we left out the production DB part 💪

Ingress works fine, as long as single node is used. Your colleague mentioned, session affinity is coming soon.

jouni:
> don't merge yet!

Draft PRs can't be merged, so not really a risk. Just mark it ready for review when you’re ready :)

mstahv:
@jouni Can you push this to language check, content wise I'm quite confident we are ready. Tested the modified chapter now with two crash-test-dummies😁 Will check it out today afternoon still with @roryp.

Issue #1655, opened 2022-08-26
Migrate docs from Webpack to Vite


tomivirkki:
Currently, Docs uses Webpack for frontend build by enabling [the feature flag](https://github.com/vaadin/docs/blob/latest/src/main/resources/vaadin-featureflags.properties#L2). However, using Webpack with Vaadin is deprecated in favor of Vite since version 23.2 and the support will be dropped in the upcoming majors.

The following changes should be done to migrate Docs to Vite:
- Create a Vite configuration that does the same things as the current [Webpack config](https://github.com/vaadin/docs/blob/latest/webpack.config.js)
  - [Excludes all vaadin and polymer imports from the application JS bundle](https://github.com/vaadin/docs/blob/latest/webpack.config.js#L40-L68)
  - [Exports the `applyTheme` function to the global object as `window.__applyTheme`](https://github.com/vaadin/docs/blob/latest/webpack.config.js#L26-L38)
- Drop the Webpack feature flag
- Remove the Webpack config file

vursen:
Does it mean that the DSPublisher config itself will be still using Webpack?

jouni:
> Does it mean that the DSPublisher config itself will be still using Webpack?

Yes, Gatsby still uses webpack.

PR #1906, opened 2022-11-21
refactor: migrate from Webpack to Vite

vursen:
The PR migrates the docs to Vite as Webpack support will be dropped in Vaadin 24.

Fixes https://github.com/vaadin/docs/issues/1655


src/main/resources/application.properties
```diff
@@ -2,10 +2,10 @@ server.port=8080
...
 server.compression.mime-types=application/json,application/xml,text/html,text/xml,text/plain,application/javascript,text/css
 # Ensure application is run in Vaadin 14/npm mode
-vaadin.compatibilityMode = false
+# vaadin.compatibilityMode = false
 logging.level.org.atmosphere = warn
 vaadin.urlMapping=/vaadin/*
 vaadin.endpoint.prefix=/connect
 # Pnpm doesn't work when the project is mounted and executed in a Docker container
-vaadin.pnpm.enable = false
+# vaadin.pnpm.enable = false
```
jouni, line 10: I assume we could just remove these instead of comment them.

vursen: Indeed, thanks! The PR isn't finished completely and needs some polish.

frontend/demo/component/icons/iconset-generator.ts
```diff
@@ -255,7 +255,7 @@ export class IconsetGenerator extends LitElement {
         enumName = 'Icons';
       }
 
-      const jsName = enumName.replaceAll(/([a-z])([A-Z]+)/g, '$1-$2').toLowerCase();
+      const jsName = enumName.replace(/([a-z])([A-Z]+)/g, '$1-$2').toLowerCase();
```
vursen, line 258: **note:** We had to replace `replaceAll` with `replace` here because the current TS config targets `ES2020`. If I change it to `ES2021`, it gets reverted to `ES2020` for some reason as soon as I start the Flow server.

webpack.config.js
```diff
@@ -1,69 +0,0 @@
...
-    // Polymer etc dependencies have purposefully been excluded from the
-    // docs project (Vaadin) bundle. However, the embedded Flow examples
-    // (like target/frontend/accordion-basic-wc.ts)
-    // now import "applyTheme" which has an indirect dependency to Polymer
-    // so we need to direct the applyTheme function to use the version
-    // bundled with docs-app. Otherwise we'd end up with conflicting imports
-    // (Vaadin/Polymer) originating from two separate bundles loaded on the same page.
-    // We'll do this by declaring "generated/theme" (generated by Vaadin to embedded components)
-    // as an external, which gets registered to the global namespace by docs-app bundle
-    // (in frontend/demo/example-resources.ts).
```
tomivirkki, line 37: This explanation could be included in the Vite config file also

vursen:
The heartbeat workaround fails to work locally from time to time. 

image

However, after a discussion with @tomivirkki, we decided to ignore that because it turns out not to work reliably even on the `latest` branch. Let's see how it works in production.

frontend/demo/init-flow-components.ts
```diff
@@ -12,6 +12,11 @@ import '@vaadin/flow-frontend/dndConnector-es6.js';
 import '@vaadin/flow-frontend/flow-component-renderer.js';
 
 // Flow component specific modules
+import '@vaadin/flow-frontend/confirmDialogConnector.js';
```
vursen, line 15: **note:** These imports were missing. We didn't catch that because JS errors seem to be suppressed in production.

tomivirkki (approved):

PR #1909, opened 2022-11-23
Add product icons for acceleration kits

MatthewVaadin:
Add some product icons for the SSO, Observability and Kubernetes Kits.


jouni (approved):

jouni:
Thanks! I was wondering about these yesterday, that when might we get icons for them.

PR #1915, opened 2022-11-24
refactor: migrate from Webpack to Vite

vursen:
It is another attempt to migrate to Vite. 

A follow-up to https://github.com/vaadin/docs/pull/1906


src/main/resources/application.properties
```diff
@@ -2,10 +2,8 @@ server.port=8080
 server.compression.enabled=true
 server.compression.mime-types=application/json,application/xml,text/html,text/xml,text/plain,application/javascript,text/css
 # Ensure application is run in Vaadin 14/npm mode
```
tomivirkki, line 4: These comments should also be removed since they describe the removed properties

vursen: Done.

tomivirkki (approved):

PR #1946, opened 2022-12-12
fix: fixed feature flag configuration example

mcollovati:
Example was missing the 'com.' prefix

PR #1971, opened 2022-12-19
Add missing prefix

Artur-:


mcollovati:
It seems a duplicate of #1946

Artur-:
So it seems

PR #1983, opened 2022-12-21
Uncaught exception bubbling.

oliveryasuna:


jouni:
Thanks for the contribution, @oliveryasuna!

@mshabarov, can you confirm that this limitation and statement is technically correct?

Content-wise, I don't understand what “Framework” means in this context, as “Flow” is also used.

oliveryasuna:
This PR can be closed without merging. I need to do more testing and will open another later.

PR #1994, opened 2022-12-28
chore: update vaadin 23.3.2

ZheSun88:


ZheSun88:
there will be 23.3.3 coming.

PR #2027, opened 2023-01-17
docs: update RTE value docs to reflect default format change (v24)

sissbruecker:
Updates the RichTextEditor "Value Format" section to reflect changes in v24 where the HTML format becomes the default for the Flow component.

Also updates the code examples to demonstrate usage of both formats.

This should only be merged after v24 has been released and docs have been upgraded.

### DIFF
--- a/frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
+++ b/frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
@@ -3,7 +3,11 @@ import 'Frontend/demo/init'; // hidden-source-line
 import { html, LitElement } from 'lit';
 import { customElement, query, state } from 'lit/decorators.js';
 import '@vaadin/rich-text-editor';
-import type { RichTextEditor, RichTextEditorChangeEvent } from '@vaadin/rich-text-editor';
+import type {
+  RichTextEditor,
+  RichTextEditorHtmlValueChangedEvent,
+  RichTextEditorValueChangedEvent,
+} from '@vaadin/rich-text-editor';
 import '@vaadin/text-area';
 import type { TextAreaChangeEvent } from '@vaadin/text-area';
 import { applyTheme } from 'Frontend/generated/theme';
@@ -19,6 +23,7 @@ export class Example extends LitElement {
 
   @state()
   private htmlValue = '';
+
   @state()
   private deltaValue = '';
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/vaadin/demo/component/richtexteditor/RichTextEditorSetGetValue.java
```diff
@@ -11,21 +11,32 @@ public class RichTextEditorSetGetValue extends Div {
...
+                "Enter something in the Rich Text Editor to see its Delta value here.");
+        deltaTextArea.setWidthFull();
+        rte.asDelta().addValueChangeListener(e -> deltaTextArea.setValue(e.getValue()));
+        deltaTextArea.addValueChangeListener(e -> {
+            if (!rte.asDelta().getValue().equals(e.getValue())) {
+                rte.asDelta().setValue(e.getValue());
+            }
+        });
+
+        add(rte, htmlTextArea);
```
tomivirkki, line 39: `deltaTextArea` should also be added

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
```diff
@@ -28,23 +30,39 @@ export class Example extends LitElement {
...
+        @change="${(e: TextAreaChangeEvent) => this.setHtmlValue(e.target.value)}"
+      ></vaadin-text-area>
+
+      <vaadin-text-area
+        label="Delta Value"
+        placeholder="Enter something in the Rich Text Editor to see its Delta value here."
+        style="width: 100%;"
+        .value="${this.deltaValue}"
+        @change="${(e: TextAreaChangeEvent) => {
+          this.deltaValue = e.target.value;
```
tomivirkki, line 54: Updating one of the format value fields doesn't affect the other one, only the RTE:

https://user-images.githubusercontent.com/1222264/218084206-e80d2382-230b-44fb-8aea-5dcf31f3775d.mp4

src/main/java/com/vaadin/demo/component/richtexteditor/RichTextEditorSetGetValue.java
sissbruecker: Thanks, fixed.

frontend/demo/component/richtexteditor/rich-text-editor-set-get-value.ts
sissbruecker: I think I ran into an infinite loop when trying this earlier. Trying it now it seems to work, so it's fixed.

tomivirkki (approved):

PR #2052, opened 2023-01-23
Hilla SSO Kit getting started

tamasmak:
Adding documentation about how to setup SSO Kit for Hilla.


jouni:
Language check to be done after merging, as this content is needed for an internal training session and hackathon.

PR #2072, opened 2023-02-01
Docs/v24 styling

rolfsmeds:
Styling tabs with CSS selector lists added to Components pages

### DIFF
--- a/dspublisher/theme/global.css
+++ b/dspublisher/theme/global.css
@@ -362,6 +362,20 @@ html[theme~='dark'] .card.card {
   background-color: inherit;
 }
 
+/* Class that removes the "box" styling of <code> blocks (for use in tables) */
+[class^='ArticleLayout'] .unboxed-code code {
+  background: transparent;
+  border: none;
+  margin: 0;
+  padding: 0;
+}
+
+/* Make <code> content in <dd>s break at non-word-chars rather than within words */
+[class^='ArticleLayout'] dd code {
+  overflow-wrap: normal;
+  word-break: normal;
+}
+
 /*
  * Work around a Chromium + MacOS issue with the scrollbar appearing on the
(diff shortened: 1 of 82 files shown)


jouni (approved):
Love the `+++<wbr>+++` parts 😄

jouni:
Oh, now I noticed you used the horizontal description list from AsciiDoc. That’s not actually needed, as the default styling in DSP is horizontal, and it automatically wraps to a column on small viewports. But the built-in horizontal style in AsciiDoc renders it as a table, and the DSP styling doesn't apply to that. So I would want to convert these pages to use the default description lists instead.

rolfsmeds:
It does actually wrap them to a column... but I now see that they seem to render identically without the `[Horizontal]` too, so no point in having it there. I'll do a PR.

jouni:
> It does actually wrap them to a column

Oh, really? That’s interesting, I need to try and see for myself. I don't remember implementing any styles for that, so I wonder how the table columns get wrapped…

dspublisher/theme/global.css
```diff
@@ -362,6 +362,20 @@ html[theme~='dark'] .card.card {
...
+  border: none;
+  margin: 0;
+  padding: 0;
+}
+
+/* Make <code> content in <dd>s break at non-word-chars rather than within words */
+[class^='ArticleLayout'] dd code {
+  overflow-wrap: normal;
+  word-break: normal;
+}
```
jouni, line 377: These styles wouldn't work with the horizontal description lists, as those get rendered inside table elements.

rolfsmeds:
Hmm, maybe it's the horizontal tag that doesn't have any effect for me locally? Because they seem to render 100% identically for me with and without that tag.

Anyhoo, doesn't matter, already merged removal of the tags.

PR #2074, opened 2023-02-01
Fix broken image

Mikaelsu:
The path `https://hilla.dev/docs/react/guides/images/codecompletion.gif` did not exist (the images are hosted in the `lit` folder) so I copied the `.gif` to the missing location. I didn't find any other broken images.


jouni (approved):
Would be nicer to be able to reference images using an absolute path, but I’m not sure if that works at the moment.

PR #2085, opened 2023-02-02
chore: update charts to set additionalOptions using JS

web-padawan:
## Description

This PR fixes the following warning from the `lit/attribute-value-entities` rule by `eslint-plugin-lit`:

```
/Users/serhii/vaadin/docs/frontend/demo/component/charts/charts-overview.ts
  110:9  error  Attribute values may not contain unencoded HTML entities, e.g. use `&gt;` instead of `>`
```
  
Also, we generally should  use TypeScript instead of JSON objects to HTML attributes 🙈

### DIFF
--- a/frontend/demo/component/charts/charts-overview.ts
+++ b/frontend/demo/component/charts/charts-overview.ts
@@ -60,6 +60,22 @@ export class Example extends LitElement {
   private columnOptions: Options = { yAxis: { title: { text: '' } } };
 
   @state()
+  private months = [
+    'Jan',
+    'Feb',
+    'Mar',
+    'Apr',
+    'May',
+    'Jun',
+    'Jul',
+    'Aug',
+    'Sep',
+    'Oct',
+    'Nov',
+    'Dec',
+  ];
+
+  @state()
   private pieOptions: Options = {
     tooltip: {
       pointFormat: '{series.name}: <b>{point.percentage:.1f}%</b>',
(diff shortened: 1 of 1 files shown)


frontend/demo/component/charts/charts-overview.ts
```diff
@@ -42,8 +43,60 @@ export class Example extends LitElement {
     }
   `;
 
+  @state()
+  private areaOptions: Options = {
```
vursen, line 47: I noticed that these objects were inside the tag snippet previously. I wonder, would they still be worth exposing to the user?

web-padawan: The purpose of this particular demo isn't to showcase usage of `additionalOptions`, but the `theme` switching.
Implementing proper TS / Java examples for charts is a separate topic, we should have an issue for that.

Related ticket about not using JSON in HTML attributes: https://github.com/vaadin/web-components/issues/1670

vursen: OK, makes sense. Should this also be extracted to a state property then?

https://github.com/vaadin/docs/blob/5a7feec7945e985ec7eb872ac0bdf5a247f9b248/frontend/demo/component/charts/charts-overview.ts#L146-L165

web-padawan: Done.

frontend/demo/component/charts/charts-overview.ts
```diff
@@ -52,133 +129,55 @@ export class Example extends LitElement {
   render() {
     return html`
       <vaadin-chart
-        id="vaadin-chart-1"
         type="column"
         categories='["Jan", "Feb", "Mar"]'
```
vursen, line 133:
```suggestion
        categories="${['Jan', 'Feb', 'Mar']}"
```

vursen (approved):

PR #2089, opened 2023-02-03
chore: fix remaining errors from eslint-config-vaadin

web-padawan:
## Description

Fixed more warnings. Now we should be ready to enable the config in a separate PR, with some rules still disabled.

### DIFF
--- a/frontend/demo/upgrade-tool/upgrade-tool.ts
+++ b/frontend/demo/upgrade-tool/upgrade-tool.ts
@@ -180,14 +180,20 @@ export default class UpgradeTool extends LitElement {
     this.updateUrlParameters();
   }
 
+  private hideMatchingElements(types: string[]) {
+    const selector = types.map((type) => `[class*="${type}"]`).join(', ');
+
+    document.querySelectorAll<HTMLElement>(selector).forEach((elem) => {
+      this.setElementVisible(elem, false);
+    });
+  }
+
   private perform14To23CleanupIfNecessary() {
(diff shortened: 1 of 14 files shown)


frontend/demo/upgrade-tool/upgrade-tool.ts
```diff
@@ -224,15 +224,15 @@ export default class UpgradeTool extends LitElement {
   }
 
   private getElementsByClassname(classname: string) {
-    return <HTMLElement[]>[...document.querySelectorAll(`.${classname}`)];
+    return [...document.querySelectorAll(`.${classname}`)] as HTMLElement[];
```
vursen, line 227:
```suggestion
    return [...document.querySelectorAll<HTMLElement>(`.${classname}`)];
```

frontend/demo/upgrade-tool/upgrade-tool.ts
```diff
@@ -187,7 +187,7 @@ export default class UpgradeTool extends LitElement {
 
     document
       .querySelectorAll("[class*='all'], [class*='spring']")
-      .forEach((elem) => this.setElementVisible(<HTMLElement>elem, false));
+      .forEach((elem) => this.setElementVisible(elem as HTMLElement, false));
```
vursen, line 190: You could pass `HTMLElement` as to the generic `querySelectorAll` method.

web-padawan: Refactored and moved this code to a separate method.

PR #2090, opened 2023-02-03
chore: update to eslint-config-vaadin, use latest husky

web-padawan:
## Description

1. Updated to use `eslint-config-vaadin` with some rules still disabled for now,
2. Upgraded `husky` and `lint-staged` to align with the [web-components](https://github.com/vaadin/web-components/blob/dc5d2d316e50eedf7b2746df7c03843712143a61/package.json#L51-L53) repo.


package.json
```diff
@@ -1,4 +1,7 @@
 {
+  "name": "docs",
+  "license": "UNLICENSED",
+  "private": true,
```
web-padawan, line 4: Moved these to the top and added `"private": true` to indicate the repo isn't published to npm.

package.json
```diff
@@ -137,19 +140,16 @@
     "@rollup/plugin-replace": "3.1.0",
     "@rollup/pluginutils": "4.1.0",
     "@types/iframe-resizer": "^3.5.9",
-    "@typescript-eslint/eslint-plugin": "^5.39.0",
-    "@typescript-eslint/parser": "^5.39.0",
```
web-padawan, line 141: These are dependencies of `eslint-config-vaadin`, no need to list them here.

package.json
```diff
@@ -137,19 +140,16 @@
...
     "async": "3.2.2",
     "commander": "^9.4.1",
-    "eslint": "^8.24.0",
-    "eslint-config-prettier": "^8.5.0",
-    "eslint-plugin-lit": "^1.6.1",
```
web-padawan, line 146: Sane as above: these are also dependencies of `eslint-config-vaadin`.

package.json
```diff
@@ -160,24 +160,18 @@
...
-      "pre-commit": "lint-staged"
-    }
-  },
-  "license": "UNLICENSED",
   "lint-staged": {
     "frontend/**/*.{js,ts,json}": [
-      "prettier --write",
-      "npm run lint"
+      "eslint --fix",
+      "prettier --write"
```
vursen, line 166: Actually, I'm not sure that we need `prettier` on this list as Prettier is already integrated into our ESLint config.

web-padawan: This is how `lint-staged` recommends it - [example](https://github.com/okonet/lint-staged#running-multiple-commands-in-a-sequence). Maybe we can still keep it for transparency.

vursen (approved):

PR #2094, opened 2023-02-04
chore: use TypeScript override keyword for LitElement methods

web-padawan:
## Description

Updated TS examples and other code (upgrade tool etc) to use `protected override` for the following methods:

- [`createRenderRoot()`](https://github.com/lit/lit/blob/7c93499ccdfc493df9397163e552356e64bfd2c3/packages/lit-element/src/lit-element.ts#L138) - already overridden in `LitElement`
- [`render()`](https://github.com/lit/lit/blob/7c93499ccdfc493df9397163e552356e64bfd2c3/packages/lit-element/src/lit-element.ts#L224) - defined in `LitElement`
- [`firstUpdated()`](https://github.com/lit/lit/blob/7c93499ccdfc493df9397163e552356e64bfd2c3/packages/reactive-element/src/reactive-element.ts#L1512) - defined in `ReactiveElement`

This is how e.g. [Material Web Components](https://github.com/material-components/material-web/blob/aef3ee3d3fcd7887cd3169eb80cddd42e89df8b3/fab/lib/fab-shared.ts#L44) use these methods. Also, it makes code more self-explanatory.
Finally, this gets us a bit closer to enabling `@typescript-eslint/member-ordering` rule in the future.

See also related section in the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag).

### DIFF
--- a/frontend/demo/component/accordion/accordion-basic.ts
+++ b/frontend/demo/component/accordion/accordion-basic.ts
@@ -9,14 +9,14 @@ import { applyTheme } from 'Frontend/generated/theme';
 
 @customElement('accordion-basic')
 export class Example extends LitElement {
-  protected createRenderRoot() {
+  protected override createRenderRoot() {
     const root = super.createRenderRoot();
     // Apply custom theme (only supported if your app uses one)
     applyTheme(root);
     return root;
   }
 
-  render() {
+  protected override render() {
     return html`
       <!-- tag::snippet[] -->
       <vaadin-accordion>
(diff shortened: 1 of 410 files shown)


tomivirkki:
Should `override` be used for [styles](https://github.com/material-components/material-web/blob/aef3ee3d3fcd7887cd3169eb80cddd42e89df8b3/button/filled-button.ts#L27) also?

web-padawan:
@tomivirkki Good point, thanks. Done.

tomivirkki (approved):

PR #2097, opened 2023-02-06
feat: add docs for NonNullApi and nullability annotation collision

Lodin:
Resolves vaadin/hilla#641

### DIFF
--- a/articles/lit/reference/type-nullability.adoc
+++ b/articles/lit/reference/type-nullability.adoc
@@ -159,6 +159,8 @@ package com.example.application;
 public class MyBean {
     public List<String> list;
     public Map<String, Integer> map;
+    @Nullable
+    public String nullable;
 }
 ----
 
@@ -168,48 +170,7 @@ public class MyBean {
 export default interface MyBean {
   list: Array<string>;
   map: Record<string, number>;
-}
-----
-
-== Nullability collision
-
-There could be a case where several annotations are applied to the same type.
-To resolve this collision, each annotation has its priority; among all applied annotations the one with the highest priority is selected.
-
-* `@NonNullApi` annotation has the lowest priority.
-This annotation applies only to types without any nullability annotation.
-* `@Nullable` annotation has intermediate priority.
-It makes any type that affected by `@NonNullApi` annotation nullable.
-* `@Nonnull` annotation has the highest priority.
-It makes any type non-nullable again.
-
-.`package-info.java`
-[source,java]
-----
-@NonNullApi
-package com.example.application;
-----
-
-.`MyBean.java`
-[source,java]
-----
-public class MyBean {
-    public List<String> list;
-    @Nullable
-    public Map<String, @Nullable Integer> nullableMap;
-    @Nonnull
-    @Nullable
-    public Short forcedNonNullNumber;
-}
-----
-
-.`MyBean.ts`
-[source,typescript]
-----
-export default interface MyBean {
-  list: Array<string>;
-  map: Record<string, number | undefined>;
-  forcedNonNullNumber: number;
+  nullable?: string;
 }
 ----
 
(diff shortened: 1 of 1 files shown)


articles/lit/reference/type-nullability.adoc
```diff
@@ -134,4 +134,83 @@ export default interface MyBean {
...
+
+There could be a case where several annotations are applied to the same type.
+To resolve this collision, each annotation has its priority; among all applied annotations the one with the highest priority is selected.
+
+* `@NonNullApi` annotation has priority of `10` which makes it the weakest.
+This annotation applies only to types without any nullability annotation.
+* `@Nullable` annotation has priority of `20`.
+It makes any type that affected by `@NonNullApi` annotation nullable.
+* `@Nonnull` annotation has priority of `30` which makes it the strongest.
+It makes any type non-nullable again.
```
cromoteca, line 184: I'd not disclose our priority numbers, they don't matter in the documentation, they are just an implementation detail. Saying the order is enough.

cromoteca (changes requested):

articles/lit/reference/type-nullability.adoc
```diff
@@ -134,4 +134,83 @@ export default interface MyBean {
...
+
+== Nullability collision
+
+There could be a case where several annotations are applied to the same type.
+To resolve this collision, each annotation has its priority; among all applied annotations the one with the highest priority is selected.
+
+* `@NonNullApi` annotation has the lowest priority.
+This annotation applies only to types without any nullability annotation.
+* `@Nullable` annotation has intermediate priority.
+It makes any type that affected by `@NonNullApi` annotation nullable.
```
cromoteca, line 182: This sentence is not clear

articles/lit/reference/type-nullability.adoc
```diff
@@ -134,4 +134,83 @@ export default interface MyBean {
...
+* `@Nonnull` annotation has the highest priority.
+It makes any type non-nullable again.
```
cromoteca, line 184: I can't figure out the "again" case

articles/lit/reference/type-nullability.adoc
```diff
@@ -134,4 +134,83 @@ export default interface MyBean {
...
+.`MyBean.java`
+[source,java]
+----
+public class MyBean {
+    public List<String> list;
+    @Nullable
+    public Map<String, @Nullable Integer> nullableMap;
+    @Nonnull
+    @Nullable
+    public Short forcedNonNullNumber;
```
cromoteca, line 202: While this is useful to explain priorities, I'd not show it as it makes no sense to put both annotations on the same field. Or at least I'd add a comment to clarify that it is just to show the priority.

cromoteca (changes requested):

articles/lit/reference/type-nullability.adoc
Lodin: Hmm... Would it be more clear this way?
```
It makes any nullable type (either default or made nullable by `@Nullable` annotation) non-nullable. 
```
or it's better to make it as simple as possible? 
```
It makes any nullable type non-nullable.
```

cromoteca: I prefer the latter

articles/lit/reference/type-nullability.adoc
cromoteca: Reading it again, I'm not sure that the whole collision section is needed: you already clarified at line 147 how to restore nullability under `NonNullApi`, and conflicts between `Nullable` and `Nonnull` are unlikely to happen.

Lodin: Hmm. You're right. I guess I just followed the implementation details.

cromoteca (approved):

platosha (approved):

PR #2103, opened 2023-02-08
chore: update MPR docs for Vaadin 24

mcollovati:
Fixed configuration snippets to use Vaadin 7 and 8 dependencies variants that work with Vaadin 24.
Added Flow, Vaadin 7 and 8 versions to `_vaadin-version.adoc` to avoid repeating values multiple times.
Updated version-update.sh to only replace `vaadin-version` value.

Related to vaadin/multiplatform-runtime-internal#675, vaadin/multiplatform-runtime-internal#676


jouni:
Seems like this should target the `next` branch instead, right?

articles/tools/mpr/introduction/1-maven-v8.asciidoc
```diff
@@ -43,14 +43,14 @@ MPR is part of Vaadin, so the supported version of the project is set for you wh
 
 === Framework 8 dependency
 
-When using MPR the minimum requirement for Vaadin 8 version is 8.6.0 or newer:
+When using MPR with Vaadin 24+ the minimum requirement for Vaadin 8 version is 8.18.0 or newer:
```
jouni, line 46: Avoid explicit Vaadin versions in the written text. Since this is the documentation for Vaadin 24, there’s no need to call that out, it’s already in the context.

```suggestion
When using MPR the minimum requirement for Vaadin 8 version is 8.18.0 or newer:
```

articles/tools/mpr/introduction/3-cdi.asciidoc
```diff
@@ -11,7 +11,7 @@ This step is needed in case your Vaadin 7 or 8 application uses CDI. If it isn't
 
 == Updating to the correct CDI version
 
-Remove any version from `com.vaadin:vaadin-cdi` as the proper Vaadin 14 compatible version for it's managed by the `vaadin-bom`:
+Remove any version from `com.vaadin:vaadin-cdi` as the proper Vaadin 24 compatible version for it's managed by the `vaadin-bom`:
```
jouni, line 14:
```suggestion
Remove any version from `com.vaadin:vaadin-cdi` as the compatible version for it is managed by the `vaadin-bom`:
```

articles/tools/mpr/introduction/1-maven-v7.asciidoc
```diff
@@ -43,14 +43,14 @@ MPR is part of Vaadin, so the supported version of the project is set for you wh
 
 === Framework 7 dependency
 
-When using MPR the minimum requirement for Vaadin 7 version is 7.7.14 or newer.
+When using MPR with Vaadin 24+ the minimum requirement for Vaadin 7 version is 7.7.36 or newer.
```
jouni, line 46:
```suggestion
When using MPR the minimum requirement for Vaadin 7 version is 7.7.36 or newer.
```

mcollovati:
> Seems like this should target the `next` branch instead, right?

Yes, my bad, sorry. I'll rebase on the correct branch

mcollovati:
Needs additional documentation on how to completely exclude `vaadin-server` artifact.

This also involves updates on the 23 to 24 migration guide.
@jouni should the migration guide be updated in another PR with target `latest` and the picked to `next`?

jouni:
> should the migration guide be updated in another PR with target latest and the picked to next?

The 23→24 upgrade guide was removed from `latest`, and is now only in the `next` branch.

mcollovati:
> The 23→24 upgrade guide was removed from `latest`, and is now only in the `next` branch.

It seems like the opposite. I can see the guide on `latest`, but not on `next`

jouni:
Right. But that’s only the situation in production, as the builds have been failing yesterday and today. But if you look at the repo branches:

- https://github.com/vaadin/docs/blob/latest/articles/upgrading/V23-V24-steps/index.adoc (the content says the page has moved to the v24 docs)
- https://github.com/vaadin/docs/blob/next/articles/upgrading/index.adoc (this is the actual guide)

PR #2119, opened 2023-02-10
docs: add missing upgrade instructions for date picker

sissbruecker:
Adds missing instructions for migrating date picker theme. Also fixes existing instructions to refer to month calendar CSS file.


rolfsmeds (approved):
Had totally missed these, thanks!

PR #2126, opened 2023-02-13
as was, it would only print "hello"

frankielc:
Changed code so that user can now print "hello ${name}"


jouni:
Thanks for the PR! I think change makes the connection between the client and the server/Java parts more clear.

jouni (approved):

PR #2132, opened 2023-02-13
Add documentation for configuring pro key for GitHub actions

TatuLund:


jouni:
Thanks for the change. But I believe we don't want to encourage the use of the `proKey` file. Developers should use the `offlineKey` file instead, which is already described in the section before this one you wanted to add.

Which problem is this change address? Is the main idea to highlight GitHub Actions explicitly?

TatuLund:
> Which problem is this change address? Is the main idea to highlight GitHub Actions explicitly?

GitHub actions is a good example. But in general this applies to any build server where the machine id is not stable, and has access to network i.e. license server.

jouni:
> But in general this applies to any build server where the machine id is not stable, and has access to network i.e. license server.

The server license key does not rely on a machine ID, and is essentially the same as an offline key, so it doesn't require network access.

> GitHub actions is a good example

I agree on that. Perhaps that could be a separate tutorial in the blog.

PR #2133, opened 2023-02-13
docs: update grid cell focus example to use global styles

sissbruecker:
Updates the grid cell focus example to use the new theming mechanism.

### DIFF
--- a/frontend/themes/docs/grid-cell-focus.css
+++ b/frontend/themes/docs/grid-cell-focus.css
@@ -1,6 +1,7 @@
-/* frontend/theme/[my-theme]/components/vaadin-grid.css */
+/* Add this to your global CSS, for example in: */
+/* frontend/theme/[my-theme]/styles.css */
 
-:host([theme~='force-focus-outline']) [part~='cell']:focus::before {
+vaadin-grid.force-focus-outline::part(cell):focus::before {
     content: '';
     position: absolute;
     top: 0;
--- a/frontend/themes/docs/grid-cell-focus.css
+++ b/frontend/themes/docs/grid-cell-focus.css
@@ -1,7 +1,7 @@
 /* Add this to your global CSS, for example in: */
 /* frontend/theme/[my-theme]/styles.css */
 
-vaadin-grid.force-focus-outline::part(cell):focus::before {
+vaadin-grid.force-focus-outline::part(focused-cell)::before {
     content: '';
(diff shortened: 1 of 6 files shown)


frontend/themes/docs/grid-cell-focus.css
```diff
@@ -1,6 +1,7 @@
-/* frontend/theme/[my-theme]/components/vaadin-grid.css */
+/* Add this to your global CSS, for example in: */
+/* frontend/theme/[my-theme]/styles.css */
 
-:host([theme~='force-focus-outline']) [part~='cell']:focus::before {
+vaadin-grid.force-focus-outline::part(cell):focus::before {
```
web-padawan, line 4: Let's use `::part(focused-cell):focus` instead as we now have this "cell button" mode for VoiceOver on Mac OS.

web-padawan (changes requested):

frontend/themes/docs/grid-cell-focus.css
sissbruecker: Thanks, though it should be `::part(focused-cell)::before`, correct?

web-padawan: Yes, indeed 👍

sissbruecker: Done

web-padawan (approved):

frontend/themes/docs/grid-cell-focus.css
```diff
@@ -1,6 +1,7 @@
-/* frontend/theme/[my-theme]/components/vaadin-grid.css */
+/* Add this to your global CSS, for example in: */
+/* frontend/theme/[my-theme]/styles.css */
 
-:host([theme~='force-focus-outline']) [part~='cell']:focus::before {
+vaadin-grid.force-focus-outline::part(focused-cell)::before {
```
vursen, line 4: Why do we prefer classes in this case? To me, it feels like a theme.

vursen: Well, perhaps we can use classes if this customization only applies to this particular example.

web-padawan: Yes, from now on we should use CSS classes for any individual customizations like this one.

vursen (approved):

PR #2134, opened 2023-02-13
docs: explain multiple fields in form layouts

### LINKED ISSUE #1005: [form-layout] Add an example of using custom-field for multiple fields inside a form-item
Placing multiple fields directly to a form-item is deprecated (https://github.com/vaadin/web-components/pull/3221). Using a custom-field is recommended instead. 

It would make sense if we had an example in the form-layout docs that showed how fields have to be wrapped with a custom-field when there is a need for multiple fields inside a single form-item.

vursen:
Added a section explaining how to use multiple fields in form layouts.

- [x] TS example
- [x] Java example

Fixes #1005

### DIFF
--- a/articles/components/form-layout/index.asciidoc
+++ b/articles/components/form-layout/index.asciidoc
@@ -135,6 +135,26 @@ include::{root}/src/main/java/com/vaadin/demo/component/formlayout/FormLayoutNat
 ----
 --
 
+== Multiple Fields
+
+Form Item only supports placing a single field inside. In the case you need to place multiple fields, <<../custom-field#,Custom Field>> should be used as a wrapper:
+
+[.example]
+--
+
+[source,js]
+----
+include::{root}/frontend/demo/component/formlayout/form-layout-custom-field.ts[render,tags=snippet,indent=0,group=TypeScript]
+----
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCustomField.java[render,tags=snippet,indent=0,group=Java]
+----
+--
+
+However, keeping fields in individual Form Items is preferable. Wrapped fields can be hard to visually distinguish since they usually have no individual label except for a placeholder which is only visible when the field has no value.
+
 == Best Practices
 
 === Sectioning
--- a/articles/components/form-layout/index.asciidoc
+++ b/articles/components/form-layout/index.asciidoc
@@ -137,7 +137,7 @@ include::{root}/src/main/java/com/vaadin/demo/component/formlayout/FormLayoutNat
 
 == Multiple Fields
 
-Form Item only supports placing a single field inside. In the case you need to place multiple fields, <<../custom-field#,Custom Field>> should be used as a wrapper:
+Form Item only supports placing a single field inside. Where you need to place multiple fields, <<../custom-field#,Custom Field>> should be used as a wrapper:
 
 [.example]
 --
@@ -153,7 +153,7 @@ include::{root}/src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCus
 ----
 --
 
-However, keeping fields in individual Form Items is preferable. Wrapped fields can be hard to visually distinguish since they usually have no individual label except for a placeholder which is only visible when the field has no value.
+Keeping fields in individual Form Items, however, is preferable. Wrapped fields can be hard to distinguish visually since they usually have no individual label except for a placeholder, which is only visible when the field has no value.
 
 == Best Practices
 
--- a/frontend/demo/component/formlayout/form-layout-custom-field.ts
+++ b/frontend/demo/component/formlayout/form-layout-custom-field.ts
@@ -1,0 +1,62 @@
+import 'Frontend/demo/init'; // hidden-source-line
+
+import { html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+import '@vaadin/form-layout';
+import '@vaadin/form-layout/vaadin-form-item';
+import '@vaadin/select';
+import '@vaadin/custom-field';
+import '@vaadin/horizontal-layout';
+
+import { applyTheme } from 'Frontend/generated/theme';
+
+@customElement('form-layout-custom-field')
+export class Example extends LitElement {
+  protected createRenderRoot() {
+    const root = super.createRenderRoot();
+    // Apply custom theme (only supported if your app uses one)
+    applyTheme(root);
+    return root;
+  }
+
+  // tag::snippet[]
+  get months() {
+    return Array.from({ length: 12 }, (_, i) => `${i + 1}`.padStart(2, '0'));
+  }
+
+  get years() {
+    return Array.from({ length: 11 }, (_, i) => `${i + new Date().getFullYear()}`);
+  }
+
+  render() {
+    return html`
+      <vaadin-form-layout>
+        <vaadin-form-item>
+          <label slot="label">Expiration</label>
+          <vaadin-custom-field
+            .parseValue="${(value: string) => {
+              return value ? value.split('/') : ['', ''];
+            }}"
+            .formatValue="${(values: unknown[]) => {
+              return values[0] && values[1] ? values.join('/') : '';
+            }}"
+          >
+            <vaadin-horizontal-layout theme="spacing-xs">
+              <vaadin-select
+                title="Month"
+                placeholder="Month"
+                .items="${this.months.map((month) => ({ label: month, value: month }))}"
+              ></vaadin-select>
+              <vaadin-select
+                title="Year"
+                placeholder="Year"
+                .items="${this.years.map((year) => ({ label: year, value: year }))}"
+              ></vaadin-select>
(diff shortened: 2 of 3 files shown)


articles/components/form-layout/index.asciidoc
```diff
@@ -145,6 +145,25 @@ include::{root}/src/main/java/com/vaadin/demo/component/formlayout/FormLayoutNat
 ----
 --
 
+=== Multiple Fields
```
vursen, line 148: It is currently a sub-section of "Native Input Fields" and therefore doesn't show up in the menu. Should we make it a top-level section?

sissbruecker: I think it would make more sense as a top-level section then, as this doesn't really have anything to do with native inputs.

frontend/demo/component/formlayout/form-layout-custom-field.ts
```diff
@@ -0,0 +1,47 @@
...
+              <vaadin-text-field
+                title="Postal code"
+                placeholder="Postal code"
+                autocomplete="postal-code"
+              ></vaadin-text-field>
+              <vaadin-text-field
+                title="Street address"
+                placeholder="Street address"
+                autocomplete="street-address"
+              ></vaadin-text-field>
```
sissbruecker, line 39: Postal code and street is an odd combination IMO. How about going with the suggestion from Knoobie to use street name + street number?

articles/components/form-layout/index.asciidoc
vursen: I personally feel that Multiple Fields is alright.

frontend/demo/component/formlayout/form-layout-custom-field.ts
sissbruecker, line 39: Side note: I can see people getting confused how just dumping individual fields into the custom field will result in a combined value. Can we make that more explicit? Or at least add a comment explaining how the combined value is generated in this case? That would also be useful in the custom-field docs.

articles/components/form-layout/index.asciidoc
sissbruecker: Agree

src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCustomField.java
```diff
@@ -0,0 +1,80 @@
...
+            postalCode.setValue(address.getPostalCode());
+            street.setValue(address.getStreet());
+        }
+    }
+
+    public FormLayoutCustomField() {
+        FormLayout formLayout = new FormLayout();
+        // tag::snippet[]
+        formLayout.addFormItem(new AddressField(), "Address");
+        // end::snippet[]
```
sissbruecker, line 74: We had several complaints in the past how just showing a single line as a snippet can be very confusing. An example should include all the relevant parts, in this case: the entity, the custom field implementation, and constructing the form. The entity could be extracted into a separate file maybe.

src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCustomField.java
```diff
@@ -0,0 +1,80 @@
...
+        public AddressField() {
+            VerticalLayout layout = new VerticalLayout();
+            layout.setPadding(false);
+            layout.setSpacing(false);
+            layout.getThemeList().add("spacing-xs");
+
+            postalCode = new TextField();
+            postalCode.setPlaceholder("Postal code");
+            postalCode.getElement().setAttribute("title", "Postal code");
+            postalCode.getElement().setAttribute("autocomplete", "postal-code");
```
sissbruecker, line 46: Double checking if this is enough to address a11y concerns?

vursen: - `title` is needed for a11y 
- `autocomplete` is generally recommended to make filling in the address easier (https://web.dev/learn/forms/address/#help-users-enter-their-address) but it is not a must-have. I can remove it

src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCustomField.java
vursen: Good point, thanks! Let me update that.

frontend/demo/component/formlayout/form-layout-custom-field.ts
sissbruecker: Expanding further on this, if we would use street name + number, then the value type for the Flow example could just be a string, rather than an entity. 

Likewise, the current output of the web component example is a string like `"99999 Fake street 123"`. The example would also benefit if it would just combine street name and number into a string.

src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCustomField.java
sissbruecker: OK thanks 👍

frontend/demo/component/formlayout/form-layout-custom-field.ts
vursen: > The example would also benefit if it would just combine street name and number into a string.

I was also thinking about that, regardless of whether we use a postal code or street number. I was wondering how you would reliably parse such strings. You have to make some assumptions e.g. that the street number never has spaces. 

There is also another argument that it is bad to split addresses into several fields because it kind of slows down the filling of the form: you have to care about where to put which part of the address.

sissbruecker: Yeah, those are valid points 🙂. Still better than zip code + street, but if we can come up with something better, then let's go for it.

sissbruecker: Phone number maybe? Regarding parsing and formatting I think it's fine not to overthink this, the point is to show how to use custom field in a form item, not a perfect implementation of a custom value type.

vursen: A phone field could be an option. I disregarded it in the first place because it seemed too natural to be a custom field. I 
intentionally wanted to show a custom field that doesn't have a combined value. Based on the internal feedback in Slack, for some people, it wasn't clear how to create a Custom Field from "relatively" independent fields like "Expiration: Month / Year".  Perhaps the expiration field could be a good alternative to address.

vursen: > Based on the internal feedback in Slack, there are cases when it isn't clear-cut whether we are talking about a single field or several.

Well, this reasoning can be applied to any custom field. I think we could use the expiration field that would have the following string representation: "mm/yyyy".

sissbruecker: Sure, sounds good

vursen: Done.

src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCustomField.java
vursen: Done.

frontend/demo/component/formlayout/form-layout-custom-field.ts
vursen: Done.

articles/components/form-layout/index.asciidoc
```diff
@@ -145,6 +145,25 @@ include::{root}/src/main/java/com/vaadin/demo/component/formlayout/FormLayoutNat
...
+include::{root}/frontend/demo/component/formlayout/form-layout-custom-field.ts[render,tags=snippet,indent=0,group=TypeScript]
+----
+
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/formlayout/FormLayoutCustomField.java[render,tags=snippet,indent=0,group=Java]
+----
+--
+
+However, keeping fields in individual Form Items is preferable. Wrapped fields can be hard to visually distinguish since they usually have no individual label except for a placeholder which is only visible when the field has no value.
```
vursen, line 166:
```suggestion
However, keeping fields in individual Form Items is preferable. Wrapped fields can be hard to visually distinguish since they usually have no individual label except for a placeholder which is only visible as long as the field doesn't have a value.
```

sissbruecker (approved):

frontend/demo/component/formlayout/form-layout-custom-field.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  @state()
+  private months = Array.from({ length: 12 }, (_, i) => `${i + 1}`.padStart(2, '0'));
+
+  @state()
+  private years = Array.from({ length: 11 }, (_, i) => `${i + new Date().getFullYear()}`);
+
+  protected override firstUpdated() {
+    // Set title for screen readers
+    this.monthField.focusElement!.setAttribute('title', 'Month');
```
vursen, line 38: **note:** After we fix https://github.com/vaadin/web-components/issues/5580, we can use `aria-label` here.

sissbruecker (approved):

PR #2135, opened 2023-02-13
fix: Add notes on which order event handlers are called

### LINKED ISSUE #1947: Unclear in which order route lifecycle methods are called
From a comment in https://vaadin.com/docs/latest/routing/lifecycle

> Regarding BeforeEnterEvent : this page mentions BeforeEnterObserver and BeforeEnterListener, but this event is also passed to HasUrlParameter.setParameter() which is not mentioned, so it's not clear in which order the three interface methods are called.

caalador:
Fixes #1947


mcollovati (approved):

jouni (approved):
I don't really like the extended use of admonition blocks, but the information is valuable to get on the page, so let’s go with it for now. We can clean it up later.

mshabarov (approved):

PR #2136, opened 2023-02-13
docs: ported and reworked Customizing System Messages documentation

### LINKED ISSUE #1699: V23 docs missing CustomizedSystemMessages
The possibility of customizing system messages is not available in the V14 or V23 documentation. For V8, the instructions are found here: https://vaadin.com/docs/v8/framework/application/application-errors/#application.errors.systemmessages ; the functionality hasn't changed to my knowledge.

mcollovati:
Documentation for Customizing System Messages was present in Vaadin 8 but not ported to Flow.

Fixes #1699

### DIFF
--- a/articles/advanced/customize-system-messages.adoc
+++ b/articles/advanced/customize-system-messages.adoc
@@ -7,21 +7,21 @@ layout: page
 
 = Configuring Customizing System Messages
 
-System messages are notifications that indicates a major invalid state that may require restarting the application.
-The invalid states handled by system messages are session expiration, internal error and cookie disabled.
+System messages are notifications that indicate a major invalid state that may require restarting the application.
+The different types of invalid state handled by system messages are the session expired, internal error, and cookie disabled states.
 
-When the client falls into one of the invalid states, a notification is shown in the browser and, after the user clicks on it, the page is reloaded.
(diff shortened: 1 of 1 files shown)


caalador (commented):
Only thing that I would perhaps like to know that is not specifically mentioned is the default status for the different messages.
`sessionExpired` is disabled by default, where as `internalError` and `cookiesDisabled` are enabled.
`CustomizedSystemMessages` doesn't say the default state in the Javadoc either

mcollovati:
> `CustomizedSystemMessages` doesn't say the default state in the Javadoc either

Defaults are documented int `SystemMessages` Javadoc.
But it absolutely makes sense to list them also on documentation

caalador:
I did note that statement, but it doesn't directly connect to `sessionExpiredNotificationEnabled` so it would be good to have the different enabled defaults written out in the document.
Also on the `SystemMessages` I only looked at the getter and setter javadoc and those do not say the default.

mcollovati:
> Also on the `SystemMessages` I only looked at the getter and setter javadoc and those do not say the default.

Good point.
Getter/Setter Javadocs are the first ones that a developer will look at. It would be good to update them

caalador (approved):

PR #2149, opened 2023-02-14
Added a note about serialization

### LINKED ISSUE #2109: Mention in Spreadsheet documentation that it is not serializable

Mikaelsu:
Spreadsheet was missing a note about it not being serializable (closes #2109)


articles/components/spreadsheet/index.asciidoc
```diff
@@ -262,9 +262,9 @@ include::{root}/src/main/java/com/vaadin/demo/component/spreadsheet/SpreadsheetR
 == Limitations
 * No provided toolbars, menus, or other controls for formatting cells.
 * Limited support for the older XSL format.
-* Limitations of Apache POI.
```
sissbruecker, line 265: It makes sense to keep this as this applies to lots of other aspects of the component, such as importing / exporting Excel files.

articles/components/spreadsheet/index.asciidoc
```diff
@@ -262,9 +262,9 @@ include::{root}/src/main/java/com/vaadin/demo/component/spreadsheet/SpreadsheetR
...
 * Using a table as a named range in formulas is not supported.
 As a workaround, you can manually convert tables to ranges. Open the spreadsheet in Excel, right-click on the table, select the Convert to Range action, and then save the spreadsheet.
+* Serialization is not supported by Apache POI.
```
sissbruecker, line 267: This could be a bit more verbose, as this doesn't imply that the component itself is not serializable, or why that would be a problem. I think this prevents using `@PreserveOnRefresh`, and probably session replication with Kubernetes (?) Kit, which would be helpful to mention.

articles/components/spreadsheet/index.asciidoc
Mikaelsu: Thanks! Updated.

articles/components/spreadsheet/index.asciidoc
Mikaelsu: Thank you. I did a less "lazy" version :) please review!

articles/components/spreadsheet/index.asciidoc
```diff
@@ -260,17 +260,21 @@ include::{root}/src/main/java/com/vaadin/demo/component/spreadsheet/SpreadsheetR
...
+
 * No provided toolbars, menus, or other controls for formatting cells.
 * Limited support for the older XSL format.
-* Limitations of Apache POI.
+* Constraints built into Apache POI, such as importing and exporting Excel files.
 * Using a table as a named range in formulas is not supported.
-As a workaround, you can manually convert tables to ranges. Open the spreadsheet in Excel, right-click on the table, select the Convert to Range action, and then save the spreadsheet.
-* The SUBTOTAL formula is limited to aggregate functions that do not ignore hidden values (i.e., function codes from 1 to 7, as well as 9) because they're not implemented in Apache POI.
+** As a workaround, you can manually convert tables to ranges. Open the spreadsheet in Excel, right-click on the table, select the Convert to Range action, and then save the spreadsheet.
+* While the component is serializable, Apache POI prevents using the `@PreserveOnRefresh` annotation for preserving objects, and therefore also blocks session replication with Kubernetes Kit.
```
sissbruecker, line 271:
```suggestion
The component is not serializable due to the internal usage of Apache POI. `@PreserveOnRefresh` and session replication with Kubernetes Kit are not supported when using Spreadsheet.
```
Stating that the component is serializable is technically not correct, so I would suggest to state outright that it's not, and why.

sissbruecker (commented):
Thanks for the update, I would still like to tweak the note about serialization somewhat.

sissbruecker (approved):

PR #2158, opened 2023-02-14
docs: add button icon tooltip example

sissbruecker:
Extends the icon button example to show how to add a tooltip.

Also improve admonition to mention that the Flow `Button` component applies the `icon` theme automatically.


jouni (commented):
Does the tooltip actually work as the label in this case, and not just as the description? I think those result in different screen reader behavior.

sissbruecker:
Good point, the tooltip sets `aria-describedby`, so I guess those are not equivalent? The current wording kind of makes it sound like they are. Maybe it should be rephrased to indicate that a tooltip can be additionally added, but an `aria-label` is always recommended.

frontend/demo/component/button/button-icons.ts
```diff
@@ -21,12 +22,18 @@ export class Example extends LitElement {
...
+        <!-- Icon button using an aria-label to provide textual alternative 
+             to screen readers -->
         <vaadin-button theme="icon" aria-label="Add item">
           <vaadin-icon icon="vaadin:plus"></vaadin-icon>
         </vaadin-button>
 
-        <vaadin-button theme="icon" aria-label="Close">
+        <!-- Icon button using a tooltip to provide textual alternative 
+             to screen readers, as well as when hovering the button with 
+             a pointing device, or focusing the button using keyboard -->
```
vursen, line 33:
```suggestion
             a pointing device, or focusing the button using the keyboard -->
```

src/main/java/com/vaadin/demo/component/button/ButtonIcons.java
```diff
@@ -13,13 +13,18 @@
...
         // tag::snippet[]
+        // Icon button using an aria-label to provide textual alternative
+        // to screen readers
         Button plusButton = new Button(new Icon(VaadinIcon.PLUS));
         plusButton.addThemeVariants(ButtonVariant.LUMO_ICON);
         plusButton.getElement().setAttribute("aria-label", "Add item");
 
+        // Icon button using a tooltip to provide textual alternative
+        // to screen readers, as well as when hovering the button with
+        // a pointing device, or focusing the button using keyboard
```
vursen, line 24:
```suggestion
        // a pointing device, or focusing the button using the keyboard
```

vursen (approved):

sissbruecker:
I restored setting the `aria-label`, and changed the text so setting a tooltip is not interpreted as an alternative to setting an `aria-label`.

src/main/java/com/vaadin/demo/component/button/ButtonIcons.java
```diff
@@ -13,13 +13,18 @@
 public class ButtonIcons extends Div {
     public ButtonIcons() {
         // tag::snippet[]
+        // Icon button using an aria-label to provide textual alternative
```
vursen, line 16:
```suggestion
        // Icon button using an aria-label to provide a textual alternative
```

vursen (approved):

src/main/java/com/vaadin/demo/component/button/ButtonIcons.java
```diff
@@ -13,13 +13,18 @@
...
         Button plusButton = new Button(new Icon(VaadinIcon.PLUS));
         plusButton.addThemeVariants(ButtonVariant.LUMO_ICON);
         plusButton.getElement().setAttribute("aria-label", "Add item");
 
+        // Icon button using a tooltip to provide textual description
+        // of the action that it triggers
         Button closeButton = new Button(new Icon(VaadinIcon.CLOSE_SMALL));
+        plusButton.getElement().setAttribute("aria-label", "Close");
         closeButton.addThemeVariants(ButtonVariant.LUMO_ICON);
-        closeButton.getElement().setAttribute("aria-label", "Close");
```
jouni, line 22: Looks like these got mixed up. The `plusButton` now as the aria-label "Close", and the `closeButton` doesn't have an aria-label.

jouni (dismissed):

src/main/java/com/vaadin/demo/component/button/ButtonIcons.java
sissbruecker: Thanks, stupid copy+paste error. Fixed.

PR #2159, opened 2023-02-14
docs: explain Custom Field value type and format

### LINKED ISSUE #1941: Better examples for CustomField
E.g. a custom field whose value is a collection of some custom class, such as `List<Person>`.

vursen:
Added a section explaining the difference between how Custom Field handles the value type and format in Java and TypeScript.

Fixes #1941

### DIFF
--- a/articles/components/custom-field/index.asciidoc
+++ b/articles/components/custom-field/index.asciidoc
@@ -12,7 +12,7 @@ page-links:
 // tag::description[]
 Custom Field is a component for wrapping multiple components as a single field.
 // end::description[]
-It has the same features as <<{components-path-prefix}input-fields#,Input Fields>>, such as its own label, helper, validation, and data binding.
+It has the same features as <<{components-path-prefix}input-fields#,Input Fields>>, such as its label, helper, validation, and data binding.
 Use it to create custom input components.
 
 [.example]
@@ -57,17 +57,17 @@ Custom Field is optimized for wrapping the following components:
 * <<../date-picker#,Date Picker>>
 * <<../time-picker#,Time Picker>>
 
-It can also be used to provide a label, helper, and the other field features, for components that don't have them built-in, such as <<../list-box#,List Box>>.
+It can also be used to give a label, helper, and other field features for components that don't have them built-in, such as <<../list-box#,List Box>>.
 
 == Value Type and Format
 
-Custom Field handles the value type and format differently in Java and TypeScript.
+Custom Field handles the value types and formats differently in Java and TypeScript.
 
 === Java
 
-Custom Field is a generic class that requires a value type. The value type can be anything: String, List, a bean, etc.
+Custom Field is a generic class that requires a value type. The value type can be anything: String, List, a bean, and so forth.
 
-In addition to that, you must implement the `generateModelValue` and `setPresentationValue` methods which define how to build the Custom Field value from the child component values and how to propagate the Custom Field value to the child components respectively.
+Implement the `generateModelValue` and `setPresentationValue` methods to define how to build the Custom Field value from the child component values and how to propagate the Custom Field value to the child components respectively.
 
 The following example shows how to set up value propagation, using a bean as the value type:
 
@@ -155,7 +155,7 @@ render() {
 
 Custom Field works with native HTML elements.
 
-The `whitespace` variant can be used when components without an outer margin are used within Custom Field, to compensate for the missing space between the label and the component itself.
+The `whitespace` variant can be used when components without an outer margin are used within Custom Field to compensate for the missing space between the label and the component itself.
 
 [.example]
 --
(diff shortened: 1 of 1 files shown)


articles/components/custom-field/index.asciidoc
```diff
@@ -59,6 +59,84 @@ Custom Field is optimized for wrapping the following components:
 
 It can also be used to provide a label, helper, and the other field features, for components that don't have them built-in, such as <<../list-box#,List Box>>.
 
+== Value Type and Format
+
+Custom Field handles the value type and format differently in Java and TypeScript.
+
+=== Java
```
vursen, line 66: Is there any better way to split these Java and TypeScript sections? Maybe using an accordion or something.

articles/components/custom-field/index.asciidoc
```diff
@@ -59,6 +59,84 @@ Custom Field is optimized for wrapping the following components:
...
+
+Custom Field is a generic class that requires a value type. The value type can be anything: String, a bean, etc.
```
vursen, line 68:
```suggestion
Custom Field is a generic class that requires a value type. The value type can be anything: String, List, a bean, etc.
```

articles/components/custom-field/index.asciidoc
```diff
@@ -59,6 +59,84 @@ Custom Field is optimized for wrapping the following components:
...
+
+In addition to that, you must provide the `generateModelValue` and `setPresentationValue` methods which specify how to build Custom Field's value from the child component values and how to propagate Custom Field's value to the child components respectively.
+
+The following example shows how to set up the value propagation:
```
vursen, line 72:
```suggestion
The following example shows how to set up the value propagation, using String as the value type:
```

articles/components/custom-field/index.asciidoc
```diff
@@ -59,6 +59,99 @@ Custom Field is optimized for wrapping the following components:
...
+In addition to that, you must provide the `generateModelValue` and `setPresentationValue` methods which specify how to build the Custom Field value from the child component values and how to propagate the Custom Field value to the child components respectively.
+
+The following example shows how to set up value propagation, using a bean as the value type:
+
+[.example]
+--
+
+[source, java]
+----
+class Phone {
```
vursen, line 79: Would it make sense to have here a bean-based example (Phone, Address) or a String-based example (Expiration)?

vursen: I thought an inline example should be enough as it rather focuses on data communication that is of course not visible to the end user.

articles/components/custom-field/index.asciidoc
```diff
@@ -59,6 +59,99 @@ Custom Field is optimized for wrapping the following components:
...
 
+== Value Type and Format
+
+Custom Field handles the value type and format differently in Java and TypeScript.
+
+=== Java
+
+Custom Field is a generic class that requires a value type. The value type can be anything: String, List, a bean, etc.
+
+In addition to that, you must provide the `generateModelValue` and `setPresentationValue` methods which specify how to build the Custom Field value from the child component values and how to propagate the Custom Field value to the child components respectively.
```
vursen, line 70:
```suggestion
In addition to that, you must implement the `generateModelValue` and `setPresentationValue` methods which define how to build the Custom Field value from the child component values and how to propagate the Custom Field value to the child components respectively.
```

articles/components/custom-field/index.asciidoc
rolfsmeds: Inline examples are probably sufficient here – I think everybody can figure out what the field is like from the code (although I suppose a comment in the code briefly describing the field (e.g. "Phone number field consisting of country code and number") would be nice.

The phone example strikes me as much better example of Java usage, so I would actually prefer to keep it. Any reason not to do a similar one for TS? (Although the month/year expiry date example is fine too)

articles/components/custom-field/index.asciidoc
rolfsmeds: I think I've seen somewhere a page that had a tabbed section mid-page, but 1. can't find it now and 2. not sure if that's something we _should_ be doing. I think separate sections like this are fine – sometimes there's just no way to combine without making everything awfully abstract and unhelpful.

articles/components/custom-field/index.asciidoc
vursen: > Any reason not to do a similar one for TS?

We could consider it for the TS example as well, although it can be a bit of a challenge to properly extract the country code from the entire phone string in the parser function since you never know the phone number's length for sure:

- +35 84412345678
- +358 4412345678

We should probably then use Space as a delimiter.

rolfsmeds: Couldn't you use _any_ custom char as a delimiter (ideally one that is not a valid phone number char), say, % or something, since you're in control of both formatting and parsing?

vursen: Well, good point. I just thought that it can generally look unnatural for the phone field to output a phone number with some delimiters, but since the current design of Custom Field is such that you can't get by without delimiters, I think it is okay. I'll update the TS example.

vursen: Ready.

rolfsmeds (approved):

PR #2167, opened 2023-02-15
fix: adjust details and accordion TS examples to V24 API

web-padawan:
## Description

Updated `vaadin-details` and `vaadin-accordion` TS examples:

1. Changed to use `summary` property shorthand where it makes sense (in most of demos),
2. Updated other demos to use `vaadin-details-summary` and `vaadin-accordion-heading`.

### DIFF
--- a/frontend/demo/component/accordion/accordion-summary.ts
+++ b/frontend/demo/component/accordion/accordion-summary.ts
@@ -7,6 +7,7 @@ import '@vaadin/button';
 import '@vaadin/combo-box';
 import '@vaadin/email-field';
 import '@vaadin/form-layout';
+import '@vaadin/horizontal-layout';
 import '@vaadin/text-field';
 import '@vaadin/vertical-layout';
 import type { AccordionOpenedChangedEvent } from '@vaadin/accordion';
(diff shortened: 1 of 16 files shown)


frontend/demo/component/accordion/accordion-summary.ts
```diff
@@ -58,7 +58,7 @@ export class Example extends LitElement {
         }}"
       >
         <vaadin-accordion-panel>
-          <div slot="summary">
+          <vaadin-accordion-heading slot="summary">
```
sissbruecker, line 61: There is a regression here in that the additional details don't show up on a new line:

Bildschirmfoto 2023-02-15 um 10 49 45

Also the example is just terrible. I already had a review comment that it doesn't work at all, until I noticed that you have to enter a text first. We should change this to just add something static like an icon or something, rather than have all the convoluted logic for binding the form values into the summary. The whole setup just distracts from what we want to demonstrate.

frontend/demo/component/details/details-summary.ts
```diff
@@ -42,22 +42,21 @@ export class Example extends LitElement {
...
 
-          <vaadin-horizontal-layout
-            style="color: var(--lumo-error-text-color); margin-left: var(--lumo-space-s)"
-          >
-            <vaadin-icon
-              icon="vaadin:exclamation-circle"
-              style="width: var(--lumo-icon-size-s); height: var(--lumo-icon-size-s);"
-            ></vaadin-icon>
-            <span>2 errors</span>
+            <vaadin-horizontal-layout
```
sissbruecker, line 49: Let's add some spacing between the icon and the text

frontend/demo/component/accordion/accordion-summary.ts
web-padawan: Updated the PR to also use `vaadin-horizontal-layout` which makes elements align correctly.
Note, this particular demo was intentionally implemented this way to illustrate a complex form.

frontend/demo/component/details/details-summary.ts
web-padawan: Done.

frontend/demo/component/accordion/accordion-summary.ts
sissbruecker: Still disagree, but it's not a good reason to block this PR.

sissbruecker (approved):

frontend/demo/component/accordion/accordion-summary.ts
web-padawan: @sissbruecker Feel free to create a new issue as a follow-up, let's discuss if we should reduce the complexity.

PR #2168, opened 2023-02-15
fix: use HTML value in RTE examples

sissbruecker:
As part of updating the RTE docs for v24, I missed to update the remaining examples to initialize the editor with an HTML value. Now that the component has a validation check that prevents setting delta values, this breaks every single example.

### DIFF
--- a/src/main/java/com/vaadin/demo/domain/Templates.java
+++ b/src/main/java/com/vaadin/demo/domain/Templates.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 public class Templates {
 
     private String richTextDelta;
+    private String richTextHtml;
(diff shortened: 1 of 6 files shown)


src/main/java/com/vaadin/demo/domain/Templates.java
```diff
@@ -17,6 +18,14 @@ public void setRichTextDelta(String richTextDelta) {
         this.richTextDelta = richTextDelta;
     }
 
+    public String getRichTextHtml() {
```
vursen, line 21: Should we remove richTextDelta-related getters and setters?

sissbruecker: Sure, done

vursen (approved):

PR #2173, opened 2023-02-15
docs: do not remove aria-labelledby when setting aria-label

vursen:
The fields' `aria-labelledby` attribute no longer contains the label id if the label is empty, which means we don't have to remove `aria-labelledby` when setting `aria-label`.

Part of #1941 

Depends on:
- https://github.com/vaadin/web-components/pull/5577


vursen:
Let's wait until the next web-components release for the fix to land in `23.3`, `24.0`.

vursen:
Closing as the above fix doesn't solve the issue completely. It is a more fundamental issue: https://github.com/vaadin/web-components/issues/5580

PR #2174, opened 2023-02-15
fix:Reference date not working according to documentation

onuridrisoglu:
Changed the example in documentation to how it actually works

Fixes: https://github.com/vaadin/flow-components/issues/4629


sissbruecker:
Thanks, however looking into it, this is actually a bug: https://github.com/vaadin/flow-components/issues/4629#issuecomment-1431495312

We should leave the docs as is, and fix the bug instead.

PR #2176, opened 2023-02-16
docs: update RTE value example to show Delta value

sissbruecker:
Updates the example under `Value format` to also show the Delta value, and allow editing it. This is adapted from the updated example from v24: https://github.com/vaadin/docs/pull/2027

There are some differences in the example compared to v24:
- It uses an improved helper text instead of a placeholder. I'll add that for v24 in a separate PR.
- In the Java example, changing the Delta value does not update the HTML value. That is not supported by the Flow component in v23.

The PR also updates the other examples to not use the deprecated `setValue` method.

### DIFF
--- a/articles/components/rich-text-editor/index.asciidoc
+++ b/articles/components/rich-text-editor/index.asciidoc
@@ -122,7 +122,11 @@ Rich Text Editor natively uses the JSON-based https://github.com/quilljs/delta[D
 
 [source,typescript]
 ----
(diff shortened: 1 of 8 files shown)


tomivirkki (approved):
> In the Java example, changing the Delta value does not update the HTML value. That is not supported by the Flow component in v23.

The value updates but it's always 1 step behind the updated value. Not a blocker for this PR.

https://user-images.githubusercontent.com/1222264/219298354-4dd15101-08a8-4ba8-8c3b-dfe62e156489.mp4

sissbruecker:
> The value updates but it's always 1 step behind the current value. Not a blocker for this PR.
Yeah, that would probably require further changes in the component. For the purpose of this example (have a playground for the delta value as well), it should be fine.

Issue #105, opened 2021-01-13
Collaboration Engine docs for V18+ have @Push on the wrong class


OlliTietavainenVaadin:
https://vaadin.com/docs/v18/ce/tutorial.html#ce.tutorial.push
The `@Push` annotation should be on the `AppShellConfigurator` class, not the main layout, in V18 and later.

pekam:
Closing as a duplicate of https://github.com/vaadin/collaboration-engine-internal/issues/483

PR #2179, opened 2023-02-16
docs: use proper error messages to fix examples a11y findings

web-padawan:
## Description

This PR aims to fix accessibility review findings for the following components:

- [x] `DatePicker`
- [x] `DateTimePicker`
- [x] `TimePicker`

Fixes https://github.com/vaadin/web-components/issues/104
Fixes https://github.com/vaadin/web-components/issues/105
Fixes https://github.com/vaadin/web-components/issues/109

## Note

The part about old "Custom validator" example in https://github.com/vaadin/web-components/issues/104 is no longer relevant:

> The same happens for the [Custom validator](https://vaadin.com/components/date-picker/html-examples/date-picker-validation-demos#custom-validator) example if the user chooses a date that is not a working day.

> The date picker with an invalid date - only the background colour of the control is changed

### DIFF
--- a/frontend/demo/component/datepicker/date-picker-min-max.ts
+++ b/frontend/demo/component/datepicker/date-picker-min-max.ts
@@ -21,34 +21,25 @@ export class Example extends LitElement {
   private errorMessage = '';
 
   @state()
-  private today = '';
+  private minDate = new Date();
 
   @state()
-  private upperLimit = '';
-
-  protected override firstUpdated() {
-    this.today = formatISO(Date.now(), { representation: 'date' });
-
-    const upperLimit = addDays(Date.now(), 60);
-    this.upperLimit = formatISO(upperLimit, { representation: 'date' });
-  }
+  private maxDate = addDays(new Date(), 60);
 
   protected override render() {
     return html`
       <!-- tag::snippet[] -->
       <vaadin-date-picker
-        .min="${this.today}"
-        .max="${this.upperLimit}"
         label="Appointment date"
         helper-text="Must be within 60 days from today"
+        .min="${formatISO(this.minDate, { representation: 'date' })}"
+        .max="${formatISO(this.maxDate, { representation: 'date' })}"
         .errorMessage="${this.errorMessage}"
         @change="${({ target }: DatePickerChangeEvent) => {
-          const [value, min, max] = [target.value, this.today, this.upperLimit].map((date) =>
-            dateFnsParse(date ?? '', 'yyyy-MM-dd', new Date())
-          );
-          if (isBefore(value, min)) {
+          const date = dateFnsParse(target.value ?? '', 'yyyy-MM-dd', new Date());
+          if (isBefore(date, this.minDate)) {
             this.errorMessage = 'Too early, choose another date';
-          } else if (isAfter(value, max)) {
+          } else if (isAfter(date, this.maxDate)) {
             this.errorMessage = 'Too late, choose another date';
           } else {
             this.errorMessage = '';
--- a/frontend/demo/component/datepicker/date-picker-min-max.ts
+++ b/frontend/demo/component/datepicker/date-picker-min-max.ts
@@ -3,8 +3,10 @@ import 'Frontend/demo/init'; // hidden-source-line
 import { html, LitElement } from 'lit';
 import { customElement, state } from 'lit/decorators.js';
 import '@vaadin/date-picker';
+import type { DatePickerChangeEvent } from '@vaadin/date-picker';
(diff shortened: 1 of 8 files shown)


frontend/demo/component/timepicker/time-picker-min-max.ts
```diff
@@ -24,6 +28,17 @@ export class Example extends LitElement {
         min="08:00"
         max="16:00"
         .step="${60 * 30}"
+        error-message="${this.errorMessage}"
+        @change="${(event: TimePickerChangeEvent) => {
+          const { min, max, value } = event.target;
+          if (value < min) {
```
vursen, line 34: Are you comparing strings here? I'm afraid this won't work.

web-padawan: It actually works at least in such simple cases. I could add a comment about it though.

Screenshot 2023-02-16 at 13 22 45

vursen: Hmm, right.

frontend/demo/component/datepicker/date-picker-min-max.ts
```diff
@@ -36,6 +41,19 @@ export class Example extends LitElement {
         .max="${this.upperLimit}"
         label="Appointment date"
         helper-text="Must be within 60 days from today"
+        .errorMessage="${this.errorMessage}"
+        @change="${({ target }: DatePickerChangeEvent) => {
+          const [value, min, max] = [target.value, this.today, this.upperLimit].map((date) =>
```
vursen, line 46:
```suggestion
          const [value, min, max] = [target.value, this.minDate, this.maxDate].map((date) =>
```

vursen: How about we align the way we define `minDate` and `maxDate` with the DateTimePicker example below?

src/main/java/com/vaadin/demo/component/datetimepicker/DateTimePickerMinMax.java
```diff
@@ -20,6 +20,18 @@ public DateTimePickerMinMax() {
         dateTimePicker.setMin(LocalDateTime.now());
         dateTimePicker.setMax(LocalDateTime.now().plusDays(60));
         dateTimePicker.setValue(LocalDateTime.now().plusDays(7));
+        dateTimePicker.addValueChangeListener(event -> {
+            LocalDateTime value = event.getValue();
+            if (value.compareTo(dateTimePicker.getMin()) < 0) {
```
vursen, line 25: You should probably call `compareTo` on `dateTimePicker.getMin()` because `value` can be `null`.

frontend/demo/component/datepicker/date-picker-min-max.ts
web-padawan: Done.

src/main/java/com/vaadin/demo/component/datetimepicker/DateTimePickerMinMax.java
web-padawan: Added `null` checks in listeners.

vursen (approved):

frontend/demo/component/datetimepicker/date-time-picker-custom-validation.ts
```diff
@@ -19,17 +19,19 @@ export class Example extends LitElement {
   // tag::snippet[]
   private binder = new Binder(this, AppointmentModel);
 
+  private errorMessage = 'The selected day of week or time is not available';
+
```
tomivirkki, line 23: Why this change? The example demonstrated that you can have different error messages for different condition violations. Not both condition violations show the same error message.

web-padawan: @tomivirkki This is a TetraLogical finding from #105: 

> For the Custom Validation example, if the user chooses both an invalid date and an invalid time, an error text is shown that only mentions the date, meaning the error for the time field is only indicated using colour.

We already have both DatePicker and TimePicker using different error messages depending on min / max.
So I thought it should be "good enough" this way to fully address the original accessibility issues.

tomivirkki: Ah, I see. Makes sense :+1: For Hilla users, there still the `custom-field-basic` example, which has multiple constraints with different error messages.

tomivirkki (approved):

web-padawan:
Changed the base branch to `latest` so that these fixes also make it to the Vaadin 23.3 documentation.

PR #2186, opened 2023-02-16
docs: use title instead of aria-label in Custom Field examples

vursen:
We internally agreed with @web-padawan that we will use the `title` attribute instead of `aria-label` in Custom Field examples until https://github.com/vaadin/web-components/issues/5580 is fixed.

### DIFF
--- a/frontend/demo/component/custom-field/custom-field-basic.ts
+++ b/frontend/demo/component/custom-field/custom-field-basic.ts
@@ -7,6 +7,7 @@ import { applyTheme } from 'Frontend/generated/theme';
 import { Binder, field } from '@hilla/form';
 import AppointmentModel from 'Frontend/generated/com/vaadin/demo/domain/AppointmentModel';
 import { differenceInDays, parseISO, isAfter } from 'date-fns';
+import { DatePicker } from '@vaadin/date-picker';
 
 @customElement('custom-field-basic')
 export class Example extends LitElement {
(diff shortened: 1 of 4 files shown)


web-padawan (approved):

frontend/demo/component/custom-field/custom-field-size-variants.ts
```diff
@@ -6,6 +6,8 @@ import '@vaadin/horizontal-layout';
 import '@vaadin/select';
 import '@vaadin/text-field';
 import { applyTheme } from 'Frontend/generated/theme';
+import { TextField } from '@vaadin/text-field';
+import { Select } from '@vaadin/select';
```
web-padawan, line 10:
```suggestion
import type { Select } from '@vaadin/select';
import type { TextField } from '@vaadin/text-field';
```

frontend/demo/component/custom-field/custom-field-basic.ts
```diff
@@ -7,6 +7,7 @@ import { applyTheme } from 'Frontend/generated/theme';
 import { Binder, field } from '@hilla/form';
 import AppointmentModel from 'Frontend/generated/com/vaadin/demo/domain/AppointmentModel';
 import { differenceInDays, parseISO, isAfter } from 'date-fns';
+import { DatePicker } from '@vaadin/date-picker';
```
web-padawan, line 10:
```suggestion
import type { DatePicker } from '@vaadin/date-picker';
```

PR #2193, opened 2023-02-17
docs: add missing imports to grid inline editor examples

### LINKED ISSUE #810: "Open in a new tab" button not working for Java only examples
Cannot confirm if that happens in all cases and I can't think of any other examples implemented only in Java, but clicking on the "Open in a new tab" button on the examples at the [Inline Editing (Java Only) for Grid](https://vaadin.com/docs/latest/ds/components/grid#inline-editing-java-only) leads to a blank box.

ugur-vaadin:
Opening the examples in new tabs did not work due to missing imports. This PR adds those imports to the 2 Java-only inline editor examples.

Fixes #810


sissbruecker:
@ugur-vaadin Any reason not to add this to `latest`?

Issue #739, opened 2021-09-13
Document Lookup/Instantiator


mvysny:
There is no documentation for custom Lookup/Instantiator. Trying to implement one is a series of try-and-fail, steps, missteps, going through crappy outdated javadoc etc. Absolutely no mention of instantiator/lookup in the documentation. This is completely ridiculous.

Vaadin 20.

For example, if I want to implement a custom Instantiator for Dagger DI:
1. I go to the Instantiator class and learn that the Java ServiceLoader is used. That is not true: javadoc of init() says that InstantiatorFactory is used.
2. The InstantiatorFactory javadoc says absolutely nothing about how it is being created, and how I can force Vaadin to use a custom one.
3. The VaadinService.loadInstantiators() sources finally reveals that the Lookup class is used, to load InstantiatorFactory and produce an Instantiator.
4. I now have no idea what is the correct way to go: should I provide a custom Lookup which is able to provide InstantiatorFactory, or should I rather override VaadinService.loadInstantiators()? What is the best practice here?
5. Say that I want to be a good citizen and provide my own Lookup interface. How should I do it? The Lookup javadoc doesn't mention at all, how exactly I can provide a custom implementations of Lookup. I try to analyze sources for LookupInitializer which contain more of abstract shit, ultimately giving up and giving Lookup the finger.
6. Overriding VaadinService and providing custom impl of loadInstantiators() is much easier - it doesn't feel right but I don't care at this point - I'm pissed off since I wasted a lot of time and I need to move forward with this shit.
7. How should I now implement the Instantiator? Should I extend something or delegate to something or implement from scratch? Should I delegate to DefaultInstantiator? Not easy: https://github.com/vaadin/flow/issues/8427 and it's NOT DOCUMENTED. If I start implementing that shit from scratch, is it okay if I return null for `getI18nProvider()`? What about `getServiceInitListeners()`, how to implement that?

This experience is utter crap. Improve it!

OlliTietavainenVaadin:
The mechanism can be found from here: https://mvysny.github.io/vaadin-custom-instantiator/

jouni:
I’m quite out of my comfort zone with this topic, but from what I gather, there are some issues in the JavaDocs. @mshabarov, could you take a look into those, if there’s incorrect documentation there?

PR #2195, opened 2023-02-17
feat: Document instantiator

caalador:
Add documentation for creating
a custom instantiator.

Part of #739

### DIFF
--- a/articles/advanced/custom-instantiators.asciidoc
+++ b/articles/advanced/custom-instantiators.asciidoc
@@ -1,0 +1,178 @@
+---
+title: Custom Instantiator
+description: Creating a custom Instantiator implementation.
+order: 750
+layout: page
+---
+
+
+= Custom Instantiator
+
+An [interfacename]`Instantiator` is used for discovering, creating and managing object instances that Flow uses.
+Usually a custom instantiator is needed for dependency injection frameworks so they can provide managed instances of objects according to the specific conventions of that DI framework.
+
+== Creating a custom Instantiator
+
+To create a custom instantiator implement the [interfacename]`Instantiator` interface and implement the methods:
+
+- `Stream<VaadinServiceInitListener> getServiceInitListeners()`
+- `T getOrCreate(Class<T> type)`
+- `T createComponent(Class<T> componentClass)`
+
+Usually it is good to extend the [classname]`DefaultInstantiator` and only implement [methodname]`getOrCreate` and [methodname]`createComponent` with dependency injection bean handling with a fallback to the default functionality.
+
+- [methodname]`getOrCreate(Class<T>)` should always return a initialized class of type T where the class returned is a initialized one from the current scope or generate a new bean instance or if not a bean handle field injection.
+- [methodname]`createComponent(Class<T>)` should always create a new component with field injection.
+
+As an example the simplified instantiator for cdi could be as follows
+
+.Simplified CustomInstantiator for CDI sample
+[source,java]
+----
+import com.vaadin.flow.component.Component;
+import com.vaadin.flow.di.DefaultInstantiator;
+
+import jakarta.enterprise.inject.spi.BeanManager;
+import jakarta.enterprise.inject.spi.Unmanaged;
+import jakarta.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CustomInstantiator extends DefaultInstantiator {
+
+    @Inject
+    private BeanManager beanManager;
+
+    public CustomInstantiator(BeanManager beanManager, VaadinService service) {
+        super(service);
+        this.beanManager = beanManager;
+    }
+
+    /**
+     * Hands over an existing bean or tries to instantiate one.
+     */
+    @Override
+    public<T> T getOrCreate(Class<T> type) {
+        Set<Bean<T>> beans = beanManager.getBeans(type);
+        if(beans == null || beans.isEmpty()) {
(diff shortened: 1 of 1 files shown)


articles/advanced/custom-instantiators.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+To create a custom instantiator implement the [interfacename]`Instantiator` interface and implement the methods:
+
+- `Stream<VaadinServiceInitListener> getServiceInitListeners()`
+- `T getOrCreate(Class<T> type)`
+- `T createComponent(Class<T> componentClass)`
+
+Usually it is good to extend the [classname]`DefaultInstantiator` and only implement [methodname]`getOrCreate` and [methodname]`createComponent` with dependency injection bean handling with a fallback to the default functionality.
+
+- [methodname]`getOrCreate(Class<T>)` should always return a initialized class of type T where the class returned is a initialized one from the current scope or generate a new bean instance or if not a bean handle field injection.
```
mcollovati, line 24:
```suggestion
- [methodname]`getOrCreate(Class<T>)` should always return a initialized object of type T where the instance returned is a initialized one from the current scope or generate a new bean instance or if not a bean handle field injection.
```

articles/advanced/custom-instantiators.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+    private static Logger getLogger() {
+        return LoggerFactory.getLogger(CdiInstantiator.class);
+    }
+}
+----
+
+== Loading the custom instantiator using Java ServiceLoader
+
+To load the instantiator you should implement the [interfacename]`InstantiatorFactory` interface and load it through the Java ServiceLoader.
```
mcollovati, line 88: Looking at the code it seems that we are directly loading `Instantiator` implementations through `ServiceLoader`
The `InstantiatorFactory` is instead used by the `Lookup` mechanism

mcollovati (changes requested):
It would be nice to add that only a single implementation of `Instantiator` is supported, otherwise the application fails during startup

articles/advanced/custom-instantiators.asciidoc
mvysny: Actually `VaadinService.loadInstantiators()` calls Lookup which uses `ServiceLoader.load` on `InstantiatorFactory`. The fallback of `ServiceLoader.load`ing `Instantiator` is never used since Lookup is always there.

mvysny (approved):
LGTM, good work

articles/advanced/custom-instantiators.asciidoc
mcollovati: Thanks, I completely missed the usage of `ServiceLoader` in `LookupInitializer`

mcollovati (approved):

articles/advanced/custom-instantiators.asciidoc
caalador: Any InstantiatorFactory implementations should be collected and handled by the LookupServletContainerInitializer on load so you shouldn't need to add it yourself as a service file, but that does work as a fallback. even in Lookup.

The big question now is how to formulate this.

articles/advanced/custom-instantiators.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+- `Stream<VaadinServiceInitListener> getServiceInitListeners()`
+- `T getOrCreate(Class<T> type)`
+- `T createComponent(Class<T> componentClass)`
+
+Usually it is good to extend the [classname]`DefaultInstantiator` and only implement [methodname]`getOrCreate` and [methodname]`createComponent` with dependency injection bean handling with a fallback to the default functionality.
+
+- [methodname]`getOrCreate(Class<T>)` should always return a initialized object of type T where the instance returned is a initialized one from the current scope or generate a new bean instance or if not a bean handle field injection.
+- [methodname]`createComponent(Class<T>)` should always create a new component with field injection.
+
+As an example the simplified instantiator for cdi could be as follows
```
mshabarov, line 27:
```suggestion
As an example the simplified instantiator for Context and Dependency Injection (CDI) could be as follows
```

articles/advanced/custom-instantiators.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+layout: page
+---
+
+
+= Custom Instantiator
+
+An [interfacename]`Instantiator` is used for discovering, creating and managing object instances that Flow uses.
+Usually a custom instantiator is needed for dependency injection frameworks so they can provide managed instances of objects according to the specific conventions of that DI framework.
+
+== Creating a custom Instantiator
```
mshabarov, line 14:
```suggestion
== Creating a Custom Instantiator
```

mshabarov (commented):
To give more examples to users, maybe worth it to put a references to the Spring and OSGi add-ons that have also their own instantiators?
https://github.com/vaadin/flow/blob/main/vaadin-spring/src/main/java/com/vaadin/flow/spring/SpringInstantiator.java

https://github.com/vaadin/osgi/tree/main/flow-osgi/src/main/java/com/vaadin/flow/osgi/support
?

caalador:
The OSGi one is not all that helpful and linking to source files seems bad as they easily go old, for instance last with the change from master to main.

mshabarov (approved):

jouni:
@caalador, when adding new pages, the `[discussion-id]´ should be a new UUID.

Issue #385, opened 2021-04-20
Mention the need to explicitly set a sort property with lazy dataproviders and a renderer


OlliTietavainenVaadin:
If you have a lazy-loading DataProvider and a column with a Renderer which gets its data from a bean property via a String property name, it should be mentioned eg. in https://vaadin.com/docs/latest/components/grid/flow/#using-a-sort-property-name that an explicit sort property name is mandatory to get the necessary sorting info to the backend.

Avec112:
sort not working
`grid.addColumn(new LocalDateTimeRenderer<>(MyClass::getDateTime, DATE_TIME_FORMAT), "dateTime"); `

sort working
`grid.addColumn(new LocalDateTimeRenderer<>(MyClass::getDateTime, DATE_TIME_FORMAT)).setSortProperty("dateTime");`

PR #2200, opened 2023-02-17
docs: improve Flow grid sorting section

sissbruecker:
Reworks the Flow-specific `Column Sorting` section. While the content was good, there were some confusing or missing parts, and usage of deprecated or removed API.

I tried to address the following issues:
- Start by explaining how to enable sorting in the first place, when it is enabled automatically, and when it is not.
- In addition to explaining in-memory and back-end sorting, also explain when a column is automatically configured for either method. 
- Removed section that used deprected methods of setting sort property names directly in `Grid.addColumn(String, String...)`. This is also required in v24, where these methods have been removed.
- Reworked the following two sections to be explicitely about configuring in-memory sorting, and back-end sorting, rather than about which API methods they use.
- Removed confusing part that mentions that both in-memory and back-end sorting can be used together. It didn't bother explaining how that would work, so it seems better to remove it.

Closes https://github.com/vaadin/docs/issues/385

### DIFF
--- a/articles/components/grid/flow.asciidoc
+++ b/articles/components/grid/flow.asciidoc
@@ -886,23 +886,23 @@ var isSortable = column.isSortable();
 ```
 
 Depending on how the column was created, it may need additional configuration.
-In order to figure out whether that is the case, let's first consider the two sorting mechanisms supported by Grid:
+In order to figure out whether that is the case, we need to consider the two sorting mechanisms supported by Grid:
 
 * *In-memory sorting* is sorting that is automatically applied by the framework when using an in-memory data provider.
 Items are sorted using a `Comparator` that can either be created automatically by the column or has to be specified manually by the developer.
 
-* *Back-end sorting* is applied by providing a list of [classname]`QuerySortOrder` objects a custom [classname]`DataProvider` implemented by the developer.
+* *Backend sorting* is applied by providing a list of [classname]`QuerySortOrder` objects a custom [classname]`DataProvider` implemented by the developer.
 Sort order objects contain the property names that the developer must use for sorting items in their backend, such as a database.
 See <<{articles}/binding-data/data-provider#Sorting,Data Providers>> for more.
 
-With that in mind, let's see which columns are automatically configured for in-memory, or back-end sorting:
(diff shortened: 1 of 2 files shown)


articles/components/grid/flow.asciidoc
```diff
@@ -866,76 +866,67 @@ If multi-sorting is enabled, the user can sort by multiple columns.
...
 
-=== Defining Column Sorting
+=== Enabling Column Sorting
 
-The difference between in-memory and back-end sorting is key to understanding the sorting mechanism:
+For columns that have been created from a property name (`grid.addColumn("firstName")`) sorting is enabled by default, if the type of the property implements `Comparable`.
+In general this only applies to grids that have been created from a class name (`new Grid(Person.class)`).
 
-* *In-memory sorting* is sorting that's applied by the framework to items fetched from the backend, before returning them to the client.
+For columns that have been created from a `ValueProvider` (`grid.addColumn(Person::getFirstName)`), or from a renderer such as `LitRenderer` (`grid.addColumn(new LitRenderer<Person>(...))`), sorting is not enabled by default.
```
sissbruecker, line 874: The inline code snippets don't look great, but I couldn't come up with anything better. They are important though, otherwise the textual description is too abstract.

articles/components/grid/index.asciidoc
```diff
@@ -354,6 +354,8 @@ include::{root}/src/main/java/com/vaadin/demo/component/grid/GridColumnWidth.jav
 Any column can be made sortable.
 Enable sorting to allow the user to sort items alphabetically, numerically, by date, etc.
 
+For detailed information on how to configure sorting in Flow, see <<flow#column_sorting, Column Sorting>>.
```
sissbruecker, line 357: I'm not sure if the link is correct, it always puts me at the top of the page.

tomivirkki: Using `column-sorting ` instead of `column_sorting` works

sissbruecker: Thanks, done

tomivirkki (approved):
Great rewrite, lgtm

PR #2201, opened 2023-02-17
chore: improve resource loading documentation

### LINKED ISSUE #920: Add non-image example for resources in Resource Cheat Sheet
Resource Cheat Sheet: https://vaadin.com/docs/latest/flow/advanced/loading-resources/#resource-cheat-sheet

Apparently, non-image resources (.pdf, .rtf files and so one) require a forward slash in order to be loaded correctly; for example, `VaadinServlet.getCurrent().getServletContext().getResource("docs/foo.pdf");` does not work, but  `VaadinServlet.getCurrent().getServletContext().getResource("/docs/foo.pdf");` does. This is opposed to Images, where a forward slash is not needed, as in the current cheat sheet's example:
<html>
<body>

new Image("img/flower.jpg", "A flower") | /src/main/webapp/img/flower.jpg
-- | --

</body>
</html>
There could be a separate row for non-image resources to point this out.

mcollovati:
Closes #920


jouni:
I assume any potential factual errors can be fixed later if needed, that there’s no harm if something isn't 100% accurate in this change (as no one has reviewed and approved them yet).

PR #2209, opened 2023-02-20
docs: udpdate jetty live update info in V24

MarcinVaadin:
V24 uses Jetty 11 with Jakarta 10 support, docs update according live update.


jouni (approved):

jouni (changes requested):
If this only applies to Vaadin 24, then the changes should target the `next` branch. Please rebase (I tried but it didn't seem to be that simple 😅 ).

MarcinVaadin:
I've first did it on `next` but then got confused which is which and did another push to `latest` :-/

MarcinVaadin:
Wrong base branch

PR #2218, opened 2023-02-21
fix: remove hardcoded versions in pom

ZheSun88:
versions used for sso-kit and the other two components are outdated, they should come from the vaadin-bom directly.


heruan (approved):

ZheSun88:
This PR should be ported to `next` branch, too

PR #2222, opened 2023-02-23
refactor: use Maven wrapper for build scripts

jouni:
This allows the script the used in environments where Maven is not installed globally (e.g. Netlify).


jouni:
Looks like Netlify only has Java 8 installed by default, with no easy way to use Java 11 (or Java 17 for Vaadin 24). 😞

PR #2225, opened 2023-02-24
fix: removed duplicated section with wrong contents for MPR for Vaadin 8

mcollovati:


mcollovati:
Thanks to @Ansku for reporting this!

PR #2246, opened 2023-03-03
fix: faulty code

caalador:
Fix formatting issue


MatthewVaadin (approved):
Good catch, thank you!

PR #2265, opened 2023-03-09
Update index.adoc

gtzluis:
updated pom.xml Vaadin version from rc1 to 24.0.0


Mikaelsu (approved):
LGTM!

jouni:
Actually, we should update this shared file, and reference the attribute from there: https://github.com/vaadin/docs/blob/latest/articles/_vaadin-version.adoc

@ZheSun88, can we get automation to update that vaadin-version.adoc file?

jouni (approved):

PR #2266, opened 2023-03-09
fix: move itemHasChildrenPath property to the grid element

tomivirkki:
The deprecated property `itemHasChildrenPath` was removed from `<vaadin-grid-tree-column>` in https://github.com/vaadin/web-components/pull/4942


vursen (approved):
Confirmed locally that it fixes the broken examples.

PR #2271, opened 2023-03-09
docs: documented simplified exclusion of legacy dependency in MPR project

mcollovati:
MPR documentation provides the instructions to find and add exclusion rules for dependencies that transitively import the legacy vaadin-server. This change documents an easy exclusion way, by explicitly adding the dependency to the legacy artifact with provided scope, mentioning also potential issues that may arise during development with this approach.


jouni:
I assume this is good to go. If there are some factual errors, those can be fixed afterwards. This is more valuable published than as a pending PR.

PR #2274, opened 2023-03-11
Add Hilla SSO Kit Client documentation

tamasmak:


articles/lit/acceleration-kits/sso-kit/getting-started.adoc
```diff
@@ -131,25 +130,31 @@ To try this feature, replace the `@AnonymousAllowed` annotation in [filename]`He
...
-
 ifdef::lit[]
-All of the essential authentication state is already available in a global variable and can be added to the application state, which is inside the `AppStore` class in [filename]`app-store.ts`:
+=== Add the SSO Kit Client dependency
 
-.frontend/stores/app-store.ts
-[source,typescript]
-----
-import SingleSignOnData from 'Frontend/generated/dev/hilla/sso/starter/SingleSignOnData';
+Add the `sso-kit-client-lit` library to the [filename]`packages.json` of a Hilla application like so:
```
tarekoraby, line 138: Which is better, editing `packages.json` or doing `npm install --save`?

articles/lit/acceleration-kits/sso-kit/getting-started.adoc
```diff
@@ -475,69 +439,26 @@ The client should then be configured on the provider's dashboard to send logout
...
-----
-
-Then, a dialog can be added to the application layout to notify the user:
+A dialog can be added to the application layout to notify the user:
 
 .frontend/views/main-layout.ts
 [source,typescript]
 ----
 import '@vaadin/confirm-dialog';
+import { singleSignOnContext } from '@sso-kit/sso-kit-client-lit/SingleSignOnContext.js';
```
cromoteca, line 450: I suggest to
```suggestion
import { singleSignOnContext as sso } from '@sso-kit/sso-kit-client-lit/SingleSignOnContext.js';
```
and then use `sso` instead of the long `singleSignOnContext` one in code.

articles/lit/acceleration-kits/sso-kit/getting-started.adoc
tamasmak: `npm install --save` is much better and preferred, changed to use that.

articles/lit/acceleration-kits/sso-kit/getting-started.adoc
tamasmak: Now, `ssoContext` is used  for `SingleSignOnContext` and `user` for `User` classes.

articles/react/acceleration-kits/sso-kit/getting-started.adoc
```diff
@@ -1,8 +1,496 @@
...
+
+// Store the logged out state
+const [loggedOut, setLoggedOut] = useState(false);
+
+// Register a callback using useEffect
+useEffect(() => {
+  ssoContext.onBackChannelLogout(() => {
+    setLoggedOut(true);
+  });
+}, []);
```
mcollovati, line 479: Can we make explicit that this code should go inside the `MenuOnLeftLayout` function?

articles/react/acceleration-kits/sso-kit/getting-started.adoc
```diff
@@ -1,8 +1,496 @@
...
+                    <version>${hilla.version}</version>
+                    <!-- Add this configuration -->
+                    <configuration>
+                        <parser>
+                            <packages>
+                                <package>com.example.application</package>
+                                <package>dev.hilla.sso.starter</package>
+                            </packages>
+                        </parser>
+                    </configuration>
```
mcollovati, line 112: As commented in the lit `getting-started`, perhaps there's no need to duplicate the configuration in the production profile

articles/react/acceleration-kits/sso-kit/getting-started.adoc
tamasmak: Yes, that would make it more clear.

articles/lit/acceleration-kits/sso-kit/getting-started.adoc
```diff
@@ -66,38 +66,56 @@ Check the https://github.com/vaadin/sso-kit/releases[SSO Kit releases] page for
...
+                    <!-- Add this configuration -->
+                    <configuration>
+                        <parser>
+                            <packages>
+                                <package>com.example.application</package>
+                                <package>dev.hilla.sso.starter</package>
+                            </packages>
+                        </parser>
+                    </configuration>
+                    <!-- ... -->
```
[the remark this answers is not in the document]
tamasmak: Good point. To add it to `build` section only is enough.

articles/react/acceleration-kits/sso-kit/getting-started.adoc
tamasmak: Same comment here as for the lit `getting started`.

articles/react/acceleration-kits/sso-kit/getting-started.adoc
tamasmak: Updated the comment part.

articles/react/acceleration-kits/sso-kit/getting-started.adoc
```diff
@@ -1,8 +1,496 @@
...
+<dependency>
+    <groupId>dev.hilla</groupId>
+    <artifactId>sso-kit-starter</artifactId>
+    <version>2.0.0</version>
+</dependency>
+
+<dependency>
+    <groupId>org.springframework.boot</groupId>
+    <artifactId>spring-boot-starter-oauth2-resource-server</artifactId>
+</dependency>
```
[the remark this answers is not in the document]
tamasmak: Actually, non of the spring dependencies were required. Removed them in both docs.

articles/lit/acceleration-kits/sso-kit/getting-started.adoc
```diff
@@ -502,118 +384,75 @@ The client should then be configured on the provider's dashboard to send logout
...
+// Add the confirm dialog to the AppLayout
+<ConfirmDialog header='Logged out' cancelButtonVisible
+             opened={loggedOut}
+             onConfirm={ssoContext.loginAgain}
+             onCancel={() => {
+               ssoContext.stayOnPage();
+               setLoggedOut(false);
+             }}>
   <p>You have been logged out. Do you want to log in again?</p>
 </ConfirmDialog>
```
heruan, line 452: This example (and its React counterpart) suggests a confusing use case: it first calls `setLoggedOut(true)` when the logout happens, then opens the dialog and if it is canceled it calls `setLoggedOut(false)` while the user is no more authenticated.

This use case should be expressed with a consistent example, possibly changing the API (see https://github.com/vaadin/sso-kit/issues/93).

heruan (changes requested):

articles/lit/acceleration-kits/sso-kit/getting-started.adoc
tamasmak: Changing the mentioned API simplified this part. Don't need to call the  `setLoggedOut(false)` anymore as the the new `logout()` function will make the page to reload and set the logged out state to its default value. Updated the documentation according to this.

cromoteca: If you reload the page in `onCancel`, that defeats the very reason to give the user a choice between logging in again immediately or not. The purpose of being able to cancel is to do something to save unfinished work. Imagine he was writing some long text, for example. He could at least copy that text and then relogin.

heruan: If we permit the user to stay on an inconsistent state, e.g. a form on a protected page, we give the false impression of being still logged in while we're actually unauthenticated. That means that if the user finishes their work and submits the form, they'll most probably get an error and lose their work anyways.

Saving the form content on a local cache so that you can recover if after logging in would be a better solution (e.g. like GitHub does), but that's out the scope of the SSO Kit documentation.

cromoteca: Yes it should be clearly stated in the dialog text that the state will be inconsistent if staying on the page. Anyway, as the local state becomes unauthenticated, the developer still has the option to do actions like disabling the submit button. Our goal is to offer useful tools and reasonable defaults.

Another solution could be to implement, instead of a confirm dialog, a modal dialog with an iframe that loads the login form, so that the user can relogin without losing the page and its unsaved changes.

jouni:
I tried looking at the conflicts and merging them, but I’m not confident in deciding what the correct end result should be. 

@heruan, you need to get someone from the team to look at this to get it merged.

@russelljtdyer, there were new content additions after your previous commit, so this one would benefit from another language review round. But let’s wait for the conflicts to get resolved first.

heruan:
This is still pending since @tamasmak has local work and he's away this week, I expect this to be ready for final language-check once he's back on Monday 🙏

heruan (approved):

heruan:
@jouni 
> I tried looking at the conflicts and merging them, but I’m not confident in deciding what the correct end result should be.

Conflicts are resolved, this is now ready for final language-check and merge

PR #2283, opened 2023-03-15
Remove specific quota / pricing information

emarc:
Remove information about specific quotas and prices, as they were not in sync with the pricing page.

### DIFF
--- a/articles/tools/collaboration/going-to-production.asciidoc
+++ b/articles/tools/collaboration/going-to-production.asciidoc
@@ -1,5 +1,7 @@
 ---
 title: Production Configuration
+page-title: Production Configuration
+description: Preparing a Vaadin application for production that uses Collaboration Kit.
 order: 4
 layout: page
 ---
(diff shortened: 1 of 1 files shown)


heruan (changes requested):
Thanks for fixing this! There are also references in [`developing-with-ce.asciidoc`](https://github.com/vaadin/docs/blob/a13cb1289189eb236cfa84c2ad5d9203a1c981e0/articles/tools/collaboration/developing-with-ce.asciidoc?plain=1#L29-L51) and in [`index.asciidoc`](https://github.com/vaadin/docs/blob/a13cb1289189eb236cfa84c2ad5d9203a1c981e0/articles/tools/collaboration/index.asciidoc?plain=1#L17-L21) that need to be adjusted.

jouni:
> The title of this documentation page is Production Configuration, but it seems almost exclusively about licenses.

Yeah, I suppose this is mainly due to “legacy”, as the previous pricing model always required a license when using Collaboration Kit in production. So I suppose it made sense to include that in the article/page that describes how to use CK in production.

> It also seems to be only for an application that uses Collaboration Kit

Well, yes, since this page is under the “Collaboration Kit” section. My assumption is that the context would be clear to the user, but perhaps not.

articles/tools/collaboration/going-to-production.asciidoc
```diff
@@ -1,71 +1,62 @@
 ---
 title: Production Configuration
+page-title: Production Configuration
```
jouni, line 3:
```suggestion
title: Configuring Collaboration Kit for Production
```

jouni: `page-title` is only ever needed if you want to override the generated `<title>` element string. For this page the generated one will be (with the suggested title) "Configuring Collaboration Kit for Production | Collaboration Kit | Tools | Vaadin Docs".

PR #2299, opened 2023-03-22
update to Vaadin 24.0.1

ZheSun88:
also, correct the `LocalDate` import .


package.json
```diff
@@ -94,6 +95,7 @@
     "@rollup/plugin-replace": "5.0.2",
     "@rollup/pluginutils": "5.0.2",
     "@types/iframe-resizer": "^3.5.9",
+    "@vaadin/vaadin-development-mode-detector": "2.0.6",
```
jouni, line 98: Probably not very important, but I’m wondering why is this dependency listed twice, once in normal deps and then in dev deps?

jouni (approved):

PR #2310, opened 2023-03-26
fix: incorrect description of where to find "edit page" link

bwajtr:
Is it possible that the link moved from upper-right corner to the bottom-right corner? Or do I missed something?
There is definitely a link for page editing in bottom right part of every docu page (at least for V24)


jouni (approved):
The link used to be in the top-right corner many years ago.

PR #2358, opened 2023-04-12
Use `mvn package` instead of install

tarekoraby:
No need to require installation to local repo.


jouni (approved):
I assume this is fine/correct. Someone else can correct it later if needed.

PR #2359, opened 2023-04-12
removed obsoleted workaround

mstahv:


articles/tutorial/unit-and-integration-testing.adoc
```diff
@@ -180,7 +180,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 
 
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT) // <1>
```
knoobie, line 183: The option random port is still king on CI :)

mstahv: It is and still used here: https://github.com/vaadin/flow-crm-tutorial/blob/v24/src/test/java/com/example/application/it/LoginE2ETest.java#L15

But this test don't need the port at all (not a test using a browser).

jouni (approved):

PR #2362, opened 2023-04-13
docs: grid lazy column rendering mode documentation

tomivirkki:
Add documentation for the Grid column rendering feature.

Closes https://github.com/vaadin/platform/issues/3909


articles/components/grid/index.asciidoc
```diff
@@ -333,6 +333,39 @@ include::{root}/src/main/java/com/vaadin/demo/component/grid/GridColumnWidth.jav
 ----
 --
 
+[role="since:com.vaadin:vaadin@V24.1"]
+=== Large Number of Columns
```
rolfsmeds, line 337: I think this section would make more sense as a sub-section of the Lazy Loading section.

articles/components/grid/index.asciidoc
```diff
@@ -333,6 +333,39 @@ include::{root}/src/main/java/com/vaadin/demo/component/grid/GridColumnWidth.jav
...
+
+When dealing with a large number of columns in the Grid, performance can become a concern.
```
rolfsmeds, line 339:
```suggestion
Grids with large numbers of columns can exhibit performance issues.
```

articles/components/grid/index.asciidoc
```diff
@@ -333,6 +333,39 @@ include::{root}/src/main/java/com/vaadin/demo/component/grid/GridColumnWidth.jav
...
+To optimize the rendering of cells in the grid for such scenarios, Vaadin Grid provides the "column rendering" feature with a special "lazy" rendering mode.
```
rolfsmeds, line 340:
```suggestion
If many of the columns are typically outside the visible viewport, rendering performance can be optimized by using a "lazy column rendering" mode.```

File: articles/components/grid/index.asciidoc [hunk 4]
```diff
@@ -333,6 +333,39 @@ include::{root}/src/main/java/com/vaadin/demo/component/grid/GridColumnWidth.jav
...
+
+- *Row Height:* When only a number of columns are visible at once, the height of a row can only be that of the highest cell currently visible on that row. Make sure each cell on a single row has the same height as all other cells on that row. If row cells have different heights, users may experience jumpiness when scrolling the grid horizontally as lazily rendered cells with different heights are scrolled into view.
+
+- *Auto-width Columns:* For the columns that are initially outside the visible viewport but still use auto-width, only the header content is taken into account when calculating the column width because the body cells of the columns outside the viewport are not initially rendered.
+
+- *Screen Reader Compatibility:* Screen readers may not be able to associate the focused cells with the correct headers when only a subset of the body cells on a row is rendered.
+
+- *Keyboard Navigation:* Tabbing through focusable elements inside the grid body may not work as expected because some of the columns that would include focusable elements in the body cells may be outside the visible viewport and thus not rendered.
+
```
rolfsmeds, line 355:
```suggestion
- *No performance improvement if all columns are visible:* The lazy column rendering mode can only improve the rendering performance of when a significant portion of the columns are outside of the Grid's visible viewport. It has no effect on Grids where all columns are visible without horizontal scrolling.
- ```

articles/components/grid/index.asciidoc
tomivirkki: Moved under "Lazy loading" for now

rolfsmeds: Well... lazy loading is not only about performance – it can also be simply impossible to load the entire dataset into memory up front. And I do think "lazy loading" is what most developers would be looking for.

But I agree that "Large number of columns" is not very descriptive. Proposals:

- Lazy rendering of columns / Lazy column rendering
- Optimizing performance with many columns

As an alternative to having the section under Lazy Loading section, we could add a Performance Optimization section to the bottom, with e.g. the following topics

- Lazy column rendering
- Using template renderers instead of component renderers
- ...maybe something else?

tarekoraby: I have no strong opinion.

rolfsmeds (approved):

PR #2373, opened 2023-04-19
Refactor contributing guides and update DSP docs

jouni:
Combine all contribution guides under one section (both code and docs). 

Update docs contribution guides to reflect our current best practices as much as possible (still missing some things).

Move all DSP related docs from docs authoring guides to DSP guides.


jouni:
Redirect rule from /contributing-docs → /contributing/docs pending approval.

PR #2383, opened 2023-04-25
Update download links to point to v2.0.0

MatthewVaadin:
The download buttons for the Observability Kit agent do not point to the correct versions.
This updates them to V2.0.0. Note: This will need to be changed to V2.1.0 for V24.1 of the docs.


MatthewVaadin:
I've removed the language label as there are no language changes.

Mikaelsu (approved):
Tested both links, all good!

PR #2386, opened 2023-04-25
feat(hilla): document Tree Grid and Multi-Select Combo Box

### LINKED ISSUE #2363: Document using tree grid with React
Vaadin Grid React component supports using tree data with a tree on par with the web component, as documented here: https://vaadin.com/docs/latest/components/tree-grid

Hilla docs website is missing the Tree Grid use case.

Let's document using a tree grid in React with a respective article in Hilla React components section.

cromoteca:
Closes #2363


articles/react/components/tree-grid/index.adoc
```diff
@@ -0,0 +1,67 @@
...
+= Tree Grid
+
+Tree Grid is an extension of the Grid component. All of Grid's features are available in Tree Grid. It isn't meant, though, to be used as a navigation menu.
+
+
+[source,tsx]
+----
+const people = usePeople();
+
+const dataProvider = async (
```
platosha, line 15: It would be great to wrap using `useCallback` or something, otherwise re-assigning the data provider to the grid on every render causes unexpected side-effects, like this one: https://github.com/vaadin/react-components/issues/74

platosha (changes requested):

articles/react/components/tree-grid/index.adoc
```diff
@@ -0,0 +1,67 @@
...
+) =>
+  callback(
+    people.filter(
+      (p) => !params.parentItem || p.managerId === params.parentItem.id
+    )
+  );
+
+return (
+  <Grid
+    items={people}
```
platosha, line 27: AFAIK, the `dataProvider`  overrides the `items`, so perhaps we should not have them both in one example.

```suggestion
    items={people}
```

platosha (changes requested):

articles/react/components/tree-grid/index.adoc
cromoteca: Done

articles/react/components/tree-grid/index.adoc
cromoteca: It should, but it doesn't work if I remove it. Do we have a bug? Maybe the callback is not called for the root?

platosha: Typically this is caused by not informing the grid of the total number of items. There are several options to do so:
- Give the `items`, then `items.length` is what the grid uses as the total number.
- Give the `size` number, then the grid uses it. 
- The data provider callback provides the total size in the callback: `callback(requestedSlice, totalSize);`

The data provider here only provides the requested slice but not the total. I guess the grid happened to use the `items.length` number as a partial fallback, and this is why it requires `items` with this setup.

cromoteca: You're right, thanks for the insight.

articles/react/components/tree-grid/index.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+const dataProvider = useCallback(
+  async (
+    params: GridDataProviderParams<Person>,
+    callback: GridDataProviderCallback<Person>
+  ) => {
+    const filtered = people.filter(
+      (p) => !params.parentItem || p.managerId === params.parentItem.id
+    );
+    return callback(filtered, filtered.length);
```
platosha, line 23: One more caveat here on my mind. The data provider is supposed to take care of paged data, as well as applying sorting and filtering. Omitting sorting and filtering here is OK, as it these features are not used anyway, but taking care of supplying more than one page of data feels essential to me. Otherwise it might suddenly break when the number of items exceeds the page size (25 by default) on any level.

The expectation is that every data provider call retrieves the data only for the requested page number, but the size of the entire data on the current tree level. 

From [the lazy loading Web Component example](https://vaadin.com/docs/latest/components/grid/#lazy-loading), the relevant step missing here is:
```typescript
  // Pagination
  const count = result.length;
  result = result.slice(page * pageSize, pageSize);
```

The `page` and `pageSize` should be available from the data provider's `params`.

Should look something like this:

```suggestion
    const { page, pageSize } = params;
    const pageOffset = page * pageSize;
    return callback(filtered.slice(pageOffset, pageSize), filtered.length);
```

cromoteca: I understand, but I simplified the example as I don't have the same `getPeople` API here in place. I'll see if I can replicate everything from the main docs.

articles/react/components/tree-grid/index.adoc
```diff
@@ -0,0 +1,67 @@
...
+  async (
+    params: GridDataProviderParams<Person>,
+    callback: GridDataProviderCallback<Person>
+  ) => {
+    const pagedPeople = usePagedPeople({
+      count: params.pageSize,
+      startIndex: params.page * params.pageSize,
+      managerId: params.parentItem ? params.parentItem.id : null,
+    });
+    console.log(pagedPeople);
```
platosha, line 23: Looks like the log line was left by accident.

```suggestion
```

platosha (approved):

PR #2408, opened 2023-05-08
Fix issue where .ts samples only render once

### LINKED ISSUE #2407: Horizontal and Vertical Layout TypeScript examples render only once
After opening either the Horizontal or Vertical Layout page once, after you navigate away and come back, the TypeScript examples no longer render. They throw the following JS error (at least in Safari):

Screenshot 2023-05-08 at 9 59 23

The error comes from setting the classname in the constructor in each example: https://github.com/vaadin/docs/blob/latest/frontend/demo/component/basiclayouts/basic-layouts-vertical-layout-horizontal-alignment.ts#L12

That line should be moved to either `connectedCallback()` or `firstUpdated()` callbacks.

Mikaelsu:
Closes #2407

### DIFF
--- a/articles/components/horizontal-layout/index.asciidoc
+++ b/articles/components/horizontal-layout/index.asciidoc
@@ -11,7 +11,7 @@ page-links:
 // tag::description[]
 Horizontal Layout places components side-by-side in a row.
 // end::description[]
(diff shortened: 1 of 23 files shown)


Mikaelsu:
Removed the language check label since this PR is 99% code.

jouni (approved):

PR #2409, opened 2023-05-09
Improve Hilla SEO

Mikaelsu:
Some optimization prompted by our SEO agency.

### DIFF
--- a/articles/lit/components/index.adoc
+++ b/articles/lit/components/index.adoc
@@ -1,6 +1,6 @@
 ---
 title: Components
-description: Hilla helps you build professional-looking UIs faster by including more than 45 UI components, provided by Vaadin, ranging from text fields to data grids.
+description: Hilla is a framework for building reactive web applications on Java backends. It seamlessly integrates a Lit TypeScript frontend with a Spring Boot backend using type-safe server communication, integrated tooling, and accessible web components. Scroll to explore Hilla’s Lit components library.
 order: 30
 section-nav: flat expanded
 ---
@@ -12,6 +12,8 @@ section-nav: flat expanded
 image::components.png["An illustration of a chart, a data grid, and a date calendar UI component",width=400]
 
 https://vaadin.com/docs/components[Browse all Vaadin components, role="button secondary water"]
+
+== Hilla’s Lit Components Library
 
 In addition to the built-in components, you can use any web components and JavaScript components.
 
--- a/articles/lit/components/index.adoc
+++ b/articles/lit/components/index.adoc
@@ -1,5 +1,6 @@
 ---
-title: Lit Web Components
+title: Components
+page-title: Lit Web Components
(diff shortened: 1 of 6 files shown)


articles/lit/components/index.adoc
```diff
@@ -1,11 +1,11 @@
 ---
-title: Components
+title: Lit Web Components
 description: Hilla helps you build professional-looking UIs faster by including more than 45 UI components, provided by Vaadin, ranging from text fields to data grids.
 order: 30
 section-nav: flat expanded
 ---
 
-= Components
+= Lit Web Components
```
jouni, line 8: Do we really need to update this text as well? Did the SEO agency explicitly want us to change the `<h1>` as well, not just the “meta title”?

articles/lit/components/index.adoc
```diff
@@ -1,11 +1,11 @@
 ---
-title: Components
+title: Lit Web Components
```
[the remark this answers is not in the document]
jouni: This is inconsistent with the React counterpart. But I assume that is intentional, that people search for “Lit web components” more than “Lit UI component library”?

articles/react/components/index.adoc
```diff
@@ -1,12 +1,15 @@
 ---
-title: Components
+title: React UI Component Library
+description: Hilla is a framework for building reactive web applications on Java backends. It seamlessly integrates a React TypeScript frontend with a Spring Boot backend using type-safe server communication, integrated tooling, and accessible UI components. Scroll to explore Hilla’s React component library.
 order: 40
 section-nav: expanded flat
 ---
 
-= Components
+= React UI Components
```
jouni, line 8: Same comment here, that do we really need to change the `<h1>` in addition to the `<title>`?

articles/lit/components/index.adoc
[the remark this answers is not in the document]
Mikaelsu: Intentional, yes, and also true that I meant to change the `<title>` element. Thanks!

articles/react/components/index.adoc
```diff
@@ -1,5 +1,7 @@
 ---
 title: Components
+page-title: React UI Component Library
```
jouni, line 3: I wonder if we should still add “Hilla” explicitly in the title as well? When the `page-title` frontmatter atteribute is used, it overrides the default generated title, and the value of the attribute is used directly (not appended/prepended with anything else).

```suggestion
page-title: React UI Component Library | Hilla Docs
```

articles/lit/components/index.adoc
```diff
@@ -1,5 +1,6 @@
 ---
 title: Components
+page-title: Lit Web Components
```
jouni, line 3:
```suggestion
page-title: Lit Web Components | Hilla Docs
```

articles/react/components/index.adoc
Mikaelsu: Good call!

PR #2413, opened 2023-05-10
feat: production bundling

caalador:
Default producion bundle,
creating optimized bundle,
lazy loading

Closes vaadin/flow#16754

### DIFF
--- a/articles/advanced/lazy-loading.asciidoc
+++ b/articles/advanced/lazy-loading.asciidoc
@@ -1,0 +1,41 @@
+---
+title: Lazy loading component JavaScript
+description: How lazy loading of JavaScript works and can be configured.
+order: 800
+layout: page
+---
+
+= Lazy loading JavaScript
+
+A production build will scan for `Routes` and make lazy loaded bundles of the components used in the routes.
+
+By default, only the routes `""` and `"login"` are eager and every other route components will be lazy loaded.
+
+[NOTE]
+Any components that are loaded using reflection should be named on the `Route` class using `@Uses` so that they get collected.
+
+=== Configuring eager routes
+
+To configure which views should load eagerly use the annotation `@LoadDependenciesOnStartup` on the `AppShellConfiguration` class.
+
+Only defining `LoadDependenciesOnStartup` will make all routes eager.
+
+[source,java]
+----
+@LoadDependenciesOnStartup
+public class Configuration implements AppShellConfigurator {
+}
+----
+
+To configure specific routes to be eagerly loaded add the route class to the value array for instance:
+
+[source,java]
+----
+@LoadDependenciesOnStartup({GeneralInfo.class, DataSearch.class})
+public class Configuration implements AppShellConfigurator {
+}
+----
+
+This will make components, scripts and css used in `GeneralInfo` and `DataSearch` loaded immediately on bootstrap and any other components used in other view as they are needed.
+
+[discussion-id]`B794EFE2-B833-466A-ACA9-7F60AA6D4332`
--- a/articles/guide/start/gradle.asciidoc
+++ b/articles/guide/start/gradle.asciidoc
@@ -484,6 +484,10 @@ Defaults to `false`.
 For production, the frontend is bundled and optimized, as described in <<{articles}/production#,Deploying to Production>>.
 Running the `vaadinBuildFrontend` task automatically switches this to `true`, so there is no need to configure anything.
 
+`forceProductionBuild: Boolean = false`::
+Whether to generate a production bundle even if an existing pre-generated bundle could be used.
+'true' will force bundle generation without validation of existing production bundle.
+
 `frontendOutputDirectory: File = null`::
 The folder where Vite should output [filename]`index.js` and other generated files.
 Defaults to `null`, which uses the automatically detected value of the main SourceSet, usually `build/resources/main/META-INF/VAADIN/webapp/`.
--- a/articles/production/production-build.adoc
+++ b/articles/production/production-build.adoc
@@ -6,15 +6,26 @@ description: Create an optimized build of your application for production deploy
 
 = Production Build
 
-To create a production build, run the following command:
+Since Vaadin 24.1 there is a default production bundle that can be used to not have to run frontend tools like npm and vite if no add-ons or frontend files are used in the application.
+
+The bundle can be optimized to use <<{articles}/advanced/lazy-loading#,lazy loading>> of components by building an optimized bundle.
+
+To create an optimized production build, run the following command:
+
+[source,terminal]
+----
+mvn clean package -Pproduction -Dforce.production.build=true
+----
+
+This builds a `JAR` or `WAR` file with all the dependencies and bundled front-end resources, ready to be deployed.
+The file can be found in the `target` folder after the build completes.
+
+To generate a production build by using a pre-created production bundle that is applicable for the application run:
 
 [source,terminal]
 ----
 mvn clean package -Pproduction
 ----
-
-This builds a `JAR` or `WAR` file with all the dependencies and bundled front-end resources, ready to be deployed.
-The file can be found in the `target` folder after the build completes.
 
 == Enabling Production Builds
 The production build command should work out of the box for Vaadin starter projects, for example projects that are generated with https://start.vaadin.com.
@@ -180,6 +191,9 @@ For `JAR` packaging, it's `target/classes/META-INF/resources/build`.
 
 `useGlobalPnpm` (default: `false`)::
     Whether to use a globally installed `pnpm` tool instead of the default supported version of `pnpm`.
+
+`forceProductionBuild` (default: `false`)::
+    Whether to generate a production bundle even if an existing pre-generated bundle could be used.
 
 === clean-frontend
 This goal removes files that may cause inconsistencies when changing versions.


articles/advanced/lazy-loading.asciidoc
```diff
@@ -0,0 +1,41 @@
+---
+title: Lazy loading component JavaScript
+description: How lazy loading of JavaScript works and can be configured.
+order: 800
+layout: page
```
jouni, line 5: This attribute is unnecessary, no need to add it.

```suggestion
```

articles/advanced/lazy-loading.asciidoc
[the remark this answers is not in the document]
mshabarov: Thumb up for moving lazy loading section into `production-build` chapter.

articles/advanced/lazy-loading.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+---
+
+= Lazy Loading JavaScript
+
+A production build scans for `Routes` and make lazy loaded bundles of the components used in the routes.
+
+By default, only the routes `""` and `"login"` are eager and every other route components become lazy loaded.
+
+[NOTE]
+Any components that are loaded using reflection should be named on the `Route` class using `@Uses` so that they get collected.
```
mshabarov, line 14: Code example would be useful, even a simple one:
```java
@Route("my-view")
@Uses(Button.class)
public class MainView extends Div {
    public MainView() {
        try {
            Class<? extends Button> buttonClass = Class.forName(
                    "com.vaadin.flow.component.button.Button");
            Button button = buttonClass.getDeclaredConstructor().newInstance();
            add(button);
        } catch (ClassNotFoundException e) {
            // handle exception
        }
    }
}
```

articles/production/production-build.adoc
```diff
@@ -6,16 +6,27 @@ description: Create an optimized build of your application for production deploy
...
 [source,terminal]
 ----
-mvn clean package -Pproduction
+mvn clean package -Pproduction -Dforce.production.build=true
 ----
 
 This builds a `JAR` or `WAR` file with all the dependencies and bundled front-end resources, ready to be deployed.
 The file can be found in the `target` folder after the build completes.
 
+To generate a production build by using a pre-created production bundle that is applicable for the application run:
```
mshabarov, line 23:
```suggestion
To generate a production build by using a pre-compiled production bundle that is applicable for the application run:
```

articles/guide/start/gradle.asciidoc
```diff
@@ -484,6 +484,10 @@ Defaults to `false`.
 For production, the frontend is bundled and optimized, as described in <<{articles}/production#,Deploying to Production>>.
 Running the `vaadinBuildFrontend` task automatically switches this to `true`, so there is no need to configure anything.
 
+`forceProductionBuild: Boolean = false`::
+Whether to generate a production bundle even if an existing pre-generated bundle could be used.
```
mshabarov, line 488:
```suggestion
Whether to generate a production bundle even if an existing pre-compiled bundle could be used.
```

articles/advanced/lazy-loading.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+
+=== Configuring Eager Routes
+
+To configure which views should load eagerly use the annotation `@LoadDependenciesOnStartup` on the `AppShellConfiguration` class.
```
mshabarov, line 18: Is it a case now that if you don't set forceProductionBuild parameter to true, your eager customisations wouldn't be applied in case if Flow decides to use pre-compiled bundle?
If it is so, I propose to emphasise that one needs to set this parameter to true to have his custom eager loadings work.

articles/production/production-build.adoc
```diff
@@ -6,16 +6,27 @@ description: Create an optimized build of your application for production deploy
 
 = Production Build
 
-To create a production build, run the following command:
+The default production bundle can be used to not have to run front-end tools like npm and Vite if no add-ons or front-end files are used in the application.
```
mshabarov, line 9:
```suggestion
The pre-compiled production bundle can be used to not have to run front-end tools like npm and Vite if no add-ons or front-end files are used in the application.
```

articles/production/production-build.adoc
```diff
@@ -6,16 +6,27 @@ description: Create an optimized build of your application for production deploy
...
+
```
mshabarov, line 10: I propose to give here a link (from dev configuration chapter) or to list explicitly the reasons when Flow decides to make a new prod bundle.

mshabarov (changes requested):

articles/advanced/lazy-loading.asciidoc
mshabarov: Did some testing.
When pre-compiled bundle is used, `@LoadDependenciesOnStartup` annotation doesn't have any effect - "heavy" components are loaded on demand: Map, Charts, Spreadsheet and RichTextEditor.
(we may want to change it later, so whenever this annotation is set, we always trigger a new bundle).

I would maybe also mention that with pre-compiled bundle all the components are loaded eagerly, excluding these 4 components. And with a custom bundle only components that are used on opened views/routes are loaded. E.g. if route A uses button and text field, only these two are loaded. If I have route B which has a combo box, then combo box is loaded once I navigate to B. This is maybe evident for us as a developers of this feature, but nevertheless it's good to give an explanation for it.

articles/advanced/lazy-loading.asciidoc
caalador: added

articles/advanced/lazy-loading.asciidoc
caalador: Fix in Flow to generate an optimized bundle if LoadDependenciesOnStartup annotation is used

articles/production/production-build.adoc
```diff
@@ -6,16 +6,36 @@ description: Create an optimized build of your application for production deploy
...
-To create a production build, run the following command:
+The pre-compiled production bundle can be used to not have to run front-end tools like npm and Vite if no add-ons or front-end files are used in the application.
+
+Vaadin generates an application bundle in the following situations:
+
+- An npm/pnpm package is added with `@NpmPackage` or directly into [filename]`package.json`;
+- CSS or JavaScript is added with `@CssImport`, `@JsModule` or `@JavaScript`;
+- Vaadin add-on with front-end customizations is added;
+- Custom theme packaged as a `JAR` dependency is added, if it defines any assets to be added to the project;
+- Exported web component is added.
```
mshabarov, line 17:
```suggestion
- Exported web component is added;
```

articles/production/production-build.adoc
```diff
@@ -198,6 +221,59 @@ The `clean-frontend` goal supports the same parameters as `prepare-frontend`.
 === dance
 This goal is synonymous with the `clean-frontend` goal.
 
+=== Lazy Loading JavaScript
```
mshabarov, line 224: I propose to reorganise it a bit:
let this be with indentation `==` and title "Bundle Chunks Loading Optimisations" with two sub chapters "Lazy Loading (On Demand)" and "Eager Loading"

articles/production/production-build.adoc
```diff
@@ -198,6 +221,59 @@ The `clean-frontend` goal supports the same parameters as `prepare-frontend`.
...
+
+A production build scans for `Routes` and make lazy loaded bundles of the components used in the routes.
+
+By default, only the routes `""` and `"login"` are eager and every other route components become lazy loaded.
+
+With the default production bundle all components are loaded eagerly apart from the heavy components `Map`, `Charts`, `Spreadsheet` and `RichTextEditor`.
```
mshabarov, line 230:
```suggestion
With the pre-compiled production bundle all components are loaded eagerly apart from the heavy components `Map`, `Charts`, `Spreadsheet` and `RichTextEditor`.
```

mshabarov (changes requested):

articles/production/production-build.adoc
```diff
@@ -198,6 +221,59 @@ The `clean-frontend` goal supports the same parameters as `prepare-frontend`.
 === dance
 This goal is synonymous with the `clean-frontend` goal.
 
+=== Lazy Loading JavaScript
+
+A production build scans for `Routes` and make lazy loaded bundles of the components used in the routes.
```
mshabarov, line 226:
```suggestion
A production build scans for `Routes` and loads lazily the components (included into bundle's chunks) used in the routes.
```

mshabarov, line 226:
```suggestion
A production build scans for `Routes` and loads lazily the components (included into bundle's chunks) used in the routes.
```

articles/production/production-build.adoc
caalador: Removed all `;` as they look strange on the ending of list items

articles/production/production-build.adoc
caalador: Went with the higher level `Bundle Component Loading Optimisations` as one doesn't actually touch chunks.

articles/production/production-build.adoc
caalador: omitted talk about chunks as that is not important and makes it more confusing than helpful

mshabarov (approved):

articles/production/production-build.adoc
```diff
@@ -6,16 +6,36 @@ description: Create an optimized build of your application for production deploy
...
+- `@LoadDependenciesOnStartup` is defined for `AppShellConfigurator`
+
+The bundle can be optimized to use <<#_lazy_loading_javascript,lazy loading>> of components by forcing a build of the production bundle.
+
+To force creation of an optimized production bundle, run the following command:
 
 [source,terminal]
 ----
-mvn clean package -Pproduction
+mvn clean package -Pproduction -Dforce.production.build=true
```
mshabarov, line 26: Suggestion from DX tester: mention `forceProductionBuild` also here and give an example how to set it with the `vaadin-maven-plugin`.

mshabarov, line 26: This is might be a separate PR, but this parameter listed below in the mvn example snipped
```
        <properties>
            <vaadin.productionMode>true</vaadin.productionMode>
        </properties>
```
has no effect anymore, so should be removed.

mshabarov (changes requested):

articles/production/production-build.adoc
caalador: removed

caalador: added

mshabarov (approved):

articles/production/production-build.adoc
tarekoraby: I think it's always better to state the simple use case first, and then suggest the more advanced use case after that. Here, it would be better to state the simple case first without `forceProductionBuild`, and then clarify to advanced users how they can accomplish the more complex thing.

tarekoraby (approved):

caalador:
Does make it clearer.

PR #2414, opened 2023-05-10
docs: new map features

sissbruecker:
Adds documentation for new map features:
- Marker texts
- Marker drag&drop

~~Marking as draft, as the marker text APIs have been renamed, and the corresponding example needs to be updated after releasing 24.1.0-alpha10.~~
Version has been updated and the PR has been rebased.


src/main/java/com/vaadin/demo/component/map/Icons.java
```diff
@@ -2,38 +2,40 @@
 
 import com.vaadin.flow.component.map.configuration.style.Icon;
 
-public class Icons {
```
sissbruecker, line 5: Ran into an issue with the marker icon in the new example not being displayed. There is a flow-components issue about `Icon` instances not being reusable across multiple UIs, which I think applies to the docs setup as well. To fix that I refactored this class to create new icon instances, and updated the remaining examples accordingly.

ugur-vaadin (approved):

PR #2417, opened 2023-05-12
Update properties.asciidoc

samie:
We went through with @mshabarov and clarified all of the available properties and their purpose. Still want to reorganize into common, dev mode and production mode, but will do that as a separate step.


articles/configuration/properties.asciidoc
```diff
@@ -102,100 +102,94 @@ If you use Spring Boot, you should add the `vaadin.` prefix, for example, `vaadi
...
 |`true`
 
 |`disable.automatic.servlet.registration`
-|Configuration name for the parameter that determines whether Vaadin should automatically register servlets that are required for the application to work.
+|Disable automatic registration of servlets that are needed for Vaadin application to work. If disabled, you should register Vaadin servlets yourself.
 |`false`
 
 |`disable-xsrf-protection`
-|Cross-site request forgery protection. The protection is enabled by default, but you may want to disable it for a certain type of testing. The check can be disabled by setting the init parameter.
+|Disbale cross-site request forgery protection. The protection is enabled by default, It is not recommended to disable protection, but you may want to disable it for a certain type of testing.
```
jouni, line 137:
```suggestion
|Disable cross-site request forgery protection. The protection is enabled by default. You should keep it enabled, but you may want to disable it for a certain type of testing.
```

articles/configuration/properties.asciidoc
```diff
@@ -102,100 +102,94 @@ If you use Spring Boot, you should add the `vaadin.` prefix, for example, `vaadi
...
 |`false`
 
 |`eagerServerLoad`
-|If the client-side bootstrap page should include the initial UIDL fragment.
+|Enable the client-side bootstrap page should also include the initial UIDL JSON data fragment.
```
jouni, line 141: I wonder if readers know what UIDL is, and if we should avoid using that term. I’m not sure it’s explained anywhere.

articles/configuration/properties.asciidoc
```diff
@@ -102,100 +102,94 @@ If you use Spring Boot, you should add the `vaadin.` prefix, for example, `vaadi
...
-|300 seconds (this equals 5 minutes)
+|UIs that are open on the client-side send a regular heartbeat to the server indicating that they are still alive, even when there is no ongoing user interaction. When the server doesn't receive a valid heartbeat from a given UI within a certain amount of time, it removes that UI from the session. The interval value is expressed in `seconds`. See also `closeIdleSessions`.
+|300 seconds (5 minutes)
 
 |`i18n.provider`
-|I18N provider property. To use localization and translation strings, the application only needs to implement `I18NProvider` and define the fully qualified class name in the property `i18n.provider`. See the <<{articles}/advanced/i18n-localization#, Localization>> documentation for more details.
+|Fully qualified class name of I18N provider property. To use localization and translation of strings, the application should implement `I18NProvider` interface and define the class name in the property `i18n.provider`. See the <<{articles}/advanced/i18n-localization#, Localization>> documentation for more details.
```
jouni, line 153:
```suggestion
|Fully qualified name of internationalization provider class. To use localization and translation of strings, the application should implement `I18NProvider` interface and define the class name in the property `i18n.provider`. See the <<{articles}/advanced/i18n-localization#, Localization>> documentation for more details.
```

articles/configuration/properties.asciidoc
```diff
@@ -102,100 +102,94 @@ If you use Spring Boot, you should add the `vaadin.` prefix, for example, `vaadi
...
 |`-1` (no timeout)
 
 |`pushMode`
-|The permitted values are "disabled", "manual" or "automatic". See <<{articles}/advanced/server-push#, Server Push>> for more information.
+|Enable the server push. The permitted values are "disabled", "manual" or "automatic". See <<{articles}/advanced/server-push#, Server Push>> for more information.
 |`disabled`
 
 |`pushServletMapping`
-|Sets the servlet mapping used for bidirectional ("push") client-server communication.
+|Servlet mapping used for bidirectional ("push") client-server communication. Some Java application servers require special context e.g. for websockets and you can specify it here.
```
jouni, line 177: Apparently we should add `websockets?` to the vocabulary.

articles/configuration/properties.asciidoc
samie: True, it is legacy. It was there, so didn't want to take it away either.

samie: I decided to remove it for now.

PR #2421, opened 2023-05-15
Add compatibility matrix for V14

Mikaelsu:


articles/flow/compatibility.adoc
```diff
@@ -0,0 +1,63 @@
...
+= List of Supported Technologies
+
+This page shows the minimum required version of supported technologies for *Vaadin {moduleMavenVersion:com.vaadin:vaadin}*. See the https://github.com/vaadin/platform/releases/tag/{moduleMavenVersion:com.vaadin:vaadin}[release notes] for more details.
+
+[cols="1,1"]
+|===
+|Technology|Version
+
+| Java| 8, 11
+| Spring Boot| 2.4.5
```
knoobie, line 16: Should this be 2.x? Otherwise Vaadin 14 is only allowed to use already EoL Spring Boot Versions.

Mikaelsu: Very true, thanks again @knoobie!

PR #2424, opened 2023-05-16
chore: upgrade to 24.1.0-alpha6

sissbruecker:
Upgrades Vaadin version to 24.1.0-alpha6 and fixes related Typescript compilation errors.

Apart from that, DSP currently fails to start, see https://github.com/vaadin/docs-app/pull/378. A new version of DSP needs to be released, and the DSP version in this repo needs to be updated.

### DIFF
--- a/dspublisher/dspublisher-scripts.js
+++ b/dspublisher/dspublisher-scripts.js
@@ -5,7 +5,7 @@ const fs = require('fs');
 const path = require('path');
 const http = require('http');
 
-const DSP_VERSION = '2.1.1';
+const DSP_VERSION = '2.1.3';
 
 async function checkPreConditions() {
   try {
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -18,7 +18,6 @@
     "noImplicitReturns": true,
     "noImplicitAny": true,
     "noImplicitThis": true,
-    "noUnusedLocals": true,
```
sissbruecker, line 21: This is currently necessary to prevent Typescript errors in generated Hilla endpoints. These contain an unused import, more info here: https://vaadin.slack.com/archives/CDNTYMS9H/p1683723045631769.

Since the files are generated the error can not be manually ignored, so disabling this check seems to be the only option until the isse is fixed in Hilla.

manolo (approved):

PR #2442, opened 2023-05-24
feat: WebPush tutorial

### LINKED ISSUE #2441: Create initial documentation on webPush usage with Flow
Generate a tutorial style document on setting up webpush for a Flow project.

caalador:
Closes #2441


articles/tutorial/setting-up-webpush.adoc
```diff
@@ -0,0 +1,296 @@
...
+
+=== Generate Service Worker With Web Push Methods
+
+Vaadin provides the `@WebPush` and `@PWA` annotations, that automatically generate the required PWA resources. Add the `@PWA` annotation on [classname]`Application.java` as follows:
+
+.`Application.java`
+[source,java]
+----
+@SpringBootApplication
+@PWA(name = "Vaadin CRM", shortName = "CRM")
```
knoobie, line 75: Wouldn't it make more sense that the PWA annotation gets `enableWebPush=true´instead of using another annotation, which could people confuse that this can work WITHOUT PWA.. even tho PWA is required for Webpush?

caalador: The angle is more that you shouldn't need all the stuff in PWA (except for iPhone I guess) to have web push functionality.
So having only webpush annotation would strip the un-needed pwa parts from the service worker if pwa is not specifically required.
But as too much time has gone with fighting against Idea I haven't gotten that far with the changes.

Mikaelsu (commented):
Is there another place in the documentation where we could put this tutorial? /tutorial is a long series building the same app and this doesn't fit there

caalador:
Configuration after PWA?

Mikaelsu (approved):
Looks good! Thank you, @caalador

PR #2450, opened 2023-05-26
chore: Add vaadin. prefix to recently added config parameters

mshabarov:
`ci.build` and `force.production.build` parameters are used in the plugin classes, where vaadin. prefix is not added by Flow automatically. We need to follow the same rules for naming, so that developers can set these two parameters as `-Dvaadin.force.production.build=true` or `Dvaadin.ci.build=true`.

Depends on https://github.com/vaadin/flow/pull/16882

Related-to https://github.com/vaadin/flow/issues/16880


caalador (approved):

articles/configuration/properties.asciidoc
```diff
@@ -108,20 +108,20 @@ If you use Spring Boot, you should add the `vaadin.` prefix. For example, `vaadi
 |Decide whether pre-compressed https://github.com/google/brotli[Brotli] files should be used if accepted by the browser. Brotli files are created during a production build and the property is only used in production mode. Set to `false` if you want to serve uncompressed static resources.
 |`true`
 
-|`ci.build`
+|`vaadin.ci.build`
```
Mikaelsu, line 111: Probably a dumb question, but should this be `Dvaadin.ci.build`?

caalador (changes requested):
Just noted that properties.asciidoc is missing the `skip.dev.bundle` property.
This can not be found in any documentation for some reason.

mshabarov:
> Just noted that properties.asciidoc is missing the `skip.dev.bundle` property. This can not be found in any documentation for some reason.

Thanks, I'll add it.

articles/configuration/properties.asciidoc
mshabarov: "D" is a JVM prefix for system properties, so it doesn't need to be specified for each particular parameter.

mshabarov:
I made the following modifications comparing to the initial version:

- added `skip.dev.bundle` parameter
- added missing description to 2 parameters where it is logical to place and mention it
- removed `vaadin.` in the properties list, because they all should contain this prefix anyway.

articles/configuration/properties.asciidoc
```diff
@@ -144,6 +144,10 @@ If you use Spring Boot, you should add the `vaadin.` prefix. For example, `vaadi
 |Enable the client-side bootstrap page to include the initial JSON data fragment.
 |`false`
 
+|`force.production.bundle`
```
caalador, line 147:
```suggestion
|`vaadin.force.production.bundle`
```
This is now complex as other parameters are usable from application.properties, but this is specifically only used in maven/gradle plugin and needs to have the vaadin prefix even if you are not using spring boot.

Feels like this should not be under configuration properties, but under a specific header `plugin properties` in this file.

Same goes for the CI build.

Also these are missing from the `Build Front-End Goal Parameters` in articles/configuration/maven.asciidoc

articles/configuration/properties.asciidoc
Mikaelsu: Gotcha, thanks for clarifying!

articles/configuration/properties.asciidoc
mshabarov: Moved them to a new section `Vaadin Plugin Properties` and added also to `Build Front-End Goal Parameters`.

caalador (approved):

PR #2456, opened 2023-05-29
New component thumbnails

rolfsmeds:
They look like this, mmkay?
image

Additionally (only for you!): a card for Tooltip!


Mikaelsu (approved):
Love it!

PR #2462, opened 2023-05-31
docs: remove tree grid scroll to index warning

tomivirkki:
Since https://github.com/vaadin/flow-components/issues/3505 has been completed in V24.1, remove the warning about using `scrollToIndex` method in `TreeGrid`.


Mikaelsu (commented):
Is there any reason to wait for GA before this is merged?

tomivirkki:
> Is there any reason to wait for GA before this is merged?

Not sure. The documentation does already cover some features that will only be available in 24.1. And the issue concerning TreeGrid scrollToIndex specifically is mentioned in the API documentation of 24.0 TreeGrid.

Mikaelsu (approved):

PR #2463, opened 2023-05-31
fix: define bundles and usage

### LINKED ISSUE #2449: Make Vaadin Production Build documentation more clear around the terms dev-bundle and production bundle
The documentation about the new features in V24.0 and V24.1 that makes it possible to skip unnecessary frontend builds and leverage from development/production bundles whenever possible needs more clarification around the terms the dev-bundle, production-bundle, pre-generated and so on. It is good to have explicit definition of their meaning and the situation they are trying to address. 

For example, the [Enabling Production Builds](https://vaadin.com/docs/latest/production/production-build/#enabling-production-builds) has the following comments in the code-snippet:
```
<!-- To always force a production build set this configuration to 'true' -->
<!-- To possibly use the pre-generated bundle remove or set to 'false' -->
```
Not having clear definition around the mentioned terms could lead to misunderstanding of what exactly `pre-generated bundle` refers to: the bundle that is built earlier when the user ran the project in development mode, or the one that shipped with Vaadin.

caalador:
Add definition on when and what
bunlde is used in production.

Closes #2449


mshabarov (approved):

articles/production/production-build.adoc
```diff
@@ -33,6 +33,13 @@ If needed, you can force the creation of an optimized production bundle by execu
 mvn clean package -Pproduction -Dforce.production.build=true
 ----
 
+[NOTE]
+The pre-compiled production bundle is not the same as a pre-compiled front-end bundle for development mode as it has debugging features enabled.
+
+[NOTE]
+The output from an optimized production bundle will not be used as a pre-compiled bundle for new later production builds.
```
taefi, line 40: Due to my rusty English, does this "new later production builds" mean those production builds that are triggered after the first optimized production build?

articles/production/production-build.adoc
```diff
@@ -63,8 +70,8 @@ The production build command works out-of-the-box for Vaadin starter projects. F
                         </execution>
                     </executions>
                     <configuration>
-                        <!-- To always force a production build set this configuration to 'true' -->
-                        <!-- To possibly use the pre-generated bundle remove or set to 'false' -->
+                        <!-- To always force an optimized production bundle set this configuration to 'true' -->
```
taefi, line 73: Is a "building" missing after force? 
To always force building an optimized production bundle...

articles/production/production-build.adoc
caalador: Yes. Would it make it clearer if we drop the `new` so we only have `for later .. builds`?

articles/production/production-build.adoc
caalador: added build as clarification of action

articles/production/production-build.adoc
Mikaelsu: I think it'd be clearer if you drop the `new`.

Mikaelsu: How about something like:
```
[NOTE]
A pre-compiled production bundle has debugging features enabled, which means it's not the same as a pre-compiled front-end bundle for development mode.
		

[NOTE]
Output from an optimized production bundle is not used as a pre-compiled bundle for later production builds.
```
?

mshabarov:
@caalador could you please resolve the conflicts?

PR #2465, opened 2023-06-01
Fix Azure registration-id in SSO tutorial

heruan:
Looks like Spring doesn't like dotted registration-ids for OpenID providers, let's use one without it to be safe.

### DIFF
--- a/articles/tools/sso/integrations/azure-ad.asciidoc
+++ b/articles/tools/sso/integrations/azure-ad.asciidoc
@@ -59,7 +59,7 @@ You can use secret identifications to establish a secure connection with your Va
 .`application.properties`
 [source,properties]
 ----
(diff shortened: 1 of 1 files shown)


Mikaelsu (approved):
This is the second oddball in this tutorial, thanks for the update! @heruan, my Azure license has expired, could you go through the tutorial once more for sanity's sake?

heruan:
> This is the second oddball in this tutorial, thanks for the update!

No worries! This also slipped the review process 😅 

> @heruan, my Azure license has expired, could you go through the tutorial once more for sanity's sake?

Mine too, I'm re-activating it now and I'll verify again all the steps!

Mikaelsu:
> Mine too, I'm re-activating it now and I'll verify again all the steps!

Thanks! I'll go through the other two just to double-check.

PR #2477, opened 2023-06-07
Recommend line-awesome add-on for V24

mshabarov:
Fixes https://github.com/vaadin/flow/issues/16899

### DIFF
--- a/articles/configuration/development-mode/index.asciidoc
+++ b/articles/configuration/development-mode/index.asciidoc
@@ -62,7 +62,7 @@ This is the default mode.
 
 == Front-End Development Server for Easier Debugging
 
-When working with a front-end heavy project, there can be a front-end development server running for fast compilation and live-reload when the front-end files are changed.
+When working with a frontend heavy project, there can be a front-end development server running for fast compilation and live-reload when the front-end files are changed.
 
 Vaadin uses Vite to quicken front-end development by enabling the client-side live reload. By doing this, JavaScript/TypeScript changes are updated immediately in the browser. Vite only re-compiles what has changed, making hot reloads fast.
 
--- a/articles/configuration/development-mode/index.asciidoc
+++ b/articles/configuration/development-mode/index.asciidoc
@@ -62,7 +62,7 @@ This is the default mode.
 
 == Front-End Development Server for Easier Debugging
 
-When working with a frontend heavy project, there can be a front end development server running for fast compilation and live-reload when the front-end files are changed.
+When working with a front-end heavy project, there can be a front-end development server running for fast compilation and live-reload when the front-end files are changed.
 
 Vaadin uses Vite to quicken front-end development by enabling the client-side live reload. By doing this, JavaScript/TypeScript changes are updated immediately in the browser. Vite only re-compiles what has changed, making hot reloads fast.
 
--- a/articles/configuration/development-mode/index.asciidoc
+++ b/articles/configuration/development-mode/index.asciidoc
@@ -4,6 +4,7 @@ description: Vaadin Development Mode makes application development faster by ena
 order: 20
 ---
 
+
 = Development Mode
 
 Vaadin applications can run in two modes: development mode; or production mode.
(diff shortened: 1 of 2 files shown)


articles/configuration/development-mode/index.asciidoc
```diff
@@ -47,20 +47,21 @@ Vaadin generates an application bundle in the following situations:
 - Custom theme packaged as a `JAR` dependency is added, if it defines any assets to be added to the project; or
 - Exported web component is added.
 
-Custom theme files are served separately from the front-end bundle. They can be modified on the fly; the browser live-reload triggers a refresh immediately -- no re-compilation is needed.
+Custom theme files are served separately from the front-end bundle. They can be modified on the fly; the browser live-reload will trigger refresh immediately -- no re-compilation is needed.
```
caalador, line 50: I would perhaps keep this as it was or change the wording all together as
```suggestion
Custom theme files are served separately from the front-end bundle. They can be modified on the fly; the browser live-reload instantly triggers a refresh of the view -- no re-compilation is needed.
```

articles/configuration/development-mode/index.asciidoc
```diff
@@ -47,20 +47,21 @@ Vaadin generates an application bundle in the following situations:
...
 
 When customizing an <<{articles}/styling/application-theme#,Application Theme>> in this mode, the following points should be taken into account:
 
+- Loading CSS into the shadow DOM of Vaadin components by placing them in the theme's `components` sub-folder is not currently supported.
```
caalador, line 54: Should we add here `without compilation/generation/bundling/()what ever it is called)`

articles/configuration/development-mode/index.asciidoc
```diff
@@ -47,20 +47,21 @@ Vaadin generates an application bundle in the following situations:
...
-- Loading CSS into the shadow DOM of Vaadin components by placing them in the theme's `components` sub-folder is not currently supported.
+- Use Vaadin https://vaadin.com/directory/component/line-awesome-icons-for-vaadin[Line-Awesome] add-on that doesn't require building a new bundle.
 
 This is the default mode.
 
 
 == Front-End Development Server for Easier Debugging
 
-When working with a frontend-heavy project, there can be a front-end development server running for fast compilation and live-reload when the front-end files are changed.
+When working with a front-end-heavy project, there can be a front-end development server running for fast compilation and live-reload when the front-end files are changed.
```
caalador, line 64:
```suggestion
When working with a frontend-heavy project, there can be a front-end development server running for fast compilation and live-reload when the front-end files are changed.
```
as vale suggests

articles/upgrading/index.adoc
```diff
@@ -449,6 +449,12 @@ Vaadin 24 and Spring-boot 3 use SLF4J library version 2.0, which has breaking ch
 // Allow CheckboxGroup
 pass:[<!-- vale Vaadin.ProductName = NO -->]
 
+== Line-Awesome Icons Set Library
+
+Vaadin 23 starter projects used https://icons8.com/line-awesome[`line-awesome`] icons set library.
+You may need `line-awesome` library in your project as well.
```
caalador, line 455:
```suggestion
An older project based on a starter may need the `line-awesome` library as well.
```

articles/upgrading/index.adoc
```diff
@@ -449,6 +449,12 @@ Vaadin 24 and Spring-boot 3 use SLF4J library version 2.0, which has breaking ch
...
+If so, we recommend to use Vaadin https://vaadin.com/directory/component/line-awesome-icons-for-vaadin[Line-Awesome] add-on with Vaadin 24.
```
caalador, line 456:
```suggestion
In these cases it is recommend to use the Vaadin https://vaadin.com/directory/component/line-awesome-icons-for-vaadin[Line-Awesome] add-on with Vaadin 24.
```

articles/upgrading/index.adoc
```diff
@@ -449,6 +449,12 @@ Vaadin 24 and Spring-boot 3 use SLF4J library version 2.0, which has breaking ch
...
+In Vaadin 24 this add-on doesn't require a new bundle building, so your app would benefit from pre-compiled bundle feature.
```
caalador, line 457:
```suggestion
In Vaadin 24 this add-on doesn't require building a new bundle, so your application would then benefit from pre-compiled bundle feature.
```

articles/configuration/development-mode/index.asciidoc
mshabarov: Yeah, we need to clarify this. Just "is not supported" sounds like it cannot be used at all, but in fact it can be used with hotdeploy mode enabled or if one forces dev-bundle to be re-build anyhow.

mshabarov: Added a clarification about starting a dev-server in this case. Recommending the rebuild doesn't make sense IMO.

caalador (approved):

PR #2481, opened 2023-06-08
Docs/component thumbnails fix

rolfsmeds:
Updated component thumbnails with non-rounded versions, and added TreeGrid to index page.


Mikaelsu (approved):
Tested that all images show and the new cards link to the corresponding docs. Cool!

Mikaelsu (approved):
Tested that all images show and the new cards link to the corresponding docs. Cool!

PR #2498, opened 2023-06-14
PR2457-Observability Rel 2.1 Hilla

russelljtdyer:
PR 2457 was merged and closed before it could be checked for language. This pull request contains edits of the files touched by that previous PR.

### DIFF
--- a/articles/lit/acceleration-kits/observability-kit/configuration.adoc
+++ b/articles/lit/acceleration-kits/observability-kit/configuration.adoc
@@ -19,7 +19,7 @@ The `service` attributes are used to differentiate traces from services or appli
 
 The service attributes are as follows:
 
-- `service.name` is an attribute that's used to distinguish a service by name.  It's the only attribute with a default value, which is `vaadin`.
+- `service.name` is an attribute that's used to distinguish a service by name. It's the only attribute with a default value, which is `vaadin`.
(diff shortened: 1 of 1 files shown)


articles/lit/acceleration-kits/observability-kit/configuration.adoc
```diff
@@ -18,10 +19,9 @@ The `service` attributes are used to differentiate traces from services or appli
 
 The service attributes are as follows:
 
-- `service.name` is an attribute that's used to distinguish a service by name. The default value is `vaadin` -- the only attribute with a default value.
-It's also the only one required.
-- `service.namespace` helps to distinguish a group of services.
-- `service.instance.id` helps to distinguish instances of the same service that exist simultaneously. It must be unique for each instance of the same `service.namespace`, `service.name` pair.
+- `service.name` is an attribute that's used to distinguish a service by name. The default value is `vaadin`. It's the only attribute with a default value and the only one required.
```
MatthewVaadin, line 22: Considering that the attribute has a default value, perhaps we can drop the required bit, like this:
```suggestion
- `service.name` is an attribute that's used to distinguish a service by name. It's the only attribute with a default value, which is `vaadin`.
```

MatthewVaadin (changes requested):

PR #2500, opened 2023-06-14
Fix link

tarekoraby:


jouni:
I prefer keeping this rule as an error. But the link was indeed incorrectly processed, because of this issue in Design System Publisher: https://github.com/vaadin/docs-app/issues/383

jouni:
I’ll fix that issue and then revert this change.

PR #2501, opened 2023-06-14
docs: Visual View Builder updates after going live

MarcinVaadin:
This PR is aligning docs with name chosen for drag and drop builder - _Visual View Builder_. Also updates screenshot of share dialog.


Mikaelsu (approved):
Thank you, @MarcinVaadin!

PR #2517, opened 2023-06-21
feat: add grid react examples

tomivirkki:
Add a few more React examples

PR #2521, opened 2023-06-22
wip: copy components examples to hilla

tomivirkki:
# Do not merge this PR!

Once all the live examples currently running on vaadin.com components docs have [React counterparts](https://github.com/vaadin/docs/pull/2517) available, the intention is to copy all example related resources from `latest` to `hilla` so that the live React examples would be available on hilla.dev (`hilla` branch is already configured to hide Java/TypeScipt examples in production).

Copying the example resources will be automated and run periodically (similarly to how the current `latest` -> `dsp` copy automation works).

This PR is an example of what the pull request generated by the bot could eventually look like.


jouni (commented):
Now that we reverted the `hiddenContent` feature from DSP, and rely on regular AsciiDoc attributes instead, this PR needs to be updated. See https://github.com/vaadin/docs/pull/2549 for more info.

PR #2526, opened 2023-06-26
Fix wrong class

Mikaelsu:


Mikaelsu:
Removed the language check label as there is no language to check

DiegoCardoso (approved):

PR #2537, opened 2023-06-29
Tab hover only as CSS pseudo-class

JkbLskw:
Hovering of a vaadin-tab is only possible via CSS pseudo-class e.g. `vaadin-tab:hover { ... }` and not via CSS pseudo-element `vaadin-tab::hover { ... }`.

Therefore, the documentation should be adapted.


jouni (approved):
Thanks for noticing this error and fixing it!

PR #2538, opened 2023-06-29
Add `meta-description`

Mikaelsu:
And other updates prompted by our SEO agency

### DIFF
--- a/articles/components/combo-box/index.asciidoc
+++ b/articles/components/combo-box/index.asciidoc
@@ -2,6 +2,7 @@
 tab-title: Usage
 layout: tabbed-page
 title: Combo Box
+meta-description: The Vaadin ComboxBox component is accessible, themable, and enterprise-grade. Made for Java and Typescript, use it with Flow or a web frontend of your choice.
 page-links:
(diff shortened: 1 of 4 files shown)


Mikaelsu:
Removed the language check to keep keywords

jouni:
There's no support for a `meta-description` frontmatter attribute. Only `description`, which does end up in the DOM as meta tag, but also as a visible page item under the page heading.

Mikaelsu:
I'll rework this PR next month.

articles/components/grid/index.asciidoc
```diff
@@ -2,6 +2,7 @@
 title: Grid
 tab-title: Usage
 layout: tabbed-page
+meta-description: The Vaadin Grid component is accessible, themable, and enterprise-grade.
```
jouni, line 5: This is a bit brief and not very informative. I would expect the words "data grid" or "tabular data" to appear somewhere.

articles/components/combo-box/index.asciidoc
```diff
@@ -2,6 +2,7 @@
 tab-title: Usage
 layout: tabbed-page
 title: Combo Box
+meta-description: The Vaadin ComboxBox component is accessible, themable, and enterprise-grade. Made for Java and Typescript, use it with Flow or a web frontend of your choice.
```
jouni, line 5: This doesn't say what the component does. But perhaps "Combo Box" is well known, and we don't need to be patronizing?

jouni (commented):
@Mikaelsu, are you still working on this?

PR #2544, opened 2023-07-03
docs: cover validation changes in upgrading guide

vursen:
The PR adds a section about field validation changes to the upgrading guide.

### DIFF
--- a/articles/upgrading/index.adoc
+++ b/articles/upgrading/index.adoc
@@ -495,7 +495,6 @@ h5 { margin-bottom: 0.25em; }
 
 pass:[<!-- vale Vaadin.ProductName = YES -->]
 
-
 === Removed Flow Component APIs
 
 .View Removed or Renamed Flow APIs.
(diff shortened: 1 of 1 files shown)


articles/upgrading/index.adoc
```diff
@@ -583,6 +583,78 @@ The following changes only affect the client-side APIs of Vaadin components:
...
+
+**Before:**
+
+Both constraint and Binder validation were only triggered on `ValueChangeEvent`, while client-side validation was completely disregarded.
+
+**After:**
+
+The server-side validation timing is aligned with client-side validation, meaning that whenever the web component validation occurs, it triggers the corresponding validation in the Flow component.
+
+In practice, it means that the server-side validation will also be performed on blur. This resolves issues of required fields not getting invalidated after the initial user interaction.
```
knoobie, line 619: > This resolves issues of required fields not getting invalidated after the initial user interaction.

I would personally argue that this is a bug in itself and should not be sold as feature 😉 

Jumping in a required field (not typing anything) and pressing tab to go to the next field, should not mark the field as invalid in my opinion.

vursen: Thanks, I removed this part to avoid potential controversies.

articles/upgrading/index.adoc
```diff
@@ -583,6 +583,76 @@ The following changes only affect the client-side APIs of Vaadin components:
...
+
+Date Picker, Integer Field, and some other similar components only accept user input that can be parsed as a `LocalDate`, `Integer`, etc. Otherwise, the value on the server falls back to `null`.
+
+**Before:**
+
+Entering bad input didn't invalidate the component because the server treated this case as an empty value.
+
+**After:**
+
+The server detects if the user has entered any input. When the input cannot be parsed, it's considered bad input, causing constraint validation to fail once the user presses Enter or removes focus from the component.
```
vursen, line 633:
```suggestion
The server is aware if the user has entered any input. When the input cannot be parsed, it's considered bad input, causing constraint validation to fail once the user presses Enter or removes focus from the component.
```

articles/upgrading/index.adoc
```diff
@@ -583,6 +583,76 @@ The following changes only affect the client-side APIs of Vaadin components:
...
+**After:**
+
+Component constraints are integrated into Binder validation. Binder checks the value against constraints before custom validators.
+
+The only exception is the required constraint. It doesn't participate in Binder validation since Binder provides its own `asRequired` validator.
+
+**Known issues:**
+
+- It's currently not possible to configure custom error messages for constraints (https://github.com/vaadin/flow-components/issues/4618[flow-components#4618]).
+- It's currently not possible to revert Binder's behavior to ignore constraints as before, except by removing constraints from the component at all.
```
vursen, line 607:
```suggestion
- It's currently not possible to revert Binder's behavior to ignore constraints as before, except by removing constraints from the component at all (https://github.com/vaadin/flow/issues/17178[flow#17178]).
```

rolfsmeds (approved):
LGTM, only thing we might want to add is a very brief explanation of why validate-on-blur is the new behavior.

articles/upgrading/index.adoc
```diff
@@ -583,6 +583,76 @@ The following changes only affect the client-side APIs of Vaadin components:
...
+
+=== Blur Triggers Server-Side Validation
+
+**Before:**
+
+Both constraint and Binder validation were only triggered on `ValueChangeEvent`, while client-side validation was completely disregarded.
+
+**After:**
+
+The server-side validation timing is aligned with client-side validation, meaning that whenever the web component validation occurs, it triggers the corresponding validation in the Flow component. In practice, it means that the server-side validation will also be performed on blur (i.e., when the component loses focus).
```
vursen, line 617:
```suggestion
The server-side validation timing is aligned with client-side validation, meaning that whenever the web component validation occurs, it triggers the corresponding validation in the Flow component. In practice, it means that the server-side validation will also be performed on blur (the web component runs validation on blur to identify possible bad input). 

```

rolfsmeds: yeah, I think that's good enough as an explanation.

articles/upgrading/index.adoc
vursen: I added https://github.com/vaadin/web-components/issues/6146 to the list of known issues. Closing the thread.

PR #2549, opened 2023-07-05
chore: bump DSP to 2.2.0-alpha.2

sissbruecker:
- Bumps DSP to 2.2.0-alpha.2
- Updates configuration to remove the `hiddenContent` setting that was removed, and adds attributes for frameworks that should be enabled
- Wrap React examples into conditional so that they are only shown when `React` attribute is defined
- Update conditionals for Flow / Java-only content

For now this only adds / updates conditionals that are required to properly display content for the Vaadin docs site. I'll open follow-up PRs to:
- Introduce a `lit` attribute, and wrap all examples with a respective conditional (`flow`, `lit`, `react`)
- Do the same version bump and configuration update in the `hilla` branch

### DIFF
--- a/dspublisher/config/default.json
+++ b/dspublisher/config/default.json
@@ -1,6 +1,6 @@
 {
   "asciidocAttributes": {
-    "Flow": true,
-    "React": true
+    "flow": true,
+    "react": true
   }
 }
--- a/dspublisher/config/default.json
+++ b/dspublisher/config/default.json
@@ -1,6 +1,6 @@
 {
   "asciidocAttributes": {
     "Flow": true,
-    "React": false
+    "React": true
   }
 }
--- a/dspublisher/config/production.json
+++ b/dspublisher/config/production.json
@@ -1,6 +1,6 @@
 {
   "asciidocAttributes": {
-    "Flow": true,
-    "React": false
+    "flow": true,
+    "react": false
   }
 }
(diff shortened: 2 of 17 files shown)


dspublisher/config/default.json
```diff
@@ -0,0 +1,6 @@
+{
+  "asciidocAttributes": {
+    "Flow": true,
```
sissbruecker, line 3: Opinionated choice to use `Flow` as attribute rather than `Java`, even though that misaligns it with the actual group names in the examples. Same would apply to the attribute for Lit-specific content, which I would rather name `Lit` than `Typescript`, since React content will also use Typescript.

Don't have a strong opinion on this though, we can change it if we consider it too confusing.

dspublisher/config/production.json
```diff
@@ -1,5 +1,6 @@
 {
-  "hiddenContent": [
-    { "path": "/components", "categories": ["React"] }
-  ]
+  "asciidocAttributes": {
+    "Flow": true,
+    "React": false
```
sissbruecker, line 4: Interestingly, defining an attribute as `false` will hide content wrapped in `ifdef`, even though the attribute is technically defined in the config. I like it, allows us to be more explicit in the config, rather than having to remove attributes to get hiding to work.

dspublisher/config/default.json
jouni: That said, we currently use lowercase `lit` and `react` attribute names in the Hilla branch.

dspublisher/config/production.json
jouni: Oh, interesting. I also would’ve thought Asciidoc would see the attribute as defined as `false`. But yeah, it’s nicer this way, work as a developer (who is not familiar with Asciidoctor) would expect.

dspublisher/config/default.json
sissbruecker: Right, other attributes in DSP and docs are lower-case as well. I'll update it.

sissbruecker: Done

jouni (approved):

PR #2564, opened 2023-07-12
chore: wrap component examples into conditionals

sissbruecker:
Wraps all component examples into conditionals, so that they are only visible when the respective `flow`, `lit` or `react` attributes are enabled.


dspublisher/config/default.json
```diff
@@ -1,6 +1,7 @@
 {
   "asciidocAttributes": {
     "flow": true,
-    "react": true
+    "lit": true,
+    "react": false
```
tomivirkki, line 5: @sissbruecker Was changing "react" to `false` intentional here? It's `false` in `production.json` as expected but we want the React examples visible during development.

PR #2569, opened 2023-07-13
Mention Microsoft Entra ID

tarekoraby:


jouni:
Need to add "Entra ID" to the vocabulary (accept.txt).

Should we now prefer Entra ID over Azure AD, i.e. `Entra ID (previously known as Azure AD)`, instead of how you did it here?

Mikaelsu (approved):
LGTM

PR #2572, opened 2023-07-17
chore: update quarkus know-issues paragraph

mcollovati:
References vaadin/quarkus#136


articles/integrations/quarkus.asciidoc
```diff
@@ -176,17 +176,19 @@ Furthermore, Java hot reload may sometimes break front-end live reload; if this
 [[quarkus.vaadin.knownissues]]
 == Known Issues
 
-Quarkus Bill-of-Materials (BOM) pins Google Guava library to a version that conflicts with Vaadin TestBench 8 and later, resulting in test failures because of changes in method signatures.
-This can be fixed by adding an explicit entry for Guava version `31.0.1-jre` in the dependency management section of the project's [filename]`pom.xml` file, immediately above the reference to Quarkus BOM.
+Quarkus Bill-of-Materials (BOM) may pin libraries to version that conflicts with Vaadin, resulting in runtime or test failures because of changes in method signatures.
```
manolo, line 179: I think we should be more explicit and say that this runtime failures are in dev-mode, and probably also paste here the meaningful part of the exception, so as copy-pasting it in google ends showing this page, wdyt?

mcollovati: Sounds like a good idea

mcollovati: updated the paragraph

manolo (approved):

Issue #2411, opened 2023-05-09
Clean up the `dsp` branch


jouni:
The `dsp` branch has several issues:

- The "Foundation" section has duplicate pages with almost identical content:
  ![Screenshot 2023-05-09 at 9 41 53](https://user-images.githubusercontent.com/66382/237020648-12946bcb-1960-47b0-8f28-f2559a01696b.png)
- The `articles/ds/` folder is unnecessary, now that the script that picks updates from the `latest` branch can move pages. We should remove the extra `ds/` folder.
- The `articles/styling/` folder is not shown on the website at all, since it’s outside the `ARTICLES_PATH` folder.
- The `scripts/dspublisher-init.js` file should be removed after the project has been initialized.

jouni:
Related, but not as important perhaps: the styling of the cards on the main index page is non-existent, making it hard for the viewer to make sense of the content structure.
![Screenshot 2023-05-09 at 10 18 34](https://user-images.githubusercontent.com/66382/237023044-10e336fa-db50-4d45-9940-18e36132ed71.png)

PR #2580, opened 2023-07-20
Clean up the `dsp` branch

jouni:
Refactor the `dsp` branch so that it’s in a usable state for new customers. 

Related to #2411

This probably causes issues for any existing customers wanting to pull updates from this branch to their own documentation versions. But I don't see any other option – the current state of the branch is a mess, and a bad starting point for any new system.

All these changes were done manually while running the development server locally, to verify the changes. 

I did not update the cherry-pick script. I think that can be handled separately, and we just need to get this branch cleaned up soon so that new customers don't continue getting a poor initial experience.

I ended up deleting most of the "Styling" section. That change can be discussed, whether or not it should still be part of the documentation of a custom design system. The remaining parts are quite much on the same level, and I don't have any clear guidelines at this point to help with the decision.

### Note

We release DSP v2.2 as soon as possible as well, so that this PR can be updated to use that (instead of the 2.2.0-alpha.2 release).


articles/components/button/index.asciidoc
```diff
@@ -428,6 +487,3 @@ include::{root}/src/main/java/com/vaadin/demo/component/button/ButtonGrid.java[r
 |<<../menu-bar#,Menu Bar>>
 |Overlay menus with items that trigger actions. This can also be used for single "menu buttons" and "button groups" without overlay menus.
 |===
-
-
-[discussion-id]`8E1BE28B-D5F0-490C-A8FA-82975D9A3B43`
```
tomivirkki, line 433: We can either wrap these (in the source asciidoc) inside `ifdef::discussionids[]` or make the sync script [remove](https://github.com/vaadin/docs/blob/76f3ad5f4a10176bf8e0dc74b9530f3b6a0da05d/scripts/pick/cherry-pick-config.js#L110-L111) discussion ids.

tomivirkki (approved):
lgtm

articles/components/button/index.asciidoc
jouni: True. Would be great to remove them with the sync script. Same for the Vale rule overrides, i.e. `pass:[<!-- vale ... -->]`.

jouni:
I’ll double check the results by setting up a new DSP installation from this branch.

jouni:
Works fine on macOS, the result is as expected.

Startup still fails on Windows, but it seems like it’s not related to the project contents directly.

PR #2595, opened 2023-07-26
feat: clone live examples foundation to hilla branch

tomivirkki:
# Description

This PR clones the necessary files to run live examples from "latest" to "hilla" branch. These files include shared data providers, Java classes for beans used across the demos (such as `Person`), demo data and images etc.

The PR has been generated with the [wip cherry-pick config](https://github.com/vaadin/docs/compare/feat/hilla-react-cherry-pick-config?expand=1) intended for synchronizing everything related to component documentation (including live React examples) from `latest` to `hilla` branch. Changes related to specific component documentation/examples have been discarded from this PR though as they're to be added separately.

Once this PR is merged it's feasible to replace individual component pages on the Hilla docs with ones generated using the cherry-pick script.

The branch of [the experimental cherry-pick PR](https://github.com/vaadin/docs/pull/2521) has been rebased on top of this one to showcase the generated component pages. The branch can be used to validate the changes of this PR.


dspublisher/config/default.json
```diff
@@ -1,3 +1,7 @@
 {
-  "hiddenContent": [{ "path": "/react/components", "categories": ["Java", "TypeScript"] }]
+  "asciidocAttributes": {
+    "flow": false,
+    "lit": false,
```
jouni, line 4: Is this intentionally `false`?

dspublisher/dspublisher-scripts.js
```diff
@@ -5,7 +5,7 @@ const fs = require('fs');
 const path = require('path');
 const http = require('http');
 
-const DSP_VERSION = '2.2.0-alpha.1';
+const DSP_VERSION = '2.2.0-alpha.2';
```
jouni, line 8:
```suggestion
const DSP_VERSION = '2.2.0-beta.1';
```

src/main/java/com/vaadin/dspublisher/licensecheck/Check.java
```diff
@@ -1,11 +1,13 @@
 package com.vaadin.dspublisher.licensecheck;
 
+import com.vaadin.pro.licensechecker.BuildType;
 import com.vaadin.pro.licensechecker.LicenseChecker;
 
 public class Check {
     public static void main(String[] args) {
         try {
-            LicenseChecker.checkLicense("vaadin-dspublisher", "2.1.1");
+            LicenseChecker.checkLicense("vaadin-dspublisher", "2.2.0-alpha.2",
```
jouni, line 9:
```suggestion
            LicenseChecker.checkLicense("vaadin-dspublisher", "2.2.0-beta.1",
```

jouni (approved):

dspublisher/config/default.json
tomivirkki: Yeah, we don't want to show `Lit` examples in https://hilla.dev/docs/react/components

jouni: But we want to show Lit examples in hilla.dev/docs/lit/components, right?

sissbruecker (approved):

dspublisher/config/default.json
sissbruecker: I was also assuming there would be shared component pages for both Lit / React where you could switch the group as in the Vaadin docs.

tomivirkki: Looks like `:lit:` and `:react:` attributes have been used quite a lot across the Hilla documentation. Now that we'd enable `react` attribute globally, it would make React content visible also on the Lit side:

Screenshot 2023-07-28 at 11 43 04

Suggestions?

jouni: I assume the component pages will be shown in both Lit and React sections/versions of the documentation. At the moment, any content that should be visible in both is maintained in the Lit version, and included using `include::[]` to the React version, and those pages define the `:react:` AsciiDoc attribute before the include. Now they should also define `:lit!:` to unset that attribute, and not show the Lit examples.

jouni: So to me, the default config should be `"flow": false, "lit": true, "react": false`, and the React pages should then define `:react:` and `:!lit:` to reverse that.

tomivirkki: I guess that would be the same as not adding the config at all and just generating `:react:` on each component page under hilla/react?

tomivirkki: Another option would be to add support for defining the path under which the attributes are effective... Something like:

```
  "asciidocAttributes": {
    "flow": false,
    "lit": "articles/lit",
    "react: "articles/react"
```

tarekoraby: > But we want to show Lit examples in hilla.dev/docs/lit/components, right?

No, that wasn't the plan we discussed. We can, of course, reconsider if we think adding Lit examples is feasible with a low effort.

tomivirkki: I'll remove the config from this PR anyways to avoid these issues for now

PR #2603, opened 2023-07-30
docs: add react examples, button focus, images and labels

ahsinsiddique:


frontend/demo/component/button/react/button-labels.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <>
+      {/* tag::snippet[] */}
+      <VerticalLayout>
+        <HorizontalLayout theme="spacing">
+          <EmailField
+            label="Primary email address"
+            value={primaryEmail}
+            onValueChanged={(event: any) => setPrimaryEmail(event.target.value)}
+          />
+          <Button className={'mt-xl'} onClick={() => setPrimaryEmail('')}>Remove</Button>
```
tomivirkki, line 22: Use `align-items: baseline` CSS for aligning the components instead of adding an arbitrary margin to the buttons, the same way as in the TS example

frontend/demo/component/button/react/button-labels.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <>
+      {/* tag::snippet[] */}
+      <VerticalLayout>
+        <HorizontalLayout theme="spacing" style={{ alignItems: 'baseline' }}>
+          <EmailField
+            label="Primary email address"
+            value={primaryEmail}
+            onValueChanged={(event: any) => setPrimaryEmail(event.target.value)}
```
tomivirkki, line 20: [Avoid](https://thoughtbot.com/blog/typescript-stop-using-any-there-s-a-type-for-that) using `any` as a type. Take a look at how the events are handled in the Lit example and use the same approach.

ahsinsiddique: yes it can be omitted but I can see that in typescript `EmailFieldValueChangedEvent` type is added for event

ahsinsiddique: simple can work as well, should I remove this type for React?

tomivirkki: Yes please, let's not add anything that can be omitted.

ahsinsiddique: removed event type.

tomivirkki (approved):

PR #2605, opened 2023-08-01
fix imports in tutorial

manolo:


articles/lit/start/in-depth/navigation-and-code-splitting.adoc
```diff
@@ -130,7 +130,7 @@ Update [filename]`routes.ts` with the following content:
 ----
 import { Route } from '@vaadin/router';
 import './views/list/list-view';
-import './main-layout.ts';
+import './main-layout';
```
cromoteca, line 133:
```suggestion
import './main-layout.js';
```

cromoteca: This is the recommended import style: using `.js` extension.

cromoteca (changes requested):

cromoteca (approved):

articles/lit/start/in-depth/navigation-and-code-splitting.adoc
manolo: having `.js` and no-extension in the same file is confusing for the user, I would update all extensions to in this project and others to `.js` at once in separated PRs

PR #2610, opened 2023-08-02
Clarify that it's the O11y example that is not production-ready

tarekoraby:
Clarify that it's the example that is not production-ready, not the observability Kit itself.


Mikaelsu (approved):
LGTM

PR #2611, opened 2023-08-02
Incremental builds for Gradle vaadinPrepareFrontend task

mshabarov:
Related-to https://github.com/vaadin/flow/issues/15342


articles/guide/start/gradle.asciidoc
```diff
@@ -227,6 +227,33 @@ To get the complete list of tasks handled by the configured plugins, execute the
...
+   * configuration parameters, see <<all-options,plugin configuration options>>
+
+- Outputs:
+   * `package.json`, `package-lock.json`, `vite.config.js` and other front-end files generated by Vaadin.
+
+Incremental build feature can be turned off with the following configuration parameter in the [filename]`gradle.build` file:
+
+----
+vaadin {
+   alwaysExecutePrepareFrontend = true
```
tarekoraby, line 251: Should this be `false` in case one wants to disable the feature?

mshabarov: No, because this flag is for always executing the task. When it true, Gradle doesn't check whether the defined inputs and outputs are changed, but run it always, so it effectively disables the feature.

tarekoraby: Ah ok

articles/guide/start/gradle.asciidoc
```diff
@@ -227,6 +227,33 @@ To get the complete list of tasks handled by the configured plugins, execute the
 ./gradlew tasks
 ----
 
+[since:com.vaadin:vaadin@V24.2]
+[[incremental.builds]]
+=== Incremental Builds
+
+Vaadin uses https://docs.gradle.org/current/userguide/incremental_build.html[Gradle Incremental Builds feature] for `vaadinPrepareFrontend` task to avoid `Node.js` installation and front-end files generation that has already been done and not changed since the previous build.
```
mcollovati, line 234: > to avoid `Node.js` installation

This seems a bit unclear to me: I read it as with this feature, nodejs is not installed at all.

Actually, even without the incremental build feature, node/npm/pnpm are installed only if required and not present or if the version of the global installed tool is unsupported.

mshabarov: Yes, you're right, this sounds incorrect. The proper explanation should be that the task can be skipped if project  configuration including desired Node version and resulted front-end files are not changed. Whether to install or not the Node isn't done with this feature.

mshabarov: Updated

mcollovati (approved):

Mikaelsu (approved):

PR #2620, opened 2023-08-04
Batch of Hilla updates

Mikaelsu:


Mikaelsu:
@jouni I marked this ready for review. The rest of the updates warrant their own PRs.

articles/lit/guides/client-caching.adoc
```diff
@@ -9,7 +9,7 @@ order: 75
 
 = Client-Side Data Caching
 
-Hilla caches client-side views out of the box, but not client-side data.
+Hilla caches client-side views out-of-the-box, but not client-side data.
```
jouni, line 12: I think this change is not correct. The term is not used as an adjective, so it shouldn't be hyphenated.

articles/lit/guides/endpoints.adoc
```diff
@@ -193,7 +193,7 @@ endif::[]
 
 Hilla automatically generates endpoint URLs and wraps them in the generated TypeScript API so the developer doesn't have to worry about them.
 
-Any public method in any Hilla endpoint can be accessed with the `\http://${base_url}/${prefix}/${endpoint_name}/${method_name}` URL format.
+Any public method in any Hilla endpoint can be accessed with the `\http://${base_url}/${prefix}/${endpoint_name}/${method_name}` URL format. Use TypeScript methods instead of the URL when calling it, i.e., `myurl = new URL("http://");`.
```
jouni, line 196: I think it would it be clearer to say “Don't use the URL directly to access the endpoint. Use the TypeScript methods instead.” Now it, with the `i.e., myurl = new URL()` example, it might look that’s what you should do, with a quick glance.

articles/lit/guides/forms/binder.asciidoc
```diff
@@ -6,9 +6,9 @@ order: 1
 
 = Binding Data to Forms
 
-Hilla provides a means of binding UI components in TypeScript form views.
+Hilla provides a way for binding input fields to a data model.
 
-The client-side [classname]`Binder` supports Java back-end endpoints for loading and saving the form data, and reuses the metadata from Java Bean validation annotations for client-side validation.
+The client-side [classname]`Binder` supports Java backend endpoints for loading and saving the form data, and reuses the metadata from Java Bean validation annotations for client-side validation.
```
jouni, line 11: I think we might want to chance this soon (and always use the non-hyphenated versions of "frontend" and "backend" everywhere), but for now, "back-end" is used here as an adjective, and should be hyphenated.

articles/lit/guides/client-caching.adoc
Mikaelsu: You're right!

articles/lit/guides/endpoints.adoc
Mikaelsu: How about this?

Mikaelsu: Thanks!

PR #2625, opened 2023-08-06
docs: add react button examples, button dialog, tertiary, and form, grid

ahsinsiddique:


frontend/demo/component/button/react/button-dialog.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <>
+      {/* tag::snippet[] */}
+      <VerticalLayout theme="spacing">
+        <FormLayout responsiveSteps={[{ columns: 2 }]}>
+          <TextField label="First name" value="John" />
+          <TextField label="Last name" value="Smith" />
+          <EmailField {...{ colspan: 2 }} label="Email address" value="john.smith@example.com" />
+        </FormLayout>
+
+        <HorizontalLayout theme="spacing" style={{ width: '100%', justifyContent: 'flex-end' }}>
```
tomivirkki, line 21: Why not use the same styles as the lit example?

frontend/demo/component/button/react/button-dialog.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { VerticalLayout } from '@hilla/react-components/VerticalLayout.js';
+import { EmailField } from '@hilla/react-components/EmailField.js';
+import { TextField } from '@hilla/react-components/TextField.js';
+import { HorizontalLayout } from '@hilla/react-components/HorizontalLayout.js';
+import { FormLayout } from '@hilla/react-components/FormLayout.js';
+
+function Example() {
+  return (
+    <>
+      {/* tag::snippet[] */}
```
tomivirkki, line 13: Turns out, if there's only 1 root component, like in this case, you can omit the fragment and use:

```tsx
return (
  // tag::snippet[]
  <Component>
    ....
  </Component>
  // end::snippet[]
)
```

frontend/demo/component/button/react/button-dialog.tsx
ahsinsiddique: it's the same style as lit example, I converted to react example

tomivirkki: This [layout in the Lit example](https://github.com/vaadin/docs/blob/491534bb722352758de78dec3cb734d931bf3563/frontend/demo/component/button/button-dialog.ts#L38) has `flex-wrap: wrap; justify-content: flex-end;` whereas here it's `width: '100%', justifyContent: 'flex-end'`

ahsinsiddique: yes, I tried this approach, here without width 100% `HorizontalLayout ` not taking full width

tomivirkki: Then the styling is not the same as in the other examples. The child layout didn't take full width because you're missing [`align-items: stretch` from the flex container](https://github.com/vaadin/docs/blob/491534bb722352758de78dec3cb734d931bf3563/frontend/demo/component/button/button-dialog.ts#L25).

ahsinsiddique: ahan, sorry my bad, I missed, thank you. fixed

tomivirkki: I wonder if you forgot to push your changes.

tomivirkki (approved):

PR #2665, opened 2023-08-10
Update authentication.adoc

kirmerzlikin:
Fix endpoint method reference (`checkUser()` => `isAdmin()`)


platosha (approved):

platosha:
Good catch, thank you!

PR #2678, opened 2023-08-11
Clarifying font file location

Mikaelsu:


Mikaelsu:
@russelljtdyer this article could use some heavy-handed editing.

jouni:
@russelljtdyer , let’s do that separately.

PR #2691, opened 2023-08-11
docs: add React examples for cookie-consent

tomivirkki:
Extracted from #2632


frontend/demo/component/cookieconsent/react/cookie-consent-localization.tsx
```diff
@@ -0,0 +1,19 @@
+import { reactExample } from 'Frontend/demo/react-example'; // hidden-source-line
+import React from 'react';
+import './example-cleanup'; // hidden-source-line
+import { CookieConsent } from '@hilla/react-components/CookieConsent.js';
+
+function Example() {
+  return (
+    // tag::snippet[]
+    <CookieConsent
+      message="Tämä sivusto käyttää evästeitä parhaan kokemuksen tarjoamiseksi"
```
web-padawan, line 10: For some reason this example doesn't show correctly, even though `message` is set.

Screenshot 2023-08-14 at 14 36 56

web-padawan (commented):
Localization seems to be broken (see the comment on the corresponding example).

This could be a problem in the component itself as it's using [`connectedCallback()`](https://github.com/vaadin/web-components/blob/21b661b5c958e2592566c9d17ca8295697879702/packages/cookie-consent/src/vaadin-cookie-consent.js#L156) and does not support updating `message` and other properties.

tomivirkki:
Localication example is hidden in `hilla.dev`for now, see https://github.com/vaadin/docs/pull/2691/files#diff-33f6aa8140cdc64988e52965fd7c0e7d56738e970e60af38a7e916cd9d93fc79R51-R52

web-padawan (approved):

PR #2693, opened 2023-08-13
Expand description of type conversion for Instant and Date

kirmerzlikin:
I implemented a basic todo list app following [this tutorial](https://hilla.dev/docs/react/start/basics), extended `Todo` entity with two new fields - `createdAtDate` and `createdAtInstant`, and then tried sending non-timestamp values for those fields from the front-end client. It worked just fine.

Source code can be found [here](https://github.com/kirmerzlikin/hilla-todo).


jouni:
Thanks! I hope @platosha can check these updates soon.

platosha (approved):

PR #2733, opened 2023-08-27
add reference.adoc for react

taefi:


platosha (approved):

articles/lit/guides/forms/reference.adoc
```diff
@@ -234,10 +315,40 @@ console.log(emptyPerson); // {"fullName": ""}
...
+ifdef::lit[]
+which are handy for template expressions.
 
 For [classname]`StringModel` in string expressions, the following are equivalent:
+endif::[]
 
+ifdef::react[]
+[source,typescriptjsx]
+----
+const [model, value] = useBinder(PersonModel);
```
Lodin, line 332: The `useBinder` hook returns an object, not array
```suggestion
const { model, value } = useBinder(PersonModel);
```

articles/lit/guides/forms/reference.adoc
```diff
@@ -200,6 +260,26 @@ The item model is instantiated for every array entry, as necessary.
...
+  <>
+    {model.people.map(personBinder => {
+      <div>
+        <TextField label="Full name"
+                   {...field(personBinder.model.fullName)}>
+        </TextField>
+        <strong>Full name:</strong>
+        {personBinder.value.fullName}
+      </div>
+    })}
```
Lodin, line 277: I doubt this code gonna work; it is not react-ish as well. The model contains models, not binders / binder nodes, so it cannot be a `personBinder` here.

I believe it requires a separate component like that:
```ts

interface PersonProps {
  model: PersonModel;
}

function Person({ model }: PersonProps) {
  const {field, model, value} = useBinderNode(model);

  return (
    <div>
      <TextField label="Full name" {...field(model.fullName)} />
      <strong>Full name:</strong>
      {value.fullName}
    </div>
  ); 
}

// Dunno what the name of the component it should be
function Main() {
  return (
    <>
      {model.people.map(person => <Person model={person} />)}
    </>
  );
}
```

articles/lit/guides/forms/reference.adoc
```diff
@@ -234,10 +315,40 @@ console.log(emptyPerson); // {"fullName": ""}
...
+----
+
+Then, it is possible to use the values in formulas using either of the followings:
+
+[source,typescriptjsx]
+----
+return (
+  Cost: {value.quantity.valueOf() * value.price.valueOf()}
+  Cost: {value.quantity * value.price}
+);
```
Lodin, line 348: It won't compile. It should look like the following:
```suggestion
return (
  <>
    Cost: {value.quantity.valueOf() * value.price.valueOf()}
    Cost: {value.quantity * value.price}
  </>
);
```

articles/lit/guides/forms/reference.adoc
```diff
@@ -234,10 +315,40 @@ console.log(emptyPerson); // {"fullName": ""}
...
+ifdef::react[]
+[source,typescriptjsx]
+----
+const [model, value] = useBinder(PersonModel);
+...
+return (
+  {value.name.toString()}
+  {value.name.valueOf()}
+  {value.name}
+)
```
Lodin, line 338: It won't compile. It should look like the following:
```suggestion
return (
  <>
    {value.name.toString()}
    {value.name.valueOf()}
    {value.name}
  </>
)
```

Lodin (changes requested):
I suggest using the `demo` folder for all the React stuff to be sure it can be compiled and doesn't have any TS errors

articles/lit/guides/forms/reference.adoc
```diff
@@ -26,17 +48,50 @@ The [methodname]`field()` directive does the main job of binding the form field
...
 However, for HTML input elements, the `invalid`, `required` and `errorMessage` states aren't displayed in the bound component.
 As a workaround, you can bind these manually in the template:
 
+ifdef::react[]
+[source,typescriptjsx]
+----
+import { useBinder, useBinderNode } from '@hilla/react-form';
+...
+const { model, field } = useBinder(PersonModel);
+const fullNameBinder = useBinderNode(model.fullName);
```
Lodin, line 71: The `useBinder` and `useBinderNode` are not designed to be used in the same component. If you need to use it, then it's probably better to move all the specific code to a separate component.

Lodin (changes requested):

articles/lit/guides/forms/reference.adoc
taefi: Well, I must admit the approach you mentioned was the first thing I started with, as we should not promote any bad practices in our docs. But, then I came up with something very long for just demonstrating how to use the `useBinderNode` like this:
```tsx
interface PersonProps { 
    personModel: PersonModel; 
}

function FullNameComponent({ personModel }: PersonProps) {
  // @ts-ignore
  const { model, field, required, errors, invalid }= useBinderNode(model.fullname);

  return (
    <>
      <label htmlFor="fullName">
        Full name
        {required ? '*' : ''}
      </label>
      <input id="fullName" {...field(model)}></input>
      <br/>
      <span className="label" style={{visibility: invalid ? 'visible' : 'hidden'}}>
          <strong>
           {errors[0]?.message}
          </strong>
        </span>
    </>
  );
}

export default function Main() {
     const { model, submit, field, read, value, invalid } = useBinder(...);

     return (
          <>
               <FullNameComponent personModel={model}></FullNameComponent>
          </>
     );
}
```

And TBH, I thought I'm making it look complicated unnecessarily. How could I improve this?

Lodin: As for me, it is pretty simple and well-splitted code. So I don't think it's gonna confuse the developer.

I fixed some places in the code you've provided to make it more consistent and extendable.

```ts
interface PersonProps { 
    model: StringModel; 
}

function FullNameComponent({ model }: PersonProps) {
  const { model, field, required, errors, invalid }= useBinderNode(model);

  return (
    <>
      <label htmlFor="fullName">
        Full name
        {required ? '*' : ''}
      </label>
      <input id="fullName" {...field(model)}></input>
      <br/>
      <span className="label" style={{visibility: invalid ? 'visible' : 'hidden'}}>
          <strong>
           {errors[0]?.message}
          </strong>
        </span>
    </>
  );
}

export default function Main() {
   const { model } = useBinder(PersonModel);
   return <FullNameComponent model={model.fullName} />;
}
```

taefi:
> Also, @taefi, what about using separate TS files to make sure they compile without errors?

I agree to improve them after the DX tests are conducted, as I believe there would be some enhancements coming to other articles and by that time a working pre-release would be available.

articles/lit/guides/forms/reference.adoc
```diff
@@ -26,17 +48,58 @@ The [methodname]`field()` directive does the main job of binding the form field
...
+----
+import { useBinder, useBinderNode } from '@hilla/react-form';
+import {StringModel} from "@hilla/form";
+...
+interface FullNameProps {
+    fullNameModel: StringModel;
+}
+
+function FullNameComponent({ fullNameModel }: FullNameProps) {
+  // @ts-ignore
```
Lodin, line 76: You also shouldn't put any `@ts-ignore` in docs unless it is absolutely necessary. It is pretty confusing.

articles/lit/guides/forms/reference.adoc
```diff
@@ -18,6 +24,22 @@ The [methodname]`field()` directive does the main job of binding the form field
...
+[source,typescriptjsx]
+----
+import {TextField} from "@hilla/react-components/TextField.js";
+...
+const { model, field } = useBinder(PersonModel);
+...
+return (
+  <TextField label="Full name"
+             {...field(model.fullName)}>
+  </TextField>
```
Lodin, line 38: Closing tag is not necessary here:

```suggestion
  <TextField label="Full name" {...field(model.fullName)} />
```

Lodin (changes requested):

Lodin (approved):

PR #2758, opened 2023-09-01
Remove hilla maven dependency and unused hilla java classes.

manolo:
at the same time, removing some deprecated code


ZheSun88 (approved):

tomivirkki:
`DummyEndpoint` was there for a reason. Without it, the TS domain types will not get generated. Currently `npm run dspublisher:build` doesn't run because of this.

manolo:
Ah ok, build passed, probably not testing that dsp thing, will take a look whether we can remove hilla bom, otherwise we need to downgrade docs, or have a proper prereleases hilla depending on prereleases flow and wc

PR #2761, opened 2023-09-01
restore maven dependencies, but not use hilla.bom and hilla.jar

manolo:
This changes should fix problem introduced in https://github.com/vaadin/docs/pull/2758. 
So as we can use hilla classes but still not introduce outdated frontend dependencies


src/main/java/com/vaadin/demo/DashboardEndpoint.java
```diff
@@ -0,0 +1,16 @@
+package com.vaadin.demo;
```
tomivirkki, line 1: Looks like `DashboardEndPoint` is still not needed.

`DummyEndpoint` is because of https://github.com/vaadin/docs/pull/2758#issuecomment-1702447517 (might be a good idea to document this in `DymmyEndpoint`)

...and `DataEndpoint` seems to be [referenced by `ts-view-with-endpoint.ts`](https://github.com/vaadin/docs/blob/30d10d846da6c81171829bb36fced0c773eb8aaa/frontend/demo/pwa/offline/ts-view-with-endpoint.ts#L9)

tomivirkki (approved):

PR #2763, opened 2023-09-04
fix: Improve navigation menu discoverability

jouni:
Show the first few items in the components navigation even when collapsed, making the following category more easily discoverable.

Before:
Screenshot 2023-09-04 at 13 24 03

After:
Screenshot 2023-09-04 at 13 22 00

### DIFF
--- a/articles/react/components/index.asciidoc
+++ b/articles/react/components/index.asciidoc
@@ -1,6 +1,6 @@
 ---
 title: Components
-section-nav: expanded flat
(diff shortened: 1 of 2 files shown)


sissbruecker (approved):

jouni:
…and right when I click the merge button I remember one issue I was supposed to fix before merging: when the menu is collapsed to the drawer, the overflow gradient is the wrong color (only in the dark theme).

PR #2764, opened 2023-09-04
Add LoginOverlay custom form area and footer examples

web-padawan:
## Description

Depends on https://github.com/vaadin/flow-components/pull/5390

Added TS and Java examples for `vaadin-login-overlay` custom content.


articles/components/login/index.asciidoc
```diff
@@ -2,19 +2,19 @@
 tab-title: Usage
 layout: tabbed-page
 title: Login
-description: Login is a component that contains a log-in form. You can use it to authenticate the user with a username and password.
+description: How to use the Login component to authenticate users.
```
rolfsmeds, line 5: @russelljtdyer I'm not a fan of this style of description. Component pages are _not_ "how to" pages, they're broad descriptions of the features and usage of a certain components, so this type of descriptions seems misleading. I would prefer that we stick to the descriptions already provided in the tag::description[] block below.

articles/components/login/index.asciidoc
```diff
@@ -2,19 +2,19 @@
...
   - 'Source: https://github.com/vaadin/web-components/tree/v{moduleNpmVersion:@vaadin/login}/packages/login[TypeScript] / https://github.com/vaadin/flow-components/tree/{moduleMavenVersion:com.vaadin:vaadin}/vaadin-login-flow-parent[Java]'
 ---
 
 
-= Login
+= Login Component
 
 // tag::description[]
-Login is a component that contains a log-in form. You can use it to authenticate the user with a username and password.
+Login is a Vaadin component that contains a log-in form. You can use it to authenticate the user with a username and password.
```
rolfsmeds, line 15: @russelljtdyer do we need to specify _Vaadin_ here? These are _all_ Vaadin components, after all. We want to keep these descriptions brief since they are used in the thumbnail index at https://vaadin.com/docs/latest/components. I don't think anybody expect these components to be, say, Apple components.

articles/components/login/index.asciidoc
```diff
@@ -199,11 +198,106 @@ include::{root}/frontend/demo/component/login/react/login-overlay-header.tsx[tag
 endif::[]
 --
 
+
+[role="since:com.vaadin:vaadin@V24.2"]
+=== Custom Form Area
+
+Overlay provides a custom form area for adding fields in addition to username and password. This area is placed above the "Submit" button. Use `name` attribute to ensure custom field `value` is submitted with the form.
```
rolfsmeds, line 205:
```suggestion
The overlay provides a custom form area for adding fields in addition to username and password. This area is placed above the "Submit" button. Use `name` attribute to ensure custom field `value` is submitted with the form.
```

articles/components/login/index.asciidoc
```diff
@@ -199,11 +198,106 @@ include::{root}/frontend/demo/component/login/react/login-overlay-header.tsx[tag
...
+include::{root}/frontend/demo/component/login/react/login-overlay-custom-form-area.tsx[tags=snippet,indent=0,group=React]
+----
+endif::[]
+--
+
+
+[role="since:com.vaadin:vaadin@V24.2"]
+=== Footer
+
+Footer can be used for placing additional custom content, such as text, buttons, etc. Adding components after the overlay is opened is not supported.
```
rolfsmeds, line 253:
```suggestion
The footer area can be used for placing additional custom content, such as text, buttons, etc. Adding components after the overlay is opened is not supported.
```

rolfsmeds (changes requested):
LGTM otherwise :)

articles/components/login/index.asciidoc
web-padawan: This change is unrelated to the PR which is about new features. Maybe we should revert it?

articles/components/login/index.asciidoc
```diff
@@ -2,19 +2,19 @@
...
-description: Login is a component that contains a log-in form. You can use it to authenticate the user with a username and password.
+description: How to use the Login component to authenticate users.
 page-links:
   - 'API: https://cdn.vaadin.com/vaadin-web-components/{moduleNpmVersion:@vaadin/login}/#/elements/vaadin-login-overlay[TypeScript] / https://vaadin.com/api/platform/{moduleMavenVersion:com.vaadin:vaadin}/com/vaadin/flow/component/login/LoginOverlay.html[Java]'
   - 'Source: https://github.com/vaadin/web-components/tree/v{moduleNpmVersion:@vaadin/login}/packages/login[TypeScript] / https://github.com/vaadin/flow-components/tree/{moduleMavenVersion:com.vaadin:vaadin}/vaadin-login-flow-parent[Java]'
 ---
 
 
-= Login
+= Login Component
```
web-padawan, line 12: None of other components have `Component` in the title. I'd suggest to revert this change as well.

articles/components/login/index.asciidoc
rolfsmeds: Thanks for the overview, that helps us understand what's going on. But that doesn't really seem to address the fact that the new description indicates a "How to" page about a specific function while the contents of the page are much broader than that. In fact, this page hardly answers that question _at all_. The Flow documentation has a page about how to actually perform the authentication in Flow: https://vaadin.com/docs/latest/tutorial/login-and-authentication

articles/components/login/index.asciidoc
rolfsmeds: The same could be said of most of our components: Button, Text Field, etc. Or perhaps you mean that the word "login" on its own doesn't indicate that we're talking about a login _UI component_, and I could agree with that. But then "Vaadin" is not the keyword at all, but "UI", so we could change it to say _"Login is a UI component"_ rather than _"Login is a Vaadin component"_.

articles/components/login/index.asciidoc
rolfsmeds: Seems like the same could be said of many others as well, such as Select, CRUD, Cookie Consent, Details, Message Input, Notification, and Upload. Would you suffix those with "Component" as well then, as well as the links to the pages in the navigation column, and on the component thumbnail index page?

web-padawan: What Rold said. Also, as a developer looking for an example of using a component, I think `Component` would be rather redundant - especially in the navigation column and thumbnail.

web-padawan: Reverted this change for now so we can proceed. Let's consider it in a separate PR.

articles/components/login/index.asciidoc
web-padawan: Reverted this change for now so we can proceed. Let's consider it in a separate PR.

articles/components/login/index.asciidoc
jouni: I suppose I didn't mention this clear to @russelljtdyer some days ago, but I did go and revert/replace the "How to" version of the descriptions to the ones we have inside the `tag::description[]` tags (on the component pages).

Let’s keep that consistent here as well, and use the same text for both the `description` frontmatter attribute, and the `tag::description[]` tag.

web-padawan: Thanks, I already restored the original description in [96d3310](https://github.com/vaadin/docs/pull/2764/commits/96d3310d5913d5cbae19397ddd58ad242f70a474).

rolfsmeds (approved):

PR #2770, opened 2023-09-05
Add docs for Configuring Spring Boot Development Tools

tltv:
Adds new chapter in Spring Configuration page for Spring Development Tool properties that helps to restart time.

Related-to: vaadin/platform#4404


articles/integrations/spring/configuration.asciidoc
```diff
@@ -102,5 +102,34 @@ public class MyConfiguration {
...
+For example, in small project developed with Eclipse, using following smaller values can speed up the restart time when changing one or few classes:
+
+.application.properties
+[source,properties]
+----
+spring.devtools.restart.poll-interval=100ms
+spring.devtools.restart.quiet-period=50ms
+----
+
+Or for example, in small or medium project developed with IntelliJ IDEA, increasing values can make sure that restart happens only once in a row after changing one or few classes:
```
mshabarov, line 122: Just to be sure, "small" is a proper word, not "large", right?

tltv: Actually it means IDEA in general. Any size of app causes multi-restarts with default values. I'll update it to "Or for example, project developed with IntelliJ...."

mshabarov (approved):

PR #2797, opened 2023-09-12
docs: add Hilla React routing documentation

sissbruecker:
Adds a basic guide on how to implement routing with Hilla React.

### DIFF
--- a/articles/react/guides/routing.adoc
+++ b/articles/react/guides/routing.adoc
@@ -33,7 +33,7 @@ The router instance is then used in the [filename]`App.tsx` file to render the a
 
 [source,tsx]
 ----
-import router from 'Frontend/routes.js';
+import router from 'Frontend/routes';
 import { RouterProvider } from 'react-router-dom';
 
 export default function App() {
--- a/articles/react/guides/routing.adoc
+++ b/articles/react/guides/routing.adoc
@@ -85,7 +85,7 @@ import { NavLink } from 'react-router-dom';
 <NavLink to="/my-route">My Route</NavLink>
 ----
 
-This will create a clickable link in your application that navigates to the specified route.
+This creates a clickable link in your application that navigates to the specified route.
 
 
 == Adding Routes with Parameters
@@ -162,7 +162,7 @@ export default function MainLayout() {
 }
 ----
 
-Now, when the `/about` route is active, the title `About us` will be displayed in the header.
+Now, when the `/about` route is active, the title `About us` is displayed in the header.
 
 
 == Programmatic Navigation
@@ -192,7 +192,7 @@ function ProductDetailView() {
 }
 ----
 
-By default, this will push a new entry to the browser's navigation history. If you want to replace the current entry instead, you can pass `{ replace: true }` as the second argument like so:
+By default, this pushes a new entry to the browser's navigation history. If you want to replace the current entry instead, you can pass `{ replace: true }` as the second argument like so:
 
 [source,tsx]
 ----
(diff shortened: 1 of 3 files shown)


articles/react/guides/routing.adoc
```diff
@@ -0,0 +1,242 @@
...
+export default createBrowserRouter(routes);
+----
+
+In this example, two routes are defined: the root path `/` and the `/about` path. These routes are associated with different components. The `<MainLayout />` component acts as the layout template for the entire application. The last line creates and exports the router instance.
+
+The router instance is then used in the [filename]`App.tsx` file to render the application:
+
+[source,tsx]
+----
+import router from 'Frontend/routes.js';
```
taefi, line 36: Is the `.js` extension needed? Or can it be omitted?
```suggestion
import router from 'Frontend/routes';
```

articles/react/guides/routing.adoc
```diff
@@ -0,0 +1,238 @@
...
+  return <RouterProvider router={router} />;
+}
+----
+
+In the next section are the details of how to work with routing in Hilla.
+
+
+== Adding Routes
+
+To add a new route to an application, first create a new component that should be displayed when the route is active. For example, create a file called [filename]`MyRoute.tsx` in the `views` folder with the following content:
```
taefi, line 49: Would it be easier to distinguish the View component from the route if we name it something like `MyView` instead of `MyRoute`?

articles/react/guides/routing.adoc
```diff
@@ -0,0 +1,238 @@
...
+
+To access this metadata from within a component, you can use the `useMatches` hook from `react-router-dom`. In the following example, the page title is used to display it in the header of the main layout:
+
+[source,tsx]
+----
+import { useMatches } from "react-router-dom";
+
+export default function MainLayout() {
+  const matches = useMatches();
+  const currentHandle = matches[matches.length - 1]?.handle as any;
```
taefi, line 148: Probably, not a comment on this documentation, but more related to the API:
Is it possible to hide these two in an API?

articles/react/guides/routing.adoc
sissbruecker: I don't think it's necessary, but it is how the starter generates it. Let's just omit it here.

Could maybe clarify in the Daily if we really need to confuse devs by giving them TS code that seems to import non-existing files.

articles/react/guides/routing.adoc
sissbruecker: Good point, `MyRoute` is a weird name for an example component. Changed it to use `MyView`.

articles/react/guides/routing.adoc
sissbruecker: This is the best that react-router gives you out of the box for working with the handle. They decided to type `handle` as `unknown` which forces you to cast it to something useful. I'd rather not add a helper for this to the starter. Maybe we could follow this code example up with how to implement a hook that gets the handle for the current route, maybe casted to a specific type.

taefi: I like that hook more than a raw `useMatch` and casting. I'd suggest naming that as `useCurrentRoute`, but no pressure, just suites my taste better :)

articles/react/guides/routing.adoc
taefi: That's a fair argument, it is kinda scary to see a starter project with so many errors, and having faith in their resolution after the first run :)

articles/react/guides/routing.adoc
sissbruecker: Well, it really only gives you the `handle` and not the rest of the route. For now I've added a note below the example to reference the helper hook.

taefi (approved):

PR #2808, opened 2023-09-19
Show React docs cards at root

tarekoraby:
Disclaimer: I didn't test this change as I cannot run the docs locally.


jouni:
The React index page was missing the tagged region. I added that. Here’s how it looks like, with the top level sections explicitly expanded in the side nav:
Screenshot 2023-09-21 at 16 13 01

jouni:
Need to restore `root: true` at least.

jouni:
After fixes:
Screenshot 2023-09-21 at 16 22 57

tarekoraby:
I think that serves as a good band-aid for now. We can revisit the topic and think of other ideas later. Feel free to merge at any time.

PR #2824, opened 2023-09-25
chore: update to vaadin 24.2.0.alpha11

jouni:


jouni:
Tested the build locally ✅

PR #2825, opened 2023-09-25
chore: update to react-components 2.2.0-beta1

jouni:


jouni:
Tested the build locally ✅

PR #2826, opened 2023-09-25
Add chart examples

jouni:


alvarezguille:
Some general comments that can be improved in a separate PRs
* Bullet chart example could use a fixed height 
image

* Code snippet for OHLC are hidden by default
image

* It's not so obvious that the implementation of `getRawData` is available when clicking "Expand code"
* Some of the new live demos are the same example that was previously used with a screenshot and code is duplicate in some cases.

alvarezguille (approved):
:shipit:

PR #2827, opened 2023-09-26
Update `hilla` from `latest` (component pages)

jouni:
Pull component page updates from latest. Fix all AsciiDoc compilation errors at the same time.


jouni:
Local build is successful ✅

PR #2830, opened 2023-09-27
Addition of proper constructor

kody435:
Binder<Contact> binder = new BeanValidationBinder<>(Contact.class); ❌

⬇️

BeanValidationBinder<Contact> binder = new BeanValidationBinder<>(Contact.class); ✅


mstahv (approved):
The only API difference (after constructors) between Binder and BeanValidationBinder is requiredConfigurator field (and its getter/setter). I have never needed that, but I guess it can be handy 👍

PR #2837, opened 2023-09-27
chore: update hilla to 2.3.0.alpha5

ZheSun88:


taefi:
What about upgrading to `2.3.0.alpha6` directly?

PR #2838, opened 2023-09-27
chore: update hilla to 2.3.0.alpha5 (hilla)

ZheSun88:


taefi:
What about upgrading to `2.3.0.alpha6` and `24.2.0.beta2` directly?

ZheSun88:
sure.. maybe create another PR for that version

PR #2836, opened 2023-09-27
docs: add AutoGrid article

platosha:

PR #2845, opened 2023-09-29
docs: add auto grid guide

sissbruecker:
Adds a guide for using the auto grid.

Currently based on #2836 to be able to render an example of the auto-grid.


articles/react/guides/data-grids.adoc
sissbruecker, line 1: Decided to put this under "Guides". If I was scanning the docs for features, I wouldn't look under "Getting Started".

Regarding the article name, I chose "Data Grids", as I'll probably not associate much with an "Auto Grid", but can relate to what a data grid is.

Both are just suggestions, let me know what you think.

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,198 @@
...
+spring.datasource.password=password
+spring.jpa.hibernate.ddl-auto=update
+----
+
+In order to have some example data, download this SQL script and put it in the `src/main/resources` folder:
+++++
+<p>
+<a
+class="button primary water"
+href="https://hilla.dev/docs/connect/_download?file=product-data.sql"
```
sissbruecker, line 149: Note that the download URL will only work in production, at least I hope it does. I couldn't find a way to get the app's context path from asciidoc.

src/main/java/com/vaadin/demo/DownloadController.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.Map;
+
+@RestController
+@RequestMapping("/connect/_download")
+public class DownloadController {
```
sissbruecker, line 16: This adds the capability to download something from the docs app, which is used in the article to download the SQL file containing example data. It needs to be mapped to some path that the docs app forwards to the Vaadin server, `/connect` was the one I got to work.

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,210 @@
...
+----
+import React from 'react';
+import { AutoGrid } from '@hilla/react-grid';
+import { ProductService } from 'Frontend/generated/endpoints';
+import ProductModel from 'Frontend/generated/com/example/application/ProductModel';
+
+export function ProductsView() {
+  return (
+    <div className='p-l'>
+      <AutoGrid service={ProductService} model={ProductModel} headerFilters={true}/>
```
Artur-, line 178:
```suggestion
      <AutoGrid service={ProductService} model={ProductModel} headerFilters />
```

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,210 @@
+---
+title: Data Grids
+description: "Learn how to quickly prototype data grids using the Auto Grid feature."
```
tarekoraby, line 3: I suppose it's possible to use AutoGrid for the real thing: not just for prototyping.

```suggestion
description: "Create data grids using the Auto Grid feature."
```

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,210 @@
...
+order: 71
+---
+
+= Data Grids
+
+Auto Grid is a powerful feature that makes it easy to show data from your database in a data grid. It is a great way to quickly prototype your application so that you have something to show. It works by connecting a JPA repository to a grid component with minimal effort. The grid automatically shows columns for all properties of the entity and loads data in a lazy-loaded manner. It can also automatically provide sorting and filtering.
```
tarekoraby, line 9:
```suggestion
Auto Grid makes it easy to show data from the database in a data grid. It works by connecting a JPA repository to a grid component with minimal effort. The grid automatically shows columns for all properties of the entity and loads data in a lazy-loaded manner. It can also automatically provide sorting and filtering.
```

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,210 @@
+---
+title: Data Grids
+description: "Learn how to quickly prototype data grids using the Auto Grid feature."
+order: 71
```
tarekoraby, line 4: Perhaps Martin would find this more valuable than forms.

```suggestion
order: 65
```

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,210 @@
...
+---
+
+= Data Grids
+
+Auto Grid is a powerful feature that makes it easy to show data from your database in a data grid. It is a great way to quickly prototype your application so that you have something to show. It works by connecting a JPA repository to a grid component with minimal effort. The grid automatically shows columns for all properties of the entity and loads data in a lazy-loaded manner. It can also automatically provide sorting and filtering.
+
+This tutorial shows how to set up a JPA entity and repository in the backend, how to expose that repository with a service, and how to connect that service to an Auto Grid in the frontend. If you don't have a Hilla project yet, you can quickly create one by running the following command from the CLI:
```
tarekoraby, line 11: Martin already knows how to set up a JPA entity and repository. Let's not teach him how to do these things, but rather focus on showing him how he can connect his JPA entities and repositories to show a grid on the frontend.

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,210 @@
...
+
+[source,shell]
+----
+npx @hilla/cli init my-hilla-app
+----
+
+== Setting up the Backend
```
tarekoraby, line 18: @Artur-, wasn't there a starter that already comes with the necessary dependencies in `pom.xml` + `application.properties` settings + sample data?

If so, we can skip the need to ask people to copy-paste a lot of code, and just go directly to the step where they add a `BrowserCallabe`. We can still list and give a brief explanation of the content of the project, but perhaps we can do so in a tabbed example similar to the one [here](https://github.com/vaadin/docs/blob/57d5b6ec1d7b9c380966705a10b4e1b959f79c91/articles/react/components/auto-grid/index.asciidoc?plain=1#L20), and/or using a collapsible section similar to the one used in the basics tutorial [here](https://github.com/vaadin/docs/blob/3d0080fb547d8aac74289956d7796bcdcf6ddd7a/articles/lit/start/basics/index.adoc?plain=1#L65).

If we don't have such a preset, then it would be valuable to create one IMO, as the current draft requires too much copy-pasting of things that are already familiar to Martin, and that don't highlight the value of Auto Grid.

sissbruecker: I've moved the database setup into a collapsible section. Downloading and adding product data is still part of the main guide, as you won't see anything otherwise if you follow the steps to create a new entity.

articles/react/guides/data-grids.adoc
sissbruecker: We could rework it so that creating a new entity is optional. Still, there are several places that reference the entity, model, and service, so it could end up being convoluted to constantly differentiate between those two options.

I feel in the current shape it already does a decent job of promoting the feature and giving you a complete setup to actually try it out. If a developer already has something (entity + data in database), I think they would be able to adapt the steps to use their own stuff.

articles/react/guides/data-grids.adoc
```diff
@@ -0,0 +1,215 @@
...
+----
+import React from 'react';
+import { AutoGrid } from '@hilla/react-grid';
+import { ProductService } from 'Frontend/generated/endpoints';
+import ProductModel from 'Frontend/generated/com/example/application/ProductModel';
+
+export function ProductsView() {
+  return (
+    <div className='p-l'>
+      <AutoGrid service={ProductService} model={ProductModel} headerFilters />
```
tarekoraby, line 183:
```suggestion
      <AutoGrid service={ProductService} model={ProductModel} />
```
Header filters are now on by default (there is `noHeaderFilters` to hide them)

tarekoraby (approved):

articles/react/guides/data-grids.adoc
tarekoraby: https://github.com/vaadin/hilla/pull/1406

PR #2851, opened 2023-10-02
Improvements to Notification examples

mstahv:


src/main/java/com/vaadin/demo/component/notification/NotificationUndo.java
```diff
@@ -66,6 +55,25 @@ public Notification show() {
         return notification;
     }
 
+    public class UndoButton extends Button {
+        public UndoButton() {
+            super("Undo");
+            addThemeVariants(ButtonVariant.LUMO_TERTIARY_INLINE);
+            setAriaLabel("Close");
```
OlliTietavainenVaadin, line 62: Should this aria label say "Undo" instead of "Close"?

mstahv: The original code (before refactoring) had no ARIA label. Maybe I'll remove that altogether, as I think all those ARIA label's are irrelevant for these examples and should be (and maybe are) covered in some other part of our docs.

mstahv: removed

jouni (approved):

jouni:
Most of the inline styling could be replaced with Lumo utility classes, if you wish to clean it up a bit more.

PR #2859, opened 2023-10-03
Fixed Typo

Leviand:
Instead of "Number Field" it was "Number Filed"


alvarezguille (approved):

jouni:
Thanks a lot for noticing and fixing it!

PR #2866, opened 2023-10-05
chore: reset loading rows background in grid styling example

tomivirkki:
Reset the background color of loading rows in the Grid styling example. Couldn't find a way to achieve this by modifying the existing selectors (instead of overriding with a new selector).

Before:

https://github.com/vaadin/docs/assets/1222264/63379269-e40c-43e1-b13c-670fff64cdbe

After:

https://github.com/vaadin/docs/assets/1222264/c26cc390-92a3-4874-ba5b-a6618db4b3cb

### DIFF
--- a/frontend/demo/component/grid/grid-styling.ts
+++ b/frontend/demo/component/grid/grid-styling.ts
@@ -39,7 +39,11 @@ export class Example extends LitElement {
 
   protected override render() {
     return html`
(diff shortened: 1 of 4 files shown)


jouni (approved):

jouni:
I must say, that this seems more like a bug in the component, rather than a documentation issue. The custom part names should be cleared at the same time as the row contents are cleared.

PR #2868, opened 2023-10-05
refactor: grid drag drop example should set parent correctly

ugur-vaadin:
Follow-up from a Flow component [issue](https://github.com/vaadin/flow-components/issues/5530).

The example removes the item on drop before adding it under a parent again. This makes the item to lose its children, which is the way `TreeData.remove` works. To update the parent of an item in the tree, `TreeData.setParent` should be used instead. 

This PR updates the example to set the parent correctly.


tomivirkki (approved):

leewatt:
It still doesn't work when the node is collapsed. Please refer to the updated steps in the original issue.

PR #2880, opened 2023-10-11
feat: Documentation for redirecting to errorView

caalador:
Adds documentation for showing non
route annotated error views for
exceptions that happend during RPC
calls.

part of vaadin/flow#4715

### DIFF
--- a/articles/advanced/custom-error-handler.asciidoc
+++ b/articles/advanced/custom-error-handler.asciidoc
@@ -42,6 +42,40 @@ VaadinSession.getCurrent().setErrorHandler(new CustomErrorHandler());
 To apply the custom error handler to all user sessions, you can use a [classname]`SessionInitListener`, which receives an event each time a new [classname]`VaadinSession` is initialized.
 See the <<session-and-ui-init-listener#,Session and UI Listeners>> article to learn how to create a [classname]`SessionInitListener`.
 
+[role="since:com.vaadin:vaadin@V24.3"]
+== Showing Error Parameter Views For Non Navigation Exceptions
+
+In the [classname]`DefaultErrorHandler` there is the possibility to enable transitioning to an [interfacename]`HasErrorParameter<T extends Exception>` error view on exceptions by setting the `enableErrorHandlerRedirect` parameter to `true`.
+
+For more information on the [interfacename]`HasErrorParameter` see <<{articles}/routing/exceptions#, Error Resolving>>.
+
+For a customized error handler the same can be done by using the [classname]`ErrorHandlerUtil` method [methodname]`handleErrorByRedirectingToErrorView`
+
+[source,java]
+----
+public class CustomErrorHandler implements ErrorHandler {
+
+    private static final Logger logger = LoggerFactory.getLogger(CustomErrorHandler.class);
+
+    @Override
+    public void error(ErrorEvent errorEvent) {
+        boolean redirected = ErrorHandlerUtil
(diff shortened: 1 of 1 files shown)


articles/advanced/custom-error-handler.asciidoc
```diff
@@ -42,6 +42,40 @@ VaadinSession.getCurrent().setErrorHandler(new CustomErrorHandler());
 To apply the custom error handler to all user sessions, you can use a [classname]`SessionInitListener`, which receives an event each time a new [classname]`VaadinSession` is initialized.
 See the <<session-and-ui-init-listener#,Session and UI Listeners>> article to learn how to create a [classname]`SessionInitListener`.
 
+[role="since:com.vaadin:vaadin@V24.3"]
+== Showing Error Parameter Views For Non Navigation Exceptions
```
mshabarov, line 46: I believe the main beneficiaries of this feature are `HasErrorParameter` users.

Thus, this update misses a note in [HasErrorParameter docs](https://vaadin.com/docs/latest/routing/exceptions#error-resolving), saying the following:
1. If a project has only `HasErrorParameter` error views, one can just enable `enableErrorHandlerRedirect=true` for extending them for the RPC scoped exceptions.
2. If a projects defines both `HasErrorParameter` and `ErrorHandler` implementations and they want to use this feature, they need, as an addition to the parameter, modify their custom error handler as shown in this chapter.

mshabarov (changes requested):

articles/advanced/custom-error-handler.asciidoc
caalador: Added info to exceptions doc also.

articles/routing/exceptions.asciidoc
```diff
@@ -235,5 +235,11 @@ public class FaultyBlogPostHandler extends Component
 }
 ----
 
+[role="since:com.vaadin:vaadin@V24.3"]
+== Showing Error View For Exception During RPC Call
+
+To use registered error views outside routing and rerouting the application can set the `enableErrorHandlerRedirect` parameter to `true`.
+
+This will enable updating the current view content to a registered [interfacename]`HasErrorParameter<T extends Exception>` that handles the exact exception thrown for RPC events also.
```
mshabarov, line 243: I propose to give a link to the chapter above and explain when they need `ErrorHandlerUtil`, because it might be that the feature wouldn't work for them because of custom error handler and they don't know why (second part isn't linked anyhow = invisible).

articles/routing/exceptions.asciidoc
```diff
@@ -235,5 +235,11 @@ public class FaultyBlogPostHandler extends Component
 }
 ----
 
+[role="since:com.vaadin:vaadin@V24.3"]
+== Showing Error View For Exception During RPC Call
+
+To use registered error views outside routing and rerouting the application can set the `enableErrorHandlerRedirect` parameter to `true`.
```
mshabarov, line 241: I would also add `enableErrorHandlerRedirect` to the "Configuration" chapter, so it would be in the list.

mshabarov (changes requested):

articles/routing/exceptions.asciidoc
caalador: added

articles/routing/exceptions.asciidoc
caalador: added

mshabarov (approved):

PR #2887, opened 2023-10-16
Improve PreserveOnRefresh documentation

tepi:
Fixes https://github.com/vaadin/flow/issues/6395 for v14


tepi:
Once this is ok, probably needs to be changed on v22, v23 and latest as well.

tltv (approved):

vite.generated.ts
```diff
@@ -0,0 +1,821 @@
+/**
```
tepi, line 1: Was this file added on purpose?

jouni: No. I’ll remove it.

PR #2890, opened 2023-10-16
Docs/styling docs fixes

rolfsmeds:
- Updated SideNav selectors to 24.2 version
- Added readonly field border selector to textfield et al


articles/components/combo-box/styling.asciidoc
```diff
@@ -32,6 +32,7 @@ Toggle Button:: `vaadin-combo-box+++<wbr>+++**::part(toggle-button)**`
 Clear button:: `vaadin-combo-box+++<wbr>+++**::part(clear-button)**`
 Clear button icon:: `vaadin-combo-box+++<wbr>+++**::part(clear-button)::before**`
 Prefix element:: `vaadin-combo-box+++<wbr>+++** > [slot="prefix"]**`
+Readonly border:: `vaadin-combo-box+++<wbr>+++**::part(input-field)::after**`
```
web-padawan, line 35: Let's use `Read-only` as in the states list.

rolfsmeds: Indeed. Fixed in latest commit.

PR #2892, opened 2023-10-17
fix: use static hosting for product data download

sissbruecker:
Updates the DS Publisher build script to copy product data file into the static content hosted by nginx, so that it can be downloaded from there.


jouni:
Perhaps we should move this to `dspublisher/theme/init.ts` instead, so that it doesn't leak to customer DSP installations?

jouni:
Right, indeed. Not a problem then.

taefi (approved):

Issue #1521, opened 2022-06-27
Mention the `eagerServerLoad` flag for a 404 response in the Router Exception Handling article


platosha:
The `eagerServerLoad` flag in Flow enables a 404 response for initial requests with client-side bootstrapping when the server-side route matching the requested path is not found.

By default with client-side bootstrapping (since Vaadin 15), without `eagerServerLoad` enabled, the server response is `200 OK` in that case to facilitate “single page” applications with client-side views.

The `Router Exception Handling` article currently describes the 404 response, which is only the case with the `eagerServerLoad` enabled. This is misleading, because the default behavior is different.

tarekoraby:
@platosha that only applies to Flow docs? or is there also somewhere to mention this in Hilla's?

PR #2896, opened 2023-10-19
docs: note about eagerServerLoad property in exceptions article

tltv:
Adds a description about `eagerServerLoad` property for Flow, which is needed in cases when HTTP code response from `RouteNotFoundError` has to go to the browser.

Fixes: #1521

### DIFF
--- a/articles/routing/exceptions.asciidoc
+++ b/articles/routing/exceptions.asciidoc
@@ -45,9 +45,9 @@ Exceptions are matched in the order below:
 
 The `404` [classname]`RouteNotFoundError` (for [classname]`NotFoundException`), and `500` [classname]`InternalServerError` (for [classname]`java.lang.Exception`) are implemented by default.
 
-By default, the initial request loads the client-side bootstrapping first with a HTTP 200 response and an error page is rendered later. The HTTP 404 response from [classname]`RouteNotFoundError` is only processed in the server side and never returned to the browser.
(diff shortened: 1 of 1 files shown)


articles/routing/exceptions.asciidoc
```diff
@@ -45,6 +45,14 @@ Exceptions are matched in the order below:
...
 The `404` [classname]`RouteNotFoundError` (for [classname]`NotFoundException`), and `500` [classname]`InternalServerError` (for [classname]`java.lang.Exception`) are implemented by default.
 
+[NOTE]
+====
+Activate <<{articles}/configuration/properties/#properties,`eagerServerLoad` property>>
+to make [classname]`RouteNotFoundError` return `404` HTTP response for the navigation request in the browser.
+By default, initial request loads client-side bootsrapping first with `200` HTTP response
+and error page is rendered later. `404` HTTP response from [classname]`RouteNotFoundError`
+is only processed in the server side and never returned to the browser.
+====
```
jouni, line 55: I don't think this needs to be a note admonition block.

I’m not entirely sure if this is correct and clear to the reader. I would like to see a clearer description of what happens by default and what the `eagerServerLoad` property controls.

```suggestion
By default, the initial request loads the client-side bootstrapping first with a HTTP 200 response and an error page is rendered later. The HTTP 404 response from [classname]`RouteNotFoundError` is only processed in the server side and never returned to the browser.

If you enable the <<{articles}/configuration/properties/#properties,`eagerServerLoad` property>>,[classname]`RouteNotFoundError` returns a HTTP 404 response for the navigation request. 
```

tltv: I'll update the text a bit to make it more clear.

mcollovati (approved):

mshabarov (approved):

PR #2930, opened 2023-10-30
docs: add new input fields theming properties docs

web-padawan:
Part of https://github.com/vaadin/web-components/issues/2954

This PR covers input field components (i.e. those that use `vaadin-input-container` internally).


articles/components/_styling-section-theming-props.adoc
```diff
@@ -0,0 +1,174 @@
+
+// tag::input-fields[]
+[role="since:com.vaadin:vaadin@V24.3"]
+== Style Properties
```
rolfsmeds, line 4: This section could have its own tag so that it can be used with non-input-fields as well.
(As the "input field components shared styles" admonition is the only part specific to input fields)

articles/components/_styling-section-theming-props.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+|Read-only border
+|`--vaadin-input-field-readonly-border`
+|`1px dashed --lumo-contrast-30pct`
+
+|===
+// end::input-surface[]
+
+
+// tag::label[]
```
rolfsmeds, line 78: Maybe the label, helper and error sections could all be in a single tag, so they could be inclued with a single include? They're shared by all input fields after all, the only part that needs to be separate is input surface and icons (and I guess icons could be part of input surface section since they're rendered inside the surface).

Then there's the detail that some input fields have extra features in the surface, like toggle buttons. So ideally we'd like to _extend_ that section with component-specific properties for those. I suppose that would require keeping the table block "open"?

rolfsmeds (changes requested):

articles/components/_styling-section-theming-props.adoc
web-padawan: There is the `"the appearance of Vaadin input field components."` statement in the first sentence. Should we rephrase it to something like `"the appearance of this component"` instead?

articles/components/_styling-section-theming-props.adoc
web-padawan: Combined label, helper and error to `label-helper-error` include.

As for other features, I'd prefer to avoid unclosed tables if possible. After all, we don't currently have any custom CSS properties that cover these individual component-specific properties at this pointl.

articles/components/_styling-section-theming-props.adoc
rolfsmeds: Ah, indeed, yes!

articles/components/_styling-section-theming-props.adoc
rolfsmeds: Yeah, I suppose we can add those to a separate section once we get there. It's not ideal, but everything's a compromise :)

articles/components/_styling-section-theming-props.adoc
web-padawan: Extracted "Style Properties" into separate include block so it can be used iwth other components too.

rolfsmeds (approved):

PR #2932, opened 2023-10-31
add auto form and auto crud component documentation

taefi:


articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,197 @@
+---
+title: Auto Form
+order: 10
+description: Used for auto populating form fields and actions based on a Java backend service.
+---
+
+= [since:dev.hilla:hilla@v2.4]#Auto Form#
+:react:
```
jouni, line 8: I assume there’s only a React-version for AutoGrid, so this attribute is unnecessary.
```suggestion
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,197 @@
...
+
+=== Customizing Visibility & Order
+
+To choose which of the data properties should be displayed as fields in Auto Form, or to specify the order of appearance in the Auto Form, use the `customLayoutRenderer` property and specify the name of desired properties in the `template`.
+
+When Auto Form is used without any template defined for `customLayoutRenderer` property, it renders all the properties of the entity except those that are annotated with `@Id` or `@Version` annotations. You can make them visible by adding their names to the `customLayoutRenderer`'s template:
+
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-template-id-version.tsx[tags=snippet,indent=0,group=React]
```
jouni, line 163: We could remove all of these, to reduce clutter. I don’t know, but I’m not expecting us to implement Lit (or Flow) versions of these examples in the near future.

```suggestion
include::{root}/frontend/demo/component/auto-form/react/auto-form-template-id-version.tsx[tags=snippet,indent=0]
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,214 @@
+---
+title: Auto Form
+order: 10
```
tarekoraby, line 3: I suggest putting AutoGrid before AutoForm as the former is more useful for more people.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,214 @@
...
+description: Used for auto populating form fields and actions based on a Java backend service.
+---
+
+:experimental:
+
+= [since:dev.hilla:hilla@v2.4]#Auto Form#
+
+// tag::description[]
+Auto Form is a component for populating form fields and actions based on a Java backend service.
```
tarekoraby, line 12:
```suggestion
Auto Form is a component for automatically populating form fields and actions based on a Java backend service.
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,214 @@
...
+[source,java]
+.Product.java
+----
+include::{root}/src/main/java/com/vaadin/demo/fusion/crud/form/Employee.java[tags=snippet,indent=0]
+----
+[source,java]
+.ProductRepository.java
+----
+include::{root}/src/main/java/com/vaadin/demo/fusion/crud/form/EmployeeRepository.java[tags=snippet,indent=0]
+----
```
tarekoraby, line 38:
```suggestion
.EmployeeService.java
----
include::{root}/src/main/java/com/vaadin/demo/fusion/crud/form/EmployeeService.java[tags=snippet,indent=0]
----
[source,java]
.Employee.java
----
include::{root}/src/main/java/com/vaadin/demo/fusion/crud/form/Employee.java[tags=snippet,indent=0]
----
[source,java]
.EmployeeRepository.java
----
include::{root}/src/main/java/com/vaadin/demo/fusion/crud/form/EmployeeRepository.java[tags=snippet,indent=0]
----
```

articles/react/components/auto-form/index.asciidoc
sissbruecker, line 1: Some things that are missing IMO:
- How to edit an existing item
- How to distinguish between creating a new item vs. editing an existing item
- How to get notified after an item was saved

Maybe not in the scope of this PR, but we have also added an option for showing a delete button, for which we would need to document:
- How to enable it
- How to get notified after the item has been deleted

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,241 @@
...
+include::{root}/frontend/demo/component/auto-form/react/auto-form-custom-renderer.tsx[render,frame,tags=snippet,indent=0]
+----
+--
+
+== Customizing Field's Properties
+
+You can customize the properties of the fields rendered by Auto Form by using the `fieldOptions` property.
+
+=== Customizing Visibility & Order
```
sissbruecker, line 172: This feels like a duplicate to compared to customizing the layout. It also feels like it's not in the right place as the parent section is about customizing individual fields, while customizing order and whether something shows up at all is arguably about the layout.

frontend/demo/component/auto-form/react/auto-form-custom-renderer.tsx
```diff
@@ -0,0 +1,30 @@
...
+} from '@hilla/react-crud';
+// tag::apply-backend[]
+import { EmployeeService } from 'Frontend/generated/endpoints';
+
+import EmployeeModel from 'Frontend/generated/com/vaadin/demo/fusion/crud/form/EmployeeModel';
+import { VerticalLayout } from '@hilla/react-components/VerticalLayout.js';
+
+function VerticalLayoutRenderer({ children, form }: AutoFormLayoutRendererProps<EmployeeModel>) {
+  return <VerticalLayout>{children}</VerticalLayout>;
+}
```
sissbruecker, line 16: This isn't a very helpful example IMO, because at this stage you probably want a more complex / nested layout where just dumping all the auto-generated fields as children into a container isn't enough. In general I can't see a use-case where you would want to make use of `children`.

I would suggest to expand this to render an actual custom layout, maybe using nested vertical and horizontal layouts, and then demonstrating how to make use of the `form` to bind individual fields.

sissbruecker (commented):
Added some high-level comments, still need to take a closer look at individual examples.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,241 @@
...
+
+== Customizing the Form's Layout
+
+As you can see in the example above, the Auto Form component by default renders the fields in a Form Layout with a two-column layout. You can customize the layout by using the `customLayoutRenderer` property, in either of the following ways:
+
+=== Defining a Template
+
+You can define a template for arranging the fields in the form's layout in two ways:
+
+- Defining the template as `string[][]`:
```
tarekoraby, line 74:
```suggestion
- Defining the template as `string[][]`:

```
This will show the example below fine.

frontend/demo/component/auto-form/react/auto-form-custom-renderer.tsx
taefi: Well, I must admit that at the time of writing this example, I wasn't quite sure that should we actually bother providing a sophisticated layouting example here, or should we just simply demonstrate the needed configurations for providing a custom `layoutRenderer` component?

The fields that are currently named as `children` should definitely be renamed to something else. 

The provided example should clarify how the `form` can be used to bind extra/custom components, and demonstrates that the passed in fields (AutoFormField instances that are currently named horribly as `children`) are already bound to the form and no need to do it again.

articles/react/components/auto-form/index.asciidoc
taefi: Done.

frontend/demo/component/auto-form/react/auto-form-custom-renderer.tsx
sissbruecker: I was mostly thinking about:
- showing how to make use of `form` to connect custom rendered fields to the form
- conveying that you can build your own nested layout structure here

Reusing the pre-rendered fields seems less valuable? You would need to have some kind of layout component that can do something sensible with them. Also not sure how to combine both of these into one example, but maybe you have an idea? 

In any case I think having an example for using `form` is more valuable, I think we could skip an example of using the pre-rendered fields if we don't have a good use-case.

sissbruecker: One use-case for a custom layout could be to split the form into sections, and then add headers between them. Something like:

```tsx
<h3>Personal information</h3>
<FormLayout>
  <TextField label="First name"/>
  <TextField label="Last name"/>
  ...
</FormLayout>

<h3>Employment information</h3>
<FormLayout>
  <TextField label="Team"/>
  <TextField label="Manager"/>
  ...
</FormLayout>
```

taefi: I like the semantics and simplicity of the suggested example. 

However, not sure what is the best way (or even is there a clean way) of utilizing the already-bound-to-form fields to avoid recreating and rebinding them here again. Maybe, if the developer has them as a mapping between of (field name => AutoFormField instance), then they can easily look them up based on the name, and add them in the desired order to the underlying layout e.g. `FormLayout` component.

taefi: Does it make sense to prepare such mapping for the fields and pass that instead to the `layoutRenderer` callback?

sissbruecker: Could be an option. But yeah I think that was basically my concern as well as to why the current `children` prop isn't too useful. I think there would also be alternatives to passing pre-rendered children, so maybe we should revisit that API later.

Again, I think it would make sense to focus on having an example that does everything custom, and maybe skip making use of `children` for now. We can also discuss removing that from the API for now if it bothers someone.

taefi: Well, not exactly sure how doing everything custom can help marketing the component as a productivity enhancement toolkit, but let's create an example to see what is feedback.

articles/react/components/auto-form/index.asciidoc
taefi: These are valid points, however, I couldn't come up with simple enough examples for some of the mentioned cases:

- How to edit an existing item: editing an item is involving an outer contexts (e.g. AutoCrud) that feeds an item to the form in order to edit it. 
- How to distinguish between creating a new item vs. editing an existing item: same argument

sissbruecker: For example:
```tsx
function Example() {
  const [editedItem, setEditedItem] = useState<Employee | null>(null);

  const handleEdit = () => {
    setEditedItem({
      firstName: 'Jennifer',
      lastName: 'Smith',
      gender: Gender.NON_BINARY,
      startDate: '2014-08-05',
      active: true,
      description: 'Customer Service Manager',
    });
  };

  const handleCreate = () => {
    setEditedItem(null);
  };

  const handleSubmit = ({ item }: { item: Employee }) => {
    const json = JSON.stringify(item);
    Notification.show('Form was submitted: ' + json);
  };

  return (
    <VerticalLayout>
      <HorizontalLayout>
        <Button onClick={handleEdit}>Edit item</Button>
        <Button onClick={handleCreate}>Create item</Button>
      </HorizontalLayout>
      <AutoForm
        service={EmployeeService}
        model={EmployeeModel}
        item={editedItem}
        afterSubmit={handleSubmit}
      />
    </VerticalLayout>
  );
}
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
+---
+title: Auto Form
+order: 15
+description: Used for auto populating form fields and actions based on a Java backend service.
```
sissbruecker, line 4:
```suggestion
description: Auto Form is a component that automatically generates a form for editing, updating and deleting items from a backend service.
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+---
+
+:experimental:
+
+= [since:dev.hilla:hilla@v2.4]#Auto Form#
+
+// tag::description[]
+Auto Form is a component for automatically populating form fields and actions based on a Java backend service.
```
sissbruecker, line 12:
```suggestion
Auto Form is a component that automatically generates a form for editing,
updating and deleting items from a backend service.
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+[source,java]
+.EmployeeRepository.java
+----
+include::{root}/src/main/java/com/vaadin/demo/fusion/crud/EmployeeRepository.java[tags=snippet,indent=0]
+----
+--
+
+Hilla generates TypeScript objects for each `@BrowserCallable` service implementing the `CrudService<T, ID>` interface.These TypeScript objects have callable methods that execute the corresponding Java service methods, enabling the Add, Update, and Delete operations as well as lazy data loading with sorting and filtering.
+
+For the `EmployeeService` example, you can import the generated TypeScript object in your React view, and use it as a value for the `<AutoForm service={EmployeeService} />` property to display the fields.
```
sissbruecker, line 43:
```suggestion
For the `EmployeeService` example, you can import the generated TypeScript object in your React view, and use it as a value for the `<AutoForm service={EmployeeService} />` property to generate the form.
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+[.example.desktop]
+--
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-basic.tsx[render,frame,tags=snippet,indent=0]
+----
+--
+
+As you can see in the above example, Auto Form component renders a form with input components for all properties of the `Employee` entity along with a Save button that associates with the respective `save` method of the `CrudService<T, ID>`, and an initially hidden Discard button that calls the form binder's `reset` operation, and it is only visible when the form has some changes.
```
sissbruecker, line 55:
```suggestion
As you can see in the above example, Auto Form component renders a form with fields for all properties of the `Employee` entity. When submitting the form, the data is automatically submitted to the `save` method of the configured `CrudService`. When an existing entity is edited the form also shows a delete button which calls the `delete` method of the configured `CrudService`.
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+.Default Responsive Breakpoints
+[NOTE]
+When the Auto Form component is rendered based on its default settings similar to the example above, it uses the <<../form-layout#,Form Layout component>>'s  default responsive breakpoints. If you want to customize the responsive breakpoints, you can see how it is done in the <<customizing-form-layout,Customizing Form Layout>> section.
```
sissbruecker, line 59:
```suggestion
Auto Form uses the <<../form-layout#,Form Layout component>>'s  default responsive breakpoints. To customize them, see the <<customizing-form-layout,Customizing Form Layout>> section.
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+== Customizing Visibility & Order
+
+As you can see in the example above, the Auto Form component by default renders input components for all the data properties in a Form Layout with a two-column layout, except for the fields annotated by `@Id` and `@Version`. You can customize the visibility and order of the fields using `visibleFields` property:
```
sissbruecker, line 63:
```suggestion
Auto Form by default renders input fields for all properties of the entity in a Form Layout with a two-column layout, except for the properties annotated by `@Id` and `@Version`. You can customize the visibility and order of fields using `visibleFields` property:
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+[.example.desktop]
+--
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-field-visibility.tsx[render,frame,tags=snippet,indent=0]
+----
+--
+
+As you can see in the above example, the `visibleFields` property is used to define the fields that should be displayed in the form as well as their order of appearance.
```
sissbruecker, line 73: This seems unnecessary as the text above already explains what it does.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+.Don't Expose Sensitive Data on Client-Side
+[CAUTION]
+When using Auto Form, it's important to check that you don't expose sensitive data to the client-side. For example, if you have a `password` property in your entity, hiding it using either of the `template` versions won't prevent the data from being sent to the client-side. Make sure your service implementation doesn't expose sensitive data to the client-side. You could do this, for example, by using a `@JsonIgnore` annotation on the property.
```
sissbruecker, line 77: This still refers to `template`.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+--
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-field-custom-label.tsx[render,frame,tags=snippet,indent=0]
+----
+--
+
+The above example shows how to customize the label of the `firstName` and `lastName` fields.
+
+==== Customizing Field's ColSpan
```
sissbruecker, line 101: The heading structure here is a bit excessive. I would suggest to explain the two uses cases for why you would want to customize the options in the text under `Customizing Field's Properties`, and then have a single example that demonstrates both of those.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+[.example.desktop]
+--
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-field-colspan.tsx[render,frame,tags=snippet,indent=0]
+----
+--
+
+==== Customizing Input Component
```
sissbruecker, line 113: Doesn't necessarily need a new header IMO, could just follow after the previous example.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+[.example.desktop]
+--
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-field-custom-renderer.tsx[render,frame,tags=snippet,indent=0]
+----
+--
+
+The above example shows how to render a Text Area component for the description field.
```
sissbruecker, line 125: We usually put description that the example does above the example.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+=== Customizing Form Layout
```
sissbruecker, line 127:
```suggestion
== Customizing Form Layout
```
Looks like this would otherwise end up within `Customizing Field's Properties`

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+--
+
+== Using Custom Renderer
+
+In case you want to render field in completely different layout, You can use a custom renderer to define any arbitrary layout using the `layoutRenderer` property.
+
+To use a custom renderer, define a function that takes the `AutoFormLayoutRendererProps` as a parameter and returns any React element. The `AutoFormLayoutRendererProps` contains the following properties:
+
+- children: the read-only list of bound fields
+- form: the form binder instance
```
sissbruecker, line 146:
```suggestion
- `children`: the read-only list of bound fields
- `form`: the form binder instance
```

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,166 @@
...
+
+[.example.desktop]
+--
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-custom-renderer-alt.tsx[render,frame,tags=snippet,indent=0]
+----
+--
+
+Alternatively, you can create input fields as needed, and use the `form` binder instance to take control over the bindings. The following example uses a custom renderer and manually does the form binding for the fields:
```
sissbruecker, line 158:
```suggestion
Alternatively, you can create input fields manually, and use the `form` instance to connect them to the form. 

The following example shows how to render fields manually:
```

frontend/demo/component/auto-crud/react/auto-crud-basic.tsx
```diff
@@ -0,0 +1,16 @@
+import { reactExample } from 'Frontend/demo/react-example'; // hidden-source-line
+import React from 'react'; // hidden-source-line
+import { autoGridHostStyles } from 'Frontend/demo/component/auto-grid/react/auto-grid-host-styles'; // hidden-source-line
+// tag::snippet[]
+import { ExperimentalAutoCrud as AutoCrud } from '@hilla/react-crud';
+// tag::apply-backend[]
```
sissbruecker, line 6: These tags seem unused. Let's remove them from all examples.

frontend/demo/component/auto-crud/react/auto-crud-basic.tsx
```diff
@@ -0,0 +1,16 @@
+import { reactExample } from 'Frontend/demo/react-example'; // hidden-source-line
+import React from 'react'; // hidden-source-line
+import { autoGridHostStyles } from 'Frontend/demo/component/auto-grid/react/auto-grid-host-styles'; // hidden-source-line
+// tag::snippet[]
+import { ExperimentalAutoCrud as AutoCrud } from '@hilla/react-crud';
```
sissbruecker, line 5: The other component examples don't show imports in the snippet, so I would suggest to remove them from the snippet in the new examples as well.

frontend/demo/component/auto-form/react/auto-form-custom-renderer-alt.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <HorizontalLayout theme="spacing padding" style={{ alignItems: 'end' }}>
+        {fieldsMapping.get('startDate')}
+        {fieldsMapping.get('shiftStartsAt')}
+        {fieldsMapping.get('active')}
+      </HorizontalLayout>
+      <h4>Other:</h4>
+      <HorizontalLayout theme="spacing padding">
+        {fieldsMapping.get('description')}
+      </HorizontalLayout>
+    </VerticalLayout>
```
sissbruecker, line 37:
```suggestion
    <VerticalLayout>
      <h4>Personal Information:</h4>
      <HorizontalLayout theme="spacing" className="pb-l">
        {fieldsMapping.get('firstName')}
        {fieldsMapping.get('lastName')}
        {fieldsMapping.get('gender')}
      </HorizontalLayout>
      <h4>Employment Information:</h4>
      <HorizontalLayout theme="spacing" className="pb-l items-baseline">
        {fieldsMapping.get('startDate')}
        {fieldsMapping.get('shiftStartsAt')}
        {fieldsMapping.get('active')}
      </HorizontalLayout>
      <h4>Other:</h4>
      <HorizontalLayout theme="spacing">
        {fieldsMapping.get('description')}
      </HorizontalLayout>
    </VerticalLayout>
```

Fixes some inconsistencies in alignment and spacing.

frontend/demo/component/auto-form/react/auto-form-custom-renderer.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <HorizontalLayout theme="spacing padding" style={{ alignItems: 'end' }}>
+        <DatePicker label="Start Date" {...field(model.startDate)} />
+        <TimePicker label="Shift Starts At" {...field(model.shiftStartsAt)} />
+        <Checkbox label="Active" {...field(model.active)} />
+      </HorizontalLayout>
+      <h4>Other:</h4>
+      <HorizontalLayout theme="spacing padding">
+        <TextArea label="Description" {...field(model.description)} style={{ flexGrow: 1 }} />
+      </HorizontalLayout>
+    </VerticalLayout>
```
sissbruecker, line 57:
```suggestion
    <VerticalLayout>
      <h4>Personal Information:</h4>
      <HorizontalLayout theme="spacing" className="pb-l">
        <TextField label="First Name" {...field(model.firstName)} />
        <TextField label="Last Name" {...field(model.lastName)} />
        <Select
          label="Gender"
          {...field(model.gender)}
          items={[
            { label: 'Female', value: Gender.FEMALE },
            { label: 'Male', value: Gender.MALE },
            { label: 'Non-Binary', value: Gender.NON_BINARY },
            { label: 'Prefer not to say', value: Gender.OTHER },
          ]}
        />
      </HorizontalLayout>
      <h4>Employment Information:</h4>
      <HorizontalLayout theme="spacing" className="pb-l items-baseline">
        <DatePicker label="Start Date" {...field(model.startDate)} />
        <TimePicker label="Shift Starts At" {...field(model.shiftStartsAt)} />
        <Checkbox label="Active" {...field(model.active)} />
      </HorizontalLayout>
      <h4>Other:</h4>
      <HorizontalLayout theme="spacing">
        <TextArea label="Description" {...field(model.description)} style={{ flexGrow: 1 }} />
      </HorizontalLayout>
    </VerticalLayout>
```

articles/react/components/auto-crud/index.asciidoc
```diff
@@ -0,0 +1,141 @@
...
+[source,java]
+.EmployeeRepository.java
+----
+include::{root}/src/main/java/com/vaadin/demo/fusion/crud/EmployeeRepository.java[tags=snippet,indent=0]
+----
+--
+
+Hilla generates TypeScript objects for each `@BrowserCallable` service implementing the `CrudService<T, ID>` interface. These TypeScript objects have callable methods that execute the corresponding Java service methods, enabling the Add, Update, and Delete operations as well as lazy data loading with sorting and filtering.
+
+For the `EmployeeService` example, you can import the generated TypeScript object in your React view, and use it as a value for the `<AutoCrud service={}/>` property to auto-populate grid columns and form fields:
```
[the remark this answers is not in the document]
taefi: I agree to the suggestion for replacing the `auto-` with `automatically. 
However, I'm against using the word `generate` in the context of Auto XXX components, as it is actually populating the at runtime, and does not generate anything in that regard. Code Generation most of the time is referring to a process happening at compile time or application startup, and refers to some codes that needed to be taken care of after generation.

articles/react/components/auto-form/index.asciidoc
taefi: The re-phrasing looks good to me and I will adopt it. Same argument for the use of the word `generates` as earlier.

articles/react/components/auto-form/index.asciidoc
taefi: Same comments here.

articles/react/components/auto-form/index.asciidoc
taefi: Same comment about `generate` as before.

articles/react/components/auto-form/index.asciidoc
taefi: The re-phrasing of the paragraph looks awesome. 
About the delete button, I'm actually covering it in a separate section (will be added soon), as the delete button only shows up when the form is in edit mode **and** it also needs the `deleteButtonVisible` to be enabled. I think announcing all of the conditions in the same sentence would make it long and complex. WDYT? (Probably after I push the rest of the doc soonish).

articles/react/components/auto-form/index.asciidoc
taefi: Well, having them as headings, helps with enabling the referral to that fragment of them docs by adding the `#some-title-heading` at the end of the url. I'd say there is trade off when referring to the exact position in the docs in stackoverflow or other Q and A websites, expert chat etc., but I don't have some strong opinions whether is this the only way to enable this in our docs or not.

articles/react/components/auto-crud/index.asciidoc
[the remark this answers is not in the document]
sissbruecker: From my perspective, I associate `populating` with filling out a form automatically, but not with creating the fields and layout of the form. For example, when populating a database, you usually mean that you insert initial data into the tables, but not that you generate the tables.

articles/react/components/auto-form/index.asciidoc
sissbruecker: Oh, you're right - let's remove the sentence about the delete button then.

frontend/demo/component/auto-crud/react/auto-crud-basic.tsx
taefi: You mean only the `// tag::apply-backend[]` tag, right?

sissbruecker: Yes

articles/react/components/auto-form/index.asciidoc
cromoteca:
```suggestion
==== Customizing Field Column Span
```
Anyway, those headers should be rephrased to remove Saxon genitive (language checking will have the final say of course) and to avoid using `colspan` as a word.

articles/react/components/auto-crud/index.asciidoc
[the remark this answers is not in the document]
taefi: Fair enough. Let's pick something other than `generating` and `populating` then :) 
What about `rendering` or `displaying`? The word `creating` also comes to mind, but not my first choice maybe.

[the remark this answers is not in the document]
sissbruecker: `render` works fine in this context 👍

articles/react/components/auto-form/index.asciidoc
taefi: @cromoteca: Applied manually.

frontend/demo/component/auto-crud/react/auto-crud-basic.tsx
taefi: Done.

articles/react/components/auto-crud/index.asciidoc
```diff
@@ -0,0 +1,141 @@
...
+Auto-Crud component is a high-level component and composed of the following components:
+
+- <<../auto-grid/index.asciidoc#,Auto Grid>>
+- <<../auto-form/index.asciidoc#,Auto Form>>
+
+Each of the above components offer some customization options, which are described in the following sections.
+
+== Customizing the Grid
+
+As you can see in the example above, the Auto-Crud component by default renders the grid with all the properties of the entity. You can customize the grid by using the `gridProps` property.
```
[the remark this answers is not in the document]
taefi: My understanding from the discussion with @tarekoraby about how much AutoCrud documentation should cover of the underlaying components was that not to just refer to those documentations, but to have some examples, so that the user don't have to jump between the component articles. There is a high chance that I have overdone it :)
WDYT @tarekoraby?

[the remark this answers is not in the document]
tarekoraby: That was a thought that we discussed, yes. But I have no strong opinion. It's your call, @taefi.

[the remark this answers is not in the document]
taefi: @sissbruecker done.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,229 @@
...
+== Using Custom Renderer
+
+To render a field in a completely different layout, you can use a custom renderer to define any arbitrary layout using the `layoutRenderer` property.
+
+To use a custom renderer, define a function that takes the `AutoFormLayoutRendererProps` as a parameter and returns any React element. The `AutoFormLayoutRendererProps` contains the following properties:
+
+- `children`: the read-only list of bound fields
+- `form`: the form binder instance
+
+As the children property is a read-only list of bound fields, you can use them right away in your custom renderer. The following example shows how to use the `children` property to render the fields in a two-column layout:
```
sissbruecker, line 154: > to render the fields in a two-column layout

This isn't quite what is does.

articles/react/components/auto-form/index.asciidoc
```diff
@@ -0,0 +1,229 @@
...
+- `onSubmitError`: a callback that is called if an unexpected error occurs while submitting the form. The callback receives the error as a parameter.
+- `onDeleteError`: a callback that is called if an unexpected error occurs while deleting the item. The callback receives the error as a parameter.
+
+The following example shows how to use the `onSubmitError` and `onDeleteError` callbacks to inform the user about the respective error:
+
+[.example.desktop]
+--
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-form/react/auto-form-on-errors.tsx[render,frame,tags=snippet,indent=0]
```
sissbruecker, line 223: Let's just use an inline code snippet here rather than a live demo. Then we also don't need to come up with ways to generate an error.

taefi: Done.

articles/react/components/auto-form/index.asciidoc
taefi: Fixed.

articles/react/components/auto-crud/index.asciidoc
```diff
@@ -0,0 +1,137 @@
...
+----
+--
+
+Hilla generates TypeScript objects for each `@BrowserCallable` service implementing the `CrudService<T, ID>` interface. These TypeScript objects have callable methods that execute the corresponding Java service methods, enabling the Add, Update, and Delete operations as well as lazy data loading with sorting and filtering.
+
+For the `EmployeeService` example, you can import the generated TypeScript object in your React view, and use it as a value for the `<AutoCRUD service={}/>` property to render automatically grid columns and form fields:
+
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-crud/react/auto-crud-basic.tsx[tags=snippet,indent=0]
```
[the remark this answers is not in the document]
taefi: Done.

sissbruecker (approved):

PR #2933, opened 2023-11-01
Make TS example consistent with Java code

tarekoraby:


frontend/demo/auth/authRoles.ts
```diff
@@ -0,0 +1,14 @@
+import { configureAuth } from '@hilla/react-auth';
+import { UserInfoService } from 'Frontend/generated/endpoints';
+
+// tag::snippet[]
+// Configure auth to use `UserInfoService.getUserInfo` and map to custom roles
+const auth = configureAuth(UserInfoService.getUserInfo, {
+  getRoles: (userInfo) => userInfo.authorities,
```
tarekoraby, line 7: @cromoteca, VS Code is complaining that `'userInfo' is of type 'unknown'`. Is that just the docs project, or should we do something like 
```
import { UserInfo } from 'Frontend/generated/endpoints';

...

getRoles: (userInfo: UserInfo) => userInfo.authorities as readonly string[],
```

cromoteca: The object is correctly typed:
image

articles/lit/guides/security/spring-login.adoc
```diff
@@ -307,14 +307,12 @@ Hilla supports role-based security out-of-the-box, provided that the user object
 }
 ----
 
-If roles are exposed differently in your user object, you can still tell Hilla how to find them by amending your authentication configuration on the client:
+If roles are exposed differently in your user object, you can still tell Hilla how to find them by amending your authentication configuration on the client. In the [classname]`UserInfo` example provided above, the roles are returned by [methodname]`getAuthorities`, so the [filename]`auth.ts` file should be modified as such:
```
jouni, line 310: Which example? 

This one: https://github.com/vaadin/docs/pull/2933/files#diff-410868dea60015bdca8ffcc0846b887ed2f258a8b3aba85e47e36024269ad2eaR120

Or this one: https://github.com/vaadin/docs/pull/2933/files#diff-410868dea60015bdca8ffcc0846b887ed2f258a8b3aba85e47e36024269ad2eaR245

We should add a in-page reference link to the one that we mean.

```suggestion
If roles are exposed differently in your user object, you can still tell Hilla how to find them by amending your authentication configuration on the client. In the <<TODO-ADD-ID,[classname]`UserInfo` example>> provided before, the roles are returned by [methodname]`getAuthorities`, so the [filename]`auth.ts` file should be modified as such:
```

tarekoraby: Ping @cromoteca

tarekoraby:
Consider re-targeting this PR to branch next, or remember to cherry-pick to next.

PR #2934, opened 2023-11-01
fix tooltip link in component overview

martinjunker:
Tooltip Component is linked to spreadsheet instead of tooltip details. had no test so please check if new link is correct. KR Martin


jouni (approved):
Thanks for noticing and fixing this!

PR #2970, opened 2023-11-15
docs: address DX test feedback for Hilla 2.4

sissbruecker:
Addresses some smaller issues that came up in DX tests.

### DIFF
--- a/articles/react/components/auto-crud/index.asciidoc
+++ b/articles/react/components/auto-crud/index.asciidoc
@@ -40,7 +40,14 @@ include::{root}/src/main/java/com/vaadin/demo/fusion/crud/EmployeeRepository.jav
 
 Hilla generates TypeScript objects for each `@BrowserCallable` service implementing the `CrudService<T, ID>` interface. These TypeScript objects have callable methods that execute the corresponding Java service methods, enabling the Add, Update, and Delete operations as well as lazy data loading with sorting and filtering.
 
-For the `EmployeeService` example, you can import the generated TypeScript object in your React view, and use it as a value for the `service` property to render automatically grid columns and form fields. Here's what the rendered component looks like:
+For the example above, you can import the generated TypeScript objects in your React view and configure the component like so:
+
+[source,tsx]
+----
+<AutoCrud service={EmployeeService} model={EmployeeModel} />
+----
+
+Here's what the rendered component looks like:
 
 [.example.desktop]
 --
--- a/articles/react/components/auto-form/index.asciidoc
+++ b/articles/react/components/auto-form/index.asciidoc
@@ -180,9 +180,9 @@ The following properties are available for managing the form state or reacting t
 
 - `item`: the item to be edited. If the item is `null`, the form is in _new_ mode, otherwise it is in _edit_ mode.
 - `disabled`: disables the whole form.
-- `afterSubmit`: a callback function that is called after the form is submitted successfully.
+- `onSubmitSuccess`: a callback function that is called after the form is submitted successfully.
 
-The following example shows how to use the `item` property to manage switching between _new_ and _edit_ modes, and `afterSubmit` for informing the user about the form submission success. In addition, it uses `disabled` property is also used to toggle the form's disabled state:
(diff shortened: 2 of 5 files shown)


articles/react/components/auto-form/index.asciidoc
```diff
@@ -40,7 +40,12 @@ include::{root}/src/main/java/com/vaadin/demo/fusion/crud/EmployeeRepository.jav
 
 Hilla generates TypeScript objects for each `@BrowserCallable` service implementing the `CrudService<T, ID>` interface. These TypeScript objects have callable methods that execute the corresponding Java service methods, enabling the Add, Update, and Delete operations as well as lazy data loading with sorting and filtering.
 
-For the `EmployeeService` example, you can import the generated TypeScript object in your React view, and use it as a value for the `<AutoForm service={EmployeeService} />` property to render the form.
```
sissbruecker, line 43: Some testers referred to the example code in this paragraph when creating the initial component, and also missed to expand the code snippet in the example below. They were then confused why they ended up with a partially configured component and had compiler errors. Changed this in both articles to show a proper code snippet with a fully configured component.

articles/react/guides/routing.adoc
```diff
@@ -115,7 +115,7 @@ export default function ProductDetailView() {
   const { productId } = useParams();
 
   // Fetch product details for the given ID
-  const { product, setProduct } = useState<Product>();
+  const [ product, setProduct ] = useState<Product>();
```
sissbruecker, line 118: Only semi-related, but discovered this syntax issue as part of the tests.

frontend/demo/component/auto-crud/react/auto-crud-customized-form.tsx
```diff
@@ -17,9 +17,7 @@ function Example() {
         fieldOptions: {
           firstName: { label: 'Name' },
           description: {
-            renderer: ({ field }) => (
-              <TextArea key={field.name} {...field} label="Full description" />
-            ),
+            renderer: ({ field }) => <TextArea {...field} label="Full description" />,
```
sissbruecker, line 20: The `key` property caused some confusion in whether it was related to mapping the renderer to the property, or what its general purpose was and whether it is needed. Turns out it's not needed as the renderer output is wrapped into an `AutoFormField` with a proper key.

articles/react/components/auto-crud/index.asciidoc
```diff
@@ -40,7 +40,14 @@ include::{root}/src/main/java/com/vaadin/demo/fusion/crud/EmployeeRepository.jav
...
 
-For the `EmployeeService` example, you can import the generated TypeScript object in your React view, and use it as a value for the `service` property to render automatically grid columns and form fields. Here's what the rendered component looks like:
+For the example above, you can import the generated TypeScript objects in your React view and configure the component like so:
+
+[source,tsx]
+----
+<AutoCrud service={EmployeeService} model={EmployeeModel} />
+----
+
+Here's what the rendered component looks like:
```
taefi, line 50: Isn't this something we discussed to remove on the original PR?

sissbruecker: What specifically? Currently one document shows a non-working code snippet, the other document only references the `service` prop, which also isn't enough. So I thought this would be a solution for both.

frontend/demo/component/auto-crud/react/auto-crud-customized-form.tsx
taefi: Have we changed something about this recently? As at the time of writing the doc, React renderer was complaining about every element should have a key ...

sissbruecker: Maybe. I did check with the auto form unit tests and didn't have any issues there when removing the key. It also makes sense, as the auto form renders an array of `AutoFormField`s, which have the key set: https://github.com/vaadin/hilla/blob/d5ae054bcbeb2de9d4bdddadd1c947aa77fbfc33/packages/ts/react-crud/src/autoform.tsx#L281-L288
`AutoFormField` calls the renderer internally, no key is needed there as the renderer result in not put into an array.

taefi (approved):

PR #2974, opened 2023-11-20
fix: set id for new items in CrudService mock

taefi:


frontend/demo/services/CrudService.ts
```diff
@@ -123,7 +123,7 @@ export class CrudMockService<T extends AbstractEntity> implements CrudService<T>
     if (existingIndex >= 0) {
       this.items[existingIndex] = item;
     } else {
-      this.items.push(item);
+      this.items.push({ ...item, id: this.items.length + 1 });
```
cromoteca, line 126: This will give wrong new IDs if items are deleted.
```suggestion
      this.items.push({ ...item, id: this.items.map(item => item.id).reduce((a, b) => Math.max(a, b), 0) + 1 });
```
I wrote this inline here, so check it before applying.

cromoteca (approved):

PR #2984, opened 2023-11-21
docs: removed unnecessary 'before 24.3 note'

tltv:
24.3 is improved to not change the previous behavior. default `RouteAccessDeniedError` reroutes to `RouteNotFoundError` now, it doesn't extend `AbstractRouteNotFoundError`. No need to mention anything here anymore.

Related-to: vaadin/flow#18082


articles/security/enabling-security.adoc
```diff
@@ -377,8 +377,6 @@ If the user is already authenticated and tries to navigate to a view for which t
 
 In development mode, Vaadin shows the [classname]`RouteAccessDeniedError` view, which shows an _Access denied_ message with a list of available routes. In production mode, Vaadin shows the [classname]`RouteAccessDeniedError` view, which shows a _Could not navigate to 'RequestedRouteName'_ message by default. For security reasons, the message doesn't say whether the navigation target exists.
```
mcollovati, line 378: > shows the [classname]`RouteAccessDeniedError` view

Should we mention that this is not really a view but a reroute to the RouteNotFoundError?

tltv: Mentioning that it is indeed "rerouting to `RouteNotFoundError` by default for backwards compatibility reason" would make sense.

mcollovati (approved):

Issue #2990, opened 2023-11-27
Hilla: [autogrid] Document order support for custom columns in Hilla 2.5+


taefi:
Document the way user should utilize the `key` property of custom columns in the `visibleColumns` to be able to define the column order for them as well.

This feature became available from 2.5 pre-releases. A _version label_ in the docs can help preventing potential confusions about which versions are supporting it.

sissbruecker:
Has been resolved with https://github.com/vaadin/docs/pull/3022

PR #2991, opened 2023-11-27
feat: [autogrid] add section for ordering the custom columns

taefi:
fixes: #2990
Also adds a section about the exposed refresh function to be called from parent component.


articles/react/components/auto-grid/index.asciidoc
```diff
@@ -120,6 +120,19 @@ include::{root}/frontend/demo/component/auto-grid/react/auto-grid-custom-columns
 [NOTE]
 In general, when it comes to customizing the automatically rendered columns, you can use the `columnOptions` property. As for adding or customizing the custom columns, use `customColumns`.
 
+==== [since:dev.hilla:hilla@v2.5]#Custom Columns Visibility & Order#
+
+You can define the order of appearance for custom columns also using the `visibleColumns` property. This means, custom columns can be freely placed before, in between, or after other columns. To achieve this, you need to provide the `key` property of the custom column in the `visibleColumns` property. The following example shows a custom column with _"supplierInfo"_ as the `key` placed between two automatically rendered columns:
```
sissbruecker, line 125:
```suggestion
By default, custom columns are added after the automatically rendered columns. You can define the order of appearance for custom columns also using the `visibleColumns` property. This means, custom columns can be freely placed before, in between, or after other columns. To achieve this, you need to provide the `key` property of the custom column in the `visibleColumns` property. The following example shows a custom column with _"supplierInfo"_ as the `key` placed between two automatically rendered columns:
```
Would suggest to note about the default behavior at the start, and then explain about customizing it.

articles/react/components/auto-grid/index.asciidoc
```diff
@@ -120,6 +120,19 @@ include::{root}/frontend/demo/component/auto-grid/react/auto-grid-custom-columns
 [NOTE]
 In general, when it comes to customizing the automatically rendered columns, you can use the `columnOptions` property. As for adding or customizing the custom columns, use `customColumns`.
 
+==== [since:dev.hilla:hilla@v2.5]#Custom Columns Visibility & Order#
```
sissbruecker, line 123: I would suggest to keep it simple and only mention order. The note about the custom columns visibility is enough to note about the specific behavior when using `visibleColumns`.

articles/react/components/auto-grid/index.asciidoc
```diff
@@ -120,6 +120,19 @@ include::{root}/frontend/demo/component/auto-grid/react/auto-grid-custom-columns
...
+
+.Auto Grid Custom Column's Visibility & Order
+[source,tsx]
+----
+include::{root}/frontend/demo/component/auto-grid/react/auto-grid-custom-columns-order.tsx[render,tags=snippet,indent=0,group=React]
+----
+
+.Custom Columns Visibility
+[NOTE]
+If the `visibleColumns` property is *not* set, by default, all the custom columns are rendered after the automatically rendered columns. Otherwise, only the custom columns that their keys are mentioned in the `visibleColumns` are rendered according to the specified order.
```
sissbruecker, line 135: With the added sentence above it would suffice to explain here that if you use `visibleColumns` you must mention the custom column keys, otherwise they won't show up.

sissbruecker:
Closing in favor of https://github.com/vaadin/docs/pull/3079

PR #2993, opened 2023-11-28
fix: Improve note about AppShellConfiguration

TatuLund:


articles/advanced/modifying-the-bootstrap-page.asciidoc
```diff
@@ -128,6 +128,7 @@ In Java code, when adding dynamic content during the bootstrap process, use the
 
 [NOTE]
 There must be a single application shell for the entire Vaadin application, so that no more than one class can implement [interfacename]`AppShellConfigurator`.
+In the case of a multimodule application, the class implementing the AppShellConfigurator interface needs to be in the main module.
```
mshabarov, line 131: A bit unclear what "main module" means here? Is it a module that being built into jar/war to be deployed on a server, and not a module that is added as a dependency to that "main" module?

PR #2956, opened 2023-11-10
docs: update articles for access denied error feature

tltv:
Updates exceptions.asciidoc example to be not mixed with build-in AccessDeniedException. Updates enabling-security.adoc to document how to customize access denied error page and how to use new AccessDeniedErrorRouter annotation to reroute to different error page for unauthorized route.

Related-to: vaadin/flow#11335

PR #3003, opened 2023-11-29
fix: Note about access control annotation for HasUrlParameter

mshabarov:
Adds an access control annotation to the example code and a note that this is needed to allow navigation.

Related to https://github.com/vaadin/docs/pull/2956

### DIFF
--- a/articles/security/enabling-security.adoc
+++ b/articles/security/enabling-security.adoc
@@ -411,6 +411,7 @@ public class CustomAccessDeniedError extends RouteAccessDeniedError {
 [source,java]
 ----
 @Tag(Tag.DIV)
+@AnonymousAllowed
 public static class CustomAccessDeniedError extends Component
         implements HasErrorParameter<AccessDeniedException> {
     @Override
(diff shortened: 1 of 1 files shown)


mcollovati (approved):

articles/security/enabling-security.adoc
```diff
@@ -422,6 +423,8 @@ public static class CustomAccessDeniedError extends Component
 }
 ----
 
+[interfacename]`HasErrorParameter` error view needs an access control annotation, so that Vaadin allows the navigation to it.
+Example above uses [annotationname]`@AnonymousAllowed`, but it can be also [annotationname]`@PermitAll`.
```
tarekoraby, line 427: I assume any annotation can be used (like the developer might want to show the error view to specific roles only). Is that right? 

```suggestion
The example above uses [annotationname]`@AnonymousAllowed`, but other annotations can also be used (for example, [annotationname]`@PermitAll`).

```

mshabarov: Yes, that's right. Thanks

Issue #3013, opened 2023-12-01
There should be documentation for the new rerouteToError usage in annotation


HerbertsVaadin:
There should be documentation about the usage of the new `@AccessDeniedErrorRouter(rerouteToError = CustomException.class)` in the following page:

https://vaadin.com/docs/latest/routing/exceptions#rerouting-to-error-view

Please also mention that the error view (`CustomErrorView extends Div implements HasErrorParameter<CustomException>`) should contain the `@AnonymousAllowed`, `@PermitAll` or any other relevant annotation, otherwise it will default to the default reroute error view. 

---

If you think this issue is important, add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

PR #3015, opened 2023-12-01
fix: Add a reference to rerouteToError annotation parameter

mshabarov:
Fixes https://github.com/vaadin/docs/issues/3013


articles/routing/exceptions.adoc
```diff
@@ -210,6 +211,7 @@ public class BlogPost extends Component
 }
 
 @Tag(Tag.DIV)
+@PermitAll
```
mcollovati, line 214: I think this should be `@AnonymousAllowed`. This error handler is not handling a security exception, so the error view should be public

articles/routing/exceptions.adoc
```diff
@@ -227,11 +229,14 @@ public class FaultyBlogPostHandler extends Component
 }
 ----
 
+[interfacename]`HasErrorParameter` error view needs an access control annotation, so that Vaadin allows navigation to it. The example above uses [annotationname]`@PermitAll`, but [annotationname]`@RolesAllowed` can also be used. [annotationname]`@AnonymousAllowed` isn't recommended, as it exposes information about access restrictions to the anonymous users.
+
+Re-routing to a custom access denial views is also possible with annotation [annotationname]`@AccessDeniedErrorRouter`, see <<{articles}/advanced/custom-error-handler#, Customizing Error Messages for Unauthorized Views>>.
```
mcollovati, line 234: Should this note maybe be moved to the previous paragraph?

mcollovati (changes requested):

articles/routing/exceptions.adoc
mshabarov: Agree.

mshabarov: Changed

articles/routing/exceptions.adoc
mshabarov: Yes, done.

mcollovati (approved):

PR #3022, opened 2023-12-05
feat: add custom column order and grid refresh sections

taefi:
fixes: #2990
Also adds a section about the exposed refresh function to be called from parent component.


taefi:
The paragraphs are taken from #2991, which was language checked before (but couldn't be merged due to technical issues).

sissbruecker (approved):

PR #3027, opened 2023-12-07
docs: auto form validation docs

sissbruecker:
Documents how to use validation in auto form.

Marking as draft as this needs a new Hilla 2.5 release.


articles/react/components/auto-form/index.asciidoc
```diff
@@ -230,6 +230,51 @@ include::{root}/frontend/demo/component/auto-form/react/auto-form-on-errors.tsx[
 [NOTE]
 Note that `onSubmitError` callback is not be called for form validation errors, which are handled automatically.
 
+== Validation
+
+Auto Form uses the same validation mechanism as <<{articles}/react/guides/forms/binder-validation#,default Hilla Forms>>. That means you can add Bean validation annotations to your entity classes and the form will automatically validate the input values on both client and server.
```
taefi, line 235:
```suggestion
Auto Form uses the same validation mechanism as <<{articles}/react/guides/forms/binder-validation#,default Hilla Forms>>. That means you can add Bean validation annotations to your entity classes and the form automatically validates the input values on both client and server.
```

taefi (changes requested):

PR #3028, opened 2023-12-07
docs: clarify upload article regarding setAcceptedFileType

tltv:
Clarified Upload component article to mention that Upload component doesn't restrict file format on the server side.


articles/components/upload/index.asciidoc
```diff
@@ -262,6 +262,11 @@ endif::[]
 Although https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types[MIME types] are widely supported, file extensions are only implemented https://caniuse.com/input-file-accept[in certain browsers] and should be avoided.
 ====
 
+.Java Flow-Specific
+[NOTE]
+====
+File format restrictions set with [methodname]`setAcceptedFileType` are checked on the client side. Uploaded file's format is not restricted on the server side by default. Upload doesn't have API to restrict uploaded files by file format or content in the server side. It's however possible to implement application specific restriction also on the server side in one or more event listeners when restriction is required.
+====
```
tarekoraby, line 270:
```suggestion
ifdef::flow[]
File format restrictions are client-side
[NOTE]
====
File format restrictions set with [methodname]`setAcceptedFileType`  method are checked only on the client side. Using this method will not restrict the uploaded file's format on the server side. The Upload component doesn't have an API to restrict uploaded files by file format or content on the server side. If required, it's possible to manually implement application-specific restrictions on the server side in one or more of the upload event listeners.
====
endif::[]

```

articles/components/upload/index.asciidoc
```diff
@@ -262,6 +262,11 @@ endif::[]
 Although https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types[MIME types] are widely supported, file extensions are only implemented https://caniuse.com/input-file-accept[in certain browsers] and should be avoided.
 ====
 
+.Java Flow-Specific
```
mshabarov, line 265:
```suggestion
.Server-side format checking
```

articles/components/upload/index.asciidoc
```diff
@@ -262,6 +262,11 @@ endif::[]
...
+[NOTE]
+====
+File format restrictions set with [methodname]`setAcceptedFileType` are checked on the client side. Uploaded file's format is not restricted on the server side by default. Upload doesn't have API to restrict uploaded files by file format or content in the server side. It's however possible to implement application specific restriction also on the server side in one or more event listeners when restriction is required.
```
mshabarov, line 268:
```suggestion
File format restrictions set with [methodname]`setAcceptedFileType` are checked in the browser and indicate the hints for users of what file types to upload. 

Uploaded file's format, as well as file's content, is not restricted on the server side by Vaadin. It's the responsibility of the application developer to implement application specific restriction on the server side in one or more Upload's event listeners when restriction is required, e.g. with `Upload::addSucceededListener`.
```

mshabarov (changes requested):

articles/components/upload/index.asciidoc
tltv: How does `ifdef::flow[]` work in this article's context? Can I choose somehow to only show Flow related topics or is it just hidden extra context?

articles/components/upload/index.asciidoc
mshabarov: Here is a merged suggestion :)
```
File format restrictions set with [methodname]`setAcceptedFileType` method are checked only on the client side and indicate the hints for users of what file types to upload. 

Using this method will not restrict the uploaded file's format on the server side. The Upload component doesn't have an API to restrict uploaded files by file format or content on the server side. If required, It's the responsibility of the application developer to implement application-specific restrictions on the server side in one or more of the Upload's event listeners, e.g. in `Upload::addSucceededListener`.
```

articles/components/upload/index.asciidoc
tarekoraby: To my understanding, this will ensure that this part will not be shown in [hilla.dev/](https://hilla.dev/docs/react/components/upload).

mshabarov (approved):

PR #3040, opened 2023-12-11
docs: add grid header footer styling section

### LINKED ISSUE #3033: Add header and footer partname APIs to Grid docs
Add a new section after the [PartNameGenerator section](https://vaadin.com/docs/latest/components/grid#part-name-generator) of Grid docs as follows:

---
### Styling Header and Footer cells <sup>Since V24.3</sup>

Header and footer cells can be similarly styled through their own custom part names.

`[ WC, Java and React samples with part names added to a header cell and some css applied to it ]`

ugur-vaadin:
Adds Grid header and footer styling section with WC, Java, and React examples.

Closes #3033


rolfsmeds (changes requested):
Applying the same styling to all header cells kind of makes it hard to spot unless you know what you're looking for.

I would propose instead just applying a part name to the Customer Rating column, and using it to make the font smaller and force the col caption to wrap, as it will otherwise not fit into the column, i.e.
```
font-size: var(--lumo-font-size-xs);
white-space: unset;
```

rolfsmeds (approved):

PR #3043, opened 2023-12-12
docs: add ContextMenu className usage examples

### LINKED ISSUE #3031: Add menu-item styling to Context Menu docs
Add the following section after the [Custom Items](https://vaadin.com/docs/latest/components/context-menu#custom-items) section of the Context Menu docs page:

---
### Styling Menu Items

Individual menu items can be styled by applying custom classnames to them <sub>(since 24.3)</sub>, and writing CSS style blocks targeting those classnames.

`[ WC, Java and React samples with a classname applied to a menu item ]`

> [!NOTE]
> **Use theme names instead of classnames in V24.2 and older**
> In versions prior to 24.3, theme names must be used instead (`theme` property / `addThemeNames` Java method). The CSS syntax for targeting a theme name is `[theme~="custom-theme"]`

web-padawan:
Fixes #3031

Added examples of using `className` API for `ContextMenu` items.


articles/components/context-menu/index.asciidoc
```diff
@@ -191,6 +191,39 @@ include::{root}/frontend/demo/component/contextmenu/react/context-menu-presentat
 endif::[]
 --
 
+== Styling Menu Items
+
+Individual menu items can be styled by [since:com.vaadin:vaadin@V24.3]#applying custom classnames# to them, and writing CSS style blocks targeting those classnames.
```
rolfsmeds, line 196: Apparently we've decided to use class names instead of classnames in docs, so let's go with that here as well.

(I seemed to recall a decision in the opposite direction, but now I can't find a discussion about it anywhere...)

rolfsmeds (changes requested):

rolfsmeds (changes requested):
How about, instead of Success, Warning, Error items, we would instead have a more realistic example, e.g. something like:

- Share
- Duplicate
- Delete

where the last item has a "danger" classname applied and colored red.

articles/components/context-menu/index.asciidoc
web-padawan: Changed.

web-padawan:
> we would instead have a more realistic example

Updated examples accordingly.

rolfsmeds (approved):

PR #3049, opened 2023-12-13
chore: update hilla version to 2.5.0.rc1

tomivirkki:
Commit automatic dependency bumps applied by Vaadin upon server startup.

IMO the dependency updates should be part of the automatic Vaadin [version bump PR](https://github.com/vaadin/docs/pull/3044)


ZheSun88:
I wonder why this change was not in commit (https://github.com/vaadin/docs/pull/3044/files)

does it need some extra way to generate those changes?

tomivirkki:
> does it need some extra way to generate those changes?

Just running `mvn` (to start up in dev)

ZheSun88:
this is what we do to generate the version update PR 
need to change something?
```
mvn -B -q versions:set-property -Dproperty=vaadin.version -DnewVersion=%version% -DgenerateBackupPoms=false && \
mvn -U -ntp install -Pproduction
```

tomivirkki:
That's only running the production build. Maybe starting up in dev mode updates additional dependencies. Try running `mvn`.

ZheSun88:
running `mvn` seems calling `spring-boot:run` which starts the server but it will not stop automatically. so the build will just hang there.  

seems the difference between profile `production` and default run is the `prepare-frontend` goal. 
maybe we should add it [here](https://github.com/vaadin/docs/blob/latest/pom.xml#L257).

ZheSun88:
ah.. no seems it is not working either..

ZheSun88 (approved):

PR #3050, opened 2023-12-13
fix: remove "accessor" and disable Lit examples

Lodin:
This PR temporary disables rendering Lit examples to avoid issues with standard/legacy decorators


sissbruecker:
Let's reconsider this after testing the Esbuild mode on the hilla branch a bit. If it works, let's switch the latest branch to use this mode as well, and only remove the accessor keywords.

PR #3054, opened 2023-12-13
fix: temporary disable render of components

Lodin:


sissbruecker:
Closing as we found a solution for using legacy decorators.

PR #3067, opened 2023-12-14
Update hilla version

tarekoraby:


mshabarov:
Oh, yes, that was wrong. Good catch.

PR #3018, opened 2023-12-04
chore(hilla): update to 2.5.0.alpha11

cromoteca:

PR #3072, opened 2023-12-15
chore: bump Hilla version to 2.5.1

sissbruecker:
Depends on https://github.com/vaadin/docs-app/pull/407


articles/lit/guides/security/spring-stateless.adoc
```diff
@@ -56,7 +56,7 @@ Modify the Spring Security configuration and use the [methodname]`HillaWebSecuri
 .`SecurityConfig.java`
 [source,java]
 ----
-include::{root}/src/main/java/com/vaadin/demo/fusion/security/stateless/SecurityConfig.java[tags="stateless-configure"]
+include::{root}/articles/lit/guides/security/SecurityConfig.java[tags="stateless-configure"]
```
cromoteca, line 59: Instead of moving this file, consider doing what I did in https://github.com/vaadin/docs/pull/3018: 

- remove `@ComponentScan` from `Application.java`
- replace `@ExcludeDemoSpringComponent` with `@Profile("this-is-just-a-demo-class")` in `SecurityConfig.java`
- remove `ExcludeDemoSpringComponent.java` as it is no longer used.

This basically puts the example into a Spring profile which will never be used. This can be a permanent solution: it can be seen as a trick, but it's effective and doesn't require any additional configuration. Feel free to select the solution you prefer.

cromoteca (changes requested):
Tried the application locally and everything works fine.

articles/lit/guides/security/spring-stateless.adoc
sissbruecker: Thanks, updated it to use your solution.

taefi (approved):

cromoteca (approved):

PR #3100, opened 2024-01-02
Update Gradle and Java version requirements

TatuLund:
See release notes of Vaadin 24.3


jouni:
Need to update https://vaadin.com/docs/latest/compatibility as well.

PR #3120, opened 2024-01-11
Made clear that RouteScopeOwner is a mandatory qualifier.

thigg:
We struggled a bit because we were not realizing that RouteScopeOwner is a Qualifier and thus mandatory for lookups.
Also added an example for programmatic lookup.

Feel free to change this to your needs. This is the text that would have helped us.


jouni:
Thanks for the help! We’ll take it from here 👍

mshabarov (commented):
The first part makes sense to me except "Beans that are `RouteScoped` are in the `RouteScope`". What is "RouteScope" ?

The second part misses some explanations about the use case.

Let me discuss with the team if this addition is clear enough for broad auditory of readers, because I might miss some details.

articles/integrations/cdi/contexts.asciidoc
```diff
@@ -230,6 +230,25 @@ public class ScopedView extends Div {
 
 The message is delivered to the `ScopedView` instance where the user already navigated. If on another view, there is no instance of this bean and the message isn't delivered to it.
 
+To resolve `RouteScoped` beans programmatically, you'll need to instantiate the `RouteScopeOwner` qualifier on some matching Route.  For example, if you have an interface that lets controllers vote on a unsaved changes dialog, you would do something like this:
```
mcollovati, line 233: Programmatic lookup of beans is a standard CDI feature, so I'm not 100% sure it needs to be added to our documentation.
If we want to add it, as an additional quick help for users, I would try to make it more explicit that it is an optional approach, e.g. by replacing `To resolve ...` with `If you need to programmatically lookup a RouteScoped bean, ....`.

I would probably also remove the `For example ....` part and simplify the example code:

* the `AnnotationLiteral` class
* its instantiation, but using `MainLayout.class` instead of `previousRoute`, to be consistent with the other examples
* the `Instance<...>` part, but using `PreservedBean` instead of `RouteLeaveVoter`, to be consistent with the other examples

mcollovati (changes requested):

mcollovati:
Closing as reworked in #3291

PR #3126, opened 2024-01-12
chore: remove react example workarounds

sissbruecker:
Removes a number of workarounds in the React examples that are not necessary anymore after the recent batch of fixes.

Fixes seem to be only included in alphas for `2.4`, so I had to bump the version. Long term the dependency / version should change to `@vaadin/react-components@24.4.0`.


frontend/demo/component/accordion/accordion-summary.ts
sissbruecker, line 1: Not related to the React fixes, but it was easy to fix. The address object contains a `country` string field, so the combo box needs to use that as value and the view needs to show the `country` property itself. The Flow view works similarly.

tomivirkki (approved):

Issue #100, opened 2021-01-11
The dom-repeat example is broken


mvysny:
The example at https://vaadin.com/docs-beta/v14/flow/polymer-templates/tutorial-template-list-bindings/ doesn't work because of this bug: https://github.com/Polymer/polymer/issues/4135

The workaround is to use `<template is="dom-repeat" items="{{employees}}">`

jouni:
Fixed in https://github.com/vaadin/docs/commit/249194ef2de1409a2e60c3163940fccfce870e45

PR #3146, opened 2024-01-26
Describe session serialization timeout property in Kubernetes Kit

tamasmak:
Issue [#100](https://github.com/vaadin/kubernetes-kit/issues/100)


tamasmak:
> It would be good form to have some text under the heading "Configuring Debug Tool" describing it, before the sub-section on Timeout. Also, are there more configuration possibilities for the debug tool? It might be useful to add them. If not, maybe you can eliminate the section heading and make the timeout sub-section into a higher level (i.e., delete one of the equal-signs).

There are no more configuration possibilities for now, so I've removed the section heading as suggested.

MatthewVaadin (approved):
Looks good - nice and clear.

PR #3188, opened 2024-02-13
docs: document menu-bar reverse-collapse

tomivirkki:
Part of https://github.com/vaadin/platform/issues/5018

### DIFF
--- a/articles/components/menu-bar/index.adoc
+++ b/articles/components/menu-bar/index.adoc
@@ -161,7 +161,9 @@ In versions prior to 24.3, theme names must be used instead  of class names (`th
 
 == Overflow
 
-Items that don't fit into the current width of the menu bar collapse into an overflow menu at the end:
+Items that don't fit into the current width of the menu bar collapse into an overflow menu at the end.
+
+By default, collapsed items are removed from the end of the menu bar, but the component can be configured to remove items from the start instead.
 
 [.example]
 --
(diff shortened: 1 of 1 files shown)


articles/components/menu-bar/index.adoc
```diff
@@ -161,7 +161,9 @@ In versions prior to 24.3, theme names must be used instead  of class names (`th
 
 == Overflow
 
-Items that don't fit into the current width of the menu bar collapse into an overflow menu at the end:
+Items that don't fit into the current width of the menu bar collapse into an overflow menu at the end.
+
+By default, collapsed items are removed from the end of the menu bar, but the component can be configured to remove items from the start instead.
```
DiegoCardoso, line 166: Should it mention the API name?

tomivirkki: We don't usually mention API specifics in the documentation because the same documentation is shared by Flow, Lit, and React users while the API is different for all these component types. The component type-specific API can usually be discovered in the example source. In this case, there is no live example, so it's only available in the API documentation.

DiegoCardoso: Right. Do we also need to add the `since...` tag here?

rolfsmeds (approved):

articles/components/menu-bar/index.adoc
tomivirkki: Done

Screenshot 2024-02-14 at 16 27 16

DiegoCardoso (approved):

PR #3201, opened 2024-02-19
Add Flow React adapter integration guide

platosha:


mshabarov (changes requested):
Looks good to me overall, two comments:

1. This should be cleaned up to not have non-related commits.
2. `react.adoc` article can be simplified a bit, if it tells how to obtain a rendered component first without any state handling, and secondly tells how to handle a state. This would be simpler for beginner reader.

We went through it 3 times in DX tests and no big comments really. Works very well, if you follow this tutorial 👍

caalador (approved):

PR #3219, opened 2024-02-26
chore: make app detected as Hilla in use

tltv:
Empty route in routes.tsx is workaround for Flow hilla detection with alpha5. Also, adds missing `@vaadin/hilla-lit-form` dependency to package.json.

Related-To: vaadin/flow#18792


pom.xml
```diff
@@ -221,22 +221,23 @@
...
+                                // Lit ones only if react is disabled.
+                                // But we need both in the docs project.
+                                // Since this project depends on "com.vaadin:vaadin",
+                                // the following workaround adds the Hilla Lit npm
+                                // dependencies.
+                                def hillaLitAndReactNpmPackages = [
                                         '@vaadin/hilla-react-auth',
                                         '@vaadin/hilla-react-form',
                                         '@vaadin/hilla-react-crud',
+                                        '@vaadin/hilla-lit-form'
```
mshabarov, line 234: This dependency should be added by default, see https://github.com/vaadin/hilla/blob/72a7d6d283fdbaa86d0dab7c7d2bdd18b1db5398/packages/java/hilla/src/main/resources/com/vaadin/flow/server/frontend/dependencies/hilla/components/react/package.json#L13

So once Hilla is detected properly, this dependency should end up in the package.json.

Thus, we probably don't even need `hillaLitAndReactNpmPackages` related functionality, these 4 packages should be added by Flow automatically.

mshabarov (changes requested):

pom.xml
tltv: yep, it works. removing hack.

mshabarov (approved):

mshabarov:
@jouni could I merge this PR by myself, or you wanna take a look as well?

This makes Docs work with Vaadin 24.4.0.

PR #3230, opened 2024-02-28
Fix link to v24 upgrade guide

chkpnt:


jouni (approved):
Thanks for the fix! And apologies for taking so long to react 🙏

PR #3239, opened 2024-02-29
docs: Use Gradle toolchain to specify Java version in upgrade guide

chkpnt:
Nowadays, it's common to use a [Java toolchain](https://docs.gradle.org/current/userguide/toolchains.html) instead of `sourceCompatibility` and `targetCompatibility`.

### DIFF
--- a/articles/upgrading/index.adoc
+++ b/articles/upgrading/index.adoc
@@ -226,16 +226,30 @@ Java 17 or later is required. Below is an example of how to use this version:
     <maven.compiler.target>17</maven.compiler.target>
 </properties>
 ----
+[source,kotlin]
+----
+<source-info group="Gradle (Kotlin DSL)"></source-info>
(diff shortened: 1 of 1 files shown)


articles/upgrading/index.adoc
```diff
@@ -226,16 +226,30 @@ Java 17 or later is required. Below is an example of how to use this version:
     <maven.compiler.target>17</maven.compiler.target>
 </properties>
 ----
+[source,kotlin]
+----
+<source-info group="Gradle (Kotlin DSL)"></source-info>
+plugins {
+    id("java")
+}
```
tepi, line 234:
```
plugins {
    java
}
```
should be enough here I think.

chkpnt: For sure!

tepi (approved):

PR #3244, opened 2024-03-01
docs: fix wrong property name for url-mapping

tepi:


mcollovati (approved):
`vaadin.urlMapping` should work as well, but better using what also the IDE proposes with autocompletion

PR #3291, opened 2024-03-14
Made clear that RouteScopeOwner is a mandatory qualifier

mcollovati:
Take over of #3120 rebased on `next` branch.
It should be back-ported to `latest`.


thigg (approved):
Thanks for picking this up :)

mshabarov (approved):

PR #3292, opened 2024-03-14
Fix how IP wildcard is shown in the online docs

oliveryasuna:


mcollovati (approved):
Needs to be fixed also on `next` branch

PR #3302, opened 2024-03-19
Remove duplicated text

bwajtr:
The text "to observe the frontend" is already a part of the link - no need to repeat it


jouni:
Let's pick this manually to `next` as well.

jouni (approved):

jouni:
`next` fixed in 4b83a34b2059411000ff05b61b7e94993a1f2b8c

PR #3322, opened 2024-04-04
chore: upgrade 24.4.0.alpha20

ZheSun88:


package.json
```diff
@@ -181,37 +184,38 @@
...
       "vite-plugin-checker": "0.6.4",
       "workbox-build": "7.0.0",
       "workbox-core": "7.0.0",
       "workbox-precaching": "7.0.0"
     },
-    "hash": "3c6a7aa18136bf819510e6ae8869817f7ce805ea2f04c48fcc93fd2de06c8db2"
-  }
+    "hash": "d663cd83294a75a6551a2c71fe3e3b3a41603489cc836f6bc8cb73fe9b33ea07"
+  },
+  "type": "module"
```
tomivirkki, line 220: @ZheSun88 I think this change broke the docs build. The project scripts are commonjs

ZheSun88: they are all generated automatically. 
should flow update that?

tomivirkki: I mean the scripts in the project that are not generated by Flow. A few examples:
- https://github.com/vaadin/docs/blob/next/dspublisher/dspublisher-scripts.js
- https://github.com/vaadin/docs/blob/next/webpack.dspublisher.js
- https://github.com/vaadin/docs/blob/next/dspublisher/theme/init.js

ZheSun88: but those are not modified in this commit. 
this one is coming with `mvn install -Pproduction`

tomivirkki: > but those are not modified in this commit.

Yes, this is the reason the build broke. The project was transformed from using commonjs modules by default to using esm, but the scripts were untouched.

ZheSun88: aww.. can you update those files? 
and, should the version update process does any changes to those files? we'd like to just keep the step as simple as possible, so it wont fail on unnecessary changes.

tomivirkki: > can you update those files?

I'm not sure if just updating the files to esm is enough. I'd consider making this kind of an auto-change in a project pretty much breaking and not something one should do in a minor bump.

ZheSun88: i am not sure about that either. the idea of this commit, was making the version update build work again. 

before, there was compilation error and all package-*.json are out of date.

ZheSun88: how ?

ZheSun88: so seems we need to do some changes like this https://github.com/vaadin/start/pull/2988/files

but why this is needed. probably we need to ask flow, as this `type` key is added between 24.3 and 24.4

PR #3371, opened 2024-04-22
docs: add examples for GridPro cell editability

### LINKED ISSUE #3355: Add docs for Grid Pro conditional cell editability
In the `next` branch of docs:
Add new top-level section above _Distinguish Editable / Read-Only Cells_:

> ### Conditional Editability `Since 24.4`
> In some situations you might need to disable editing of specific cells in an edit column. This can be achieved through a function that determines the editability of each cell. This function is automatically re-evaluated when an item in the Grid is modified, so that modifying the value in one cell can immediately affect the editability of other cells.
> 
> `[sample: the example used in the R&D demo by Sascha was fine, so we could probably just use that]`
>

web-padawan:
Fixes #3355

Added the example based on the Flow components IT for the feature:

Screenshot 2024-04-22 at 17 03 38

### DIFF
--- a/src/main/java/com/vaadin/demo/component/gridpro/GridProCellEditability.java
+++ b/src/main/java/com/vaadin/demo/component/gridpro/GridProCellEditability.java
@@ -1,0 +1,69 @@
+package com.vaadin.demo.component.gridpro;
+
+import com.vaadin.flow.component.gridpro.GridPro;
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.router.Route;
+import com.vaadin.demo.DemoExporter; // hidden-source-line
+
+@Route("grid-pro-cell-editability")
+public class GridProCellEditability extends Div {
+    public GridProCellEditability() {
+        GridPro<Transaction> grid = new GridPro<>();
+        grid.setItems(new Transaction("Transaction 1", 100, true),
+                new Transaction("Transaction 2", 200, false));
+        grid.addEditColumn(Transaction::getName).text(Transaction::setName)
+                .setHeader("Name");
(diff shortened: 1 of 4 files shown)


rolfsmeds (approved):

src/main/java/com/vaadin/demo/component/gridpro/GridProCellEditability.java
tomivirkki, line 1: The logic is somehow wrong with the Flow example. It allows me to edit name, but not the other columns

src/main/java/com/vaadin/demo/component/gridpro/GridProCellEditability.java
```diff
@@ -0,0 +1,69 @@
...
+        grid.addEditColumn(Transaction::isApproved)
+                .withCellEditableProvider(
+                        transaction -> !transaction.isApproved())
+                .checkbox(Transaction::setApproved)
+                .setHeader("Approved");
+        // end::snippet[]
+        add(grid);
+    }
+
+    public static class Transaction {
```
tomivirkki, line 31: Not sure why we define a new data type for this particular example. Not a blocker.

web-padawan: Updated to use `Person` like other GridPro examples.

src/main/java/com/vaadin/demo/component/gridpro/GridProCellEditability.java
web-padawan: Replaced the example with a different one.

tomivirkki (approved):

PR #3384, opened 2024-04-25
feat: document Menu annotation

tltv:
Adds "Include Route automatically to Hilla main menu" chapter for Flow's Hilla integration article.

Related-to: vaadin/flow#19109, vaadin/flow#19127


articles/flow/integrations/hilla.adoc
```diff
@@ -105,4 +105,75 @@ import { NavLink } from 'react-router-dom';
...
+@Route("hello")
+public class HelloView extends VerticalLayout {
+    public HelloView() {
+    }
+}
+----
+
+Menu annotation has a `title`, `order` and `icon` attributes. Values are passed to the client side and are available when using `createMenuItems()` utility function to build the menu.
+
+Server side routes are sent to the client menu conditionally by three modes. Default is automatic mode that returns server routes only if root main layout exists when using <<../../hilla/guides/routing.adoc#File Based Routing,File Based Routing>>. Other options are to either always or never send server routes. Mode is configurable with [interfacename]`MenuAccessControl` interface with `PopulateClientMenu` enum.
```
mshabarov, line 132: This should be more explanatory about conditions that are met to see the server routes in menu:
- server route has a Manu annotation
- main layout exists (already mentioned) in `src/main/frontend/views/`, e.g. `src/main/frontend/views/@layout.tsx`,
- navigation to this route is allowed for a given user by Vaadin (see `NavigationAccessControl` chapter).

articles/flow/integrations/hilla.adoc
```diff
@@ -105,4 +105,75 @@ import { NavLink } from 'react-router-dom';
...
+
+Following example changes the default mode to `ALWAYS` in a Spring Framework application:
```
mshabarov, line 134: This needs an explanation what "ALWAYS" mode means.

Currently, the impl is so that "send routes to client no matter if main layout is present or not". I thought that this should simply ignore the pre-requisites above, having only a must have `Menu` annotation as a prerequisite.

Do you remember what was an idea behind this mode?

articles/flow/integrations/hilla.adoc
```diff
@@ -105,4 +105,75 @@ import { NavLink } from 'react-router-dom';
...
+    public MenuAccessControl customMenuAccessControl() {
+        DefaultMenuAccessControl menuAccessControl = new DefaultMenuAccessControl();
+        menuAccessControl.setPopulateClientSideMenu(
+                MenuAccessControl.PopulateClientMenu.ALWAYS);
+        return menuAccessControl;
+    }
+}
+----
+
+Another example changes the default mode to `NEVER` in a non-Spring application by using <<../configuration/properties.adoc#Servlet Initialization Parameters,Servlet Initialization Parameters>> `menu.access.control` with value `org.vaadin.example.CustomMenuAccessControl`. [classname]`DefaultMenuAccessControl` implements [interfacename]`MenuAccessControl`:
```
mshabarov, line 162: "NEVER" needs more explanations: doesn't show server routes in main menu no matter if prerequisites are met or not.

articles/flow/integrations/hilla.adoc
```diff
@@ -105,4 +105,75 @@ import { NavLink } from 'react-router-dom';
 <NavLink to="/flow-route">Navigate to a Flow View</NavLink>
 ----
 
+
+=== Include Route automatically to Hilla main menu
```
mshabarov, line 109: As the Menu annotation will be used also in future for automatic Flow menus, the overall description of `Menu` should be in Flow routing docs, e.g. in https://vaadin.com/docs/latest/routing/layout. This chapter may have less docs, maybe just one sentence and reference link.

This, however, should add a note that automatic menus are only available with Hilla's main layouts, but will be extended to Flow in the future.

mshabarov (changes requested):

articles/flow/integrations/hilla.adoc
tltv: As discussed in daily, it's like this for security reason when using access control at all. When access control is not setup with NavigationAccessControl/ViewAccessChecker, then only Menu annotated routes are send to client. I'll update this with info about the access checks which is completely missing atm.

articles/flow/integrations/hilla.adoc
tltv: In current state server side menu is not supporting the annotation yet in any way. Would it make sense to move this when server side menu is supported? now it could be a bit misleading to move it. In fact, the server side menu documentation could be worth a completely new article when it's done as I believe there may be some additional access control filtering etc. added then.

articles/flow/integrations/hilla.adoc
tltv: updated

articles/flow/integrations/hilla.adoc
tltv: updated

articles/flow/integrations/hilla.adoc
tltv: updated

articles/flow/integrations/hilla.adoc
mshabarov: Ok, sounds reasonable. If we placed the annotation to the proper place from the beginning, it would then make users to get used to it. However, this also would require a tricky explanation why it cannot be used for auto menus in router layout right now. Let's keep it here until we get more feedback.

mshabarov (approved):

PR #3389, opened 2024-04-25
docs: add section about SideNav query parameters support

### LINKED ISSUE #3357: Add docs for SideNav query params support
in the `next` branch of docs:
Add the following top-level section below the intro section (above _Prefix & Suffix elements_):

> ### Automatic Highlighting of Current Item
> The navigation item matching the current URL is automatically highlighted as active.
> #### Query Parameters `Since 24.4`
> If the item’s path contains query parameters, only URLs containing those parameters are considered a match. Additional parameters in the URL not specified in the item path are ignored (i.e. the URL is considered a match).

Also Remove the sentence _"The navigation item whose path matches the current URL is highlighted as active."_ from the intro section.

web-padawan:
Fixes #3357


web-padawan:
Updated to use the correct text for Query Parameters section (as specified in the issue).

rolfsmeds (approved):

Issue #3385, opened 2024-04-25
Hilla FAQ - SSR


landsman:
Hello guys,

I did not find any mention of SSR for the front end. I think that this is a crucial topic for frontend developers, who care about SEO.

Can you please write down if it's possible to use it or even use partial rendering?

I am missing it here:
https://hilla.dev/docs/react/start/faq

I am happy to create PR if you provide me with info or links on how to make it work.

---

If you think this issue is important, add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

Legioth:
Hilla does currently not target the type of application where SEO would be important. Furthermore, there's some significant technical challenges with server-side rendering of a TypeScript frontend on a Java server. I understand that this could be doable with GraalVM but that's quite a special case.

landsman:
Okay, so client side only. This is important to say, ideally in FAQ. Do you agree?

landsman:
PR created, waiting for review.

PR #3390, opened 2024-04-25
docs: add SSR to Hilla FAQ

landsman:
Resolving: https://github.com/vaadin/docs/issues/3385


landsman:
I am not able to see what goes wrong on AWS, can someone check it please?
cc @jouni, @tomivirkki

Btw: It would be good to have build in GitHub Actions for these situations to let me handle it. Then send artifacts to AWS.

platosha (approved):

jouni:
The preview builds are useless at the moment, they don't work for any PRs. And only our IT department has access to that AWS build, to check why they fail.

@landsman, thanks for the suggestion, about doing the build with GH Actions instead. I assume Java and Maven are no problem, and we could run the build here. I’ll try to get that running eventually.

landsman:
@jouni thanks, it is actually even free for OSS project I think. Let me know if I can help with that - I did many github actions workflows similar to this.

PR #3399, opened 2024-04-26
docs: extract file handling sections into separate tab

vursen:
The PR reorganizes the Upload component article so that all information on handling uploaded files in Flow, Lit and React is located under a separate tab "File Handling".

Here is an overview of the changes:

- Renamed "Flow Usage" tab to "File Handling".
- Removed the "Setting Maximum Files" section from "File Handling" (already covered on the main tab).
- Combined "Handling Uploaded Files in Flow" from the main tab with "Defining a Receiver Instance" from "File Handling", as they are essentially about the same.
- Renamed "Handling Upload Requests on Client-Side" to "Handling Upload Requests in Lit and React" and moved it from the main tab to "File Handling".


articles/components/upload/file-handling.adoc
```diff
@@ -0,0 +1,161 @@
...
+    // Do something with the file data...
+});
+----
+
+
+=== FileBuffer
+
+The [classname]`FileBuffer` can handle a single file upload at once. It saves the file on the file system, in the current working directory of the Java application.
+
+Using [classname]`FileBuffer`, configure the component so that only a single file can be selected.
```
vursen, line 82:
```suggestion
Using [classname]`FileBuffer` will configure the component so that only a single file can be selected.
```

rolfsmeds (approved):

PR #3403, opened 2024-04-29
feat(hilla): use the new FilterTransformer in example and adapt doc

### LINKED ISSUE #3346: Document the new Hilla `FilterTransformer`
https://github.com/vaadin/hilla/pull/2321 introduced a filter transformer to help with filters when DTO mappings are in place.

This can be used to simplify the documentation examples about DTOs.

cromoteca:
In Vaadin 24.4 the AutoGrid has a new FilterTransformer that simplifies the handling of DTO field names that do not map directly to entity field names.

This allows to greatly simplify the advanced DTO example.

I chose to keep the old documentation paragraph and move it lower in the page, since in theory it's always possible to find a use case where a completely custom implementation is needed.

Closes #3346.


taefi (approved):
LGTM!

PR #3428, opened 2024-05-15
Hilla-dev exclusions are not needed for non-Spring projects

mshabarov:
As now `hilla-dev` marks `hilla-endpoints` dependency as optional and, if no `vaadin-spring-boot-starter` in a project, the Hilla endpoints are not added, i.e. no need to exclude.

Note that the `hilla-dev` module will be removed soon as a part of unification for Vaadin HotSwap agent plugin.


ZheSun88:
this is just remove some out dated contents. so let us skip the language check for it

ZheSun88 (approved):

PR #3465, opened 2024-05-31
Describe what Flow is

peholmst:


articles/flow/what-is-flow.adoc
```diff
@@ -0,0 +1,42 @@
...
+rich Java API while still being accessible when and where needed. But how does it work under the surface?
+
+Flow acts as a remote control for HTML elements in the browser. Those HTML elements can be simple `<div>`-elements or more complex elements like a `<vaadin-grid>` web component. Flow controls the attributes, properties and children of these HTML elements and can even perform custom JavaScript invocations. On the server side, there are corresponding Java objects that the developer is interacting with. Flow takes care of keeping the server side Java objects and the client side HTML elements in sync with each other.
+
+== Bootstrapping
+
+The initial HTML that is loaded from the browser only contains some static placeholder elements and Flow's client side rendering engine script. This script queries the server for the rendering instructions for the initial view. The server sends over the instructions as JSON. The client side rendering engine interprets these instructions and builds or updates the https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model:[DOM] tree in the browser.
```
Legioth, line 16: > This script queries the server

It depends. The initial UIDL might also be inlined in the initial HTML. Could avoid that detail by using e.g. "gets" which can be interpreted in both ways.

Legioth, line 16: > DOM tree

If we're targeting beginners, then it might be good to make he relationship between DOM and HTML more explicit rather than relying on an external link that doesn't spell things out explicitly.

Legioth (approved):

PR #3472, opened 2024-06-04
Update Hilla docs for 24.4 authentication changes

platosha:


frontend/demo/auth/LoginView.tsx
```diff
@@ -10,47 +10,20 @@ export const config: ViewConfig = {
...
-
-    return null;
-}
-
 // tag::snippet[]
 export default function LoginView() {
   useSignals(); // hidden-source-line
-  const { state, login } = useAuth();
+  const { login } = useAuth();
   const hasError = useSignal<boolean>(false);
```
cromoteca, line 17:
```suggestion
  const hasError = useSignal(false);
```

articles/hilla/lit/guides/upgrading/index.adoc
```diff
@@ -257,6 +257,141 @@ In case you've used Hilla CRUD helpers (e.g., AutoCRUD, AutoGrid, and AutoForm),
 ----
 
 
+[discrete]
+==== Update Authentication Code
```
cromoteca, line 261: Can the `since` tag be useful here?

platosha: The entire section in the upgrade guide is about 24.4. I think, the tag does not belong here.

taefi (approved):

PR #3499, opened 2024-06-19
fix: excluding hilla for a gradle project

caalador:
documents vaadin/platform#6519


mcollovati (approved):

articles/upgrading/index.adoc
```diff
@@ -392,9 +390,9 @@ When using maven:
 </dependency>
 ----
 
-When using gradle:
+This example is for when you're using Gradle:
 
-.build.gradle
+.Build Gradle
```
mcollovati, line 395: `build.gradle` is the name of the file. It seems a bit incorrect to me to change it to `Build Gradle`.
Same for `gradle.properties`.

caalador: does our template support `.[filename]build.gradle` else I guess we should add build.gradle and gradle.properties to the ignore list like we have for pom.xml

mshabarov:
@caalador could you please rebase to `latest`, as now the PR has conflicts.

articles/upgrading/index.adoc
jouni: For code example titles, you can do ``.`build.gradle` ``, or if you really want to, ``.[filename]`build.gradle` ``

In any case, monospace formatted text (backticks) is excluded from Vale rules.

PR #3542, opened 2024-07-19
chore: update vaadin to 24.5.0.alpha5

manolo:
At the same time I needed to remove CE licensing demo and references to fix compilation errors


web-padawan (commented):
We probably should not remove CE articles but just the block that mentions `MyVaadinInitListener`

web-padawan:
The compilation is now fixed by moving the broken Java file in #3553. I will revert related changes from this PR.

web-padawan (approved):
LGTM. Tested locally after rebasing and `npm run dspublisher:start` works fine.

PR #3562, opened 2024-07-29
Compatibility Mode not Support in 14.12

AlainaFaisal:
Adding info boxes for removal of compatibility mode since Vaadin 14.12.

### DIFF
--- a/articles/flow/integrations/osgi/tutorial-osgi-basic.asciidoc
+++ b/articles/flow/integrations/osgi/tutorial-osgi-basic.asciidoc
@@ -15,7 +15,7 @@ include::{articles}/_commercial-banner.asciidoc[opts=optional]
 
 
 [NOTE]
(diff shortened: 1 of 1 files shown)


mcollovati:
Looks good to me.
I've found two other "compatibility mode" mentions, where we can potentially improve the docs:

https://vaadin.com/docs/v14/flow/styling/custom-theme: improve the first note, adding a comment about compatibility mode end of support

https://vaadin.com/docs/v14/flow/integrations/embedding/tutorial-webcomponent-theming#using-multiple-themes has a mention of compatibility mode, maybe add a note there as well?

tepi:
Good catch! We need the note on those two pages too. @AlainaFaisal Can you copy-paste the note to relevant locations on these pages as well?

tepi (approved):

PR #3565, opened 2024-07-29
Document button warning variant.

FrediWa:
This PR adds a code example for the warning button variant.

Technically, it also replaces the error examples with "danger" examples that contain both warning and error variants.

Fixes #3564


FrediWa:
Noticed that there was already an open PR for this

PR #3574, opened 2024-07-31
docs: add Combo Box item class name generator example

### LINKED ISSUE #3544: Add docs for item classname API in ComboBox & MSCB
- Rewrite the ComboBox _"Custom item presentation"_ section to accommodate classnames in addition to renderers.
- Include a live sample
- Add a section to MSCB page with a sample demonstrating chip styling (and link to the CB section about renderers)
- Remember to add the "since 24.5" tag

web-padawan:
Fixes #3544

### DIFF
--- a/articles/components/combo-box/index.adoc
+++ b/articles/components/combo-box/index.adoc
@@ -166,6 +166,40 @@ endif::[]
 
 Use a custom filter to allow the user to search by the rendered properties. It's recommended to make filtering case insensitive.
 
+[role="since:com.vaadin:vaadin@V24.5"]
+== Item Class Names
+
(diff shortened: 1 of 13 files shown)


rolfsmeds (changes requested):
I worry that even/odd striping as an example may give the wrong impression about this feature – that a quick glance at the sample doesn't convey how flexible it is.

How about instead a combobox with something like a small set of fruit or something, mapped to css colors, e.g.
apple: coral
banana: gold
orange: orange
pear: yellowgreen

And maybe apply text color instead of background color in the dropdown, because it kind of looks better?

web-padawan:
> How about instead a combobox with something like a small set of fruit or something, mapped to css colors

Updated.

rolfsmeds (approved):

PR #3139, opened 2024-01-15
add Hilla docs to the `next` branch

taefi:
Fixes #3132

PR #3575, opened 2024-08-01
fix: fix connect client prefix property name (#3128) (CP: latest)

mcollovati:


mcollovati:
This commit was missed in #3139 when Flow and Hilla docs were merged

web-padawan (approved):

PR #3577, opened 2024-08-02
chore: add a note about Unremovable annotation for Quarkus beans

mcollovati:
Related to https://github.com/vaadin/quarkus/issues/147


articles/flow/integrations/quarkus.adoc
```diff
@@ -121,6 +121,10 @@ See these documentation pages for Vaadin CDI features:
 - <<cdi/service-beans#, Vaadin Service Interfaces as CDI Beans>>
 - <<cdi/instantiated-beans#, Using CDI Beans>>
 
+[NOTE]
+====
+Quarkus attempts to remove all unused beans during build by default. However it can't detect the programmatic lookup performed by Vaadin [classname]`QuarkusInstantiator`. Therefore, when implementing beans that are not directly referenced in the application code, you might need to add the [annotationname]`Unremovable` annotation to the class.
+====
```
jouni, line 127: Does this need to be an admonition? Was this something that was hard to notice before or simply missing from the documentation?

```suggestion
Quarkus attempts to remove all unused beans during the build by default. However, it can't detect the programmatic lookup performed by the [classname]`QuarkusInstantiator` class in Vaadin. Therefore, when implementing beans that are not directly referenced by the application code, you might need to add the [annotationname]`@Unremovable` annotation to the class.
```

mcollovati: For CDI features, we are redirecting users to the CDI add-on documentation. However, the unused bean removal is specific to Quarkus. Probably experienced Quarkus user are aware of this, but for beginners might not understand why their bean implementing Vaadin service interfaces are not detected.

mcollovati: At least the admonition should make it more evident that an additional step might be required

jouni: I added back the admonition, and added a title to it.

mshabarov (approved):

PR #3589, opened 2024-08-08
Remove step no longer needed to deploy Vaadin app to Control Center

sosa-vaadin:
- [x] Remove instruction to add a boostrap.yml file in Vaadin apps to be deployed through Control Center.
- [x] Add warning about adding `bootstrap.yml` file or application name to a Vaadin application to be deployed through Control Center.

Fixes https://github.com/vaadin/control-center/issues/405


heruan (approved):

articles/control-center/application-discovery/index.adoc
```diff
@@ -23,26 +23,17 @@ First, you must add the `control-center-starter` dependency to your Vaadin appli
...
+Then build a Docker image of your application with an appropriate [filename]`Dockerfile` file.
 
 [source,shell]
 ----
 docker build -t vaadin/example:0.1.1 .
 ----
 
+.Potential Side-Effects
+[WARNING]
+If your application has its own `bootstrap.yml` file or a name defined (i.e. the property `spring.application.name`) it may cause unexpected side-effects when trying to deploy it through Control Center.
```
jouni, line 35: What are the unexpected side-effects? Let's not leave the reader wondering and guessing what those might be, but rather explain everything we know could happen.

heruan: Good point. On a recent commit we also removed the bootstrap file, so no need to mentioned it here, let's only mention that Control Center will set the `SPRING_APPLICATION_NAME` environment variable with the name choose in the deployment dialog and that will override any application name set in the application properties. Maybe this can be a `NOTE` instead of a `WARNING`?

articles/control-center/application-discovery/index.adoc
```diff
@@ -23,26 +23,17 @@ First, you must add the `control-center-starter` dependency to your Vaadin appli
...
+Then build a Docker image of your application with an appropriate [filename]`Dockerfile` file.
 
 [source,shell]
 ----
 docker build -t vaadin/example:0.1.1 .
 ----
 
+.Potential Side-Effects
+[NOTE]
+Control Center will set the `SPRING_APPLICATION_NAME` environment variable with the name chosen in the deployment dialog and that will override any application name set in the application's properties.
```
heruan, line 35: A lint warning recommends to avoid using "will", let's rephrase:
```suggestion
Control Center sets the `SPRING_APPLICATION_NAME` environment variable with the name chosen in the deployment dialog and that might override any application name set in the application's properties.
```

heruan (changes requested):

heruan (approved):

PR #3587, opened 2024-08-07
docs: add Popover documentation and live examples

web-padawan:
Added Popover documentation and live examples based on the [document](https://docs.google.com/document/d/1ucCxGcHK6xtjTvepL9UzMqESSaPEMUaks70uyJL6UPE/edit?usp=sharing).

PR #3606, opened 2024-08-21
fix: import global theme to enable Lumo utility classes

web-padawan:
This fixes the [ContextMenu styling example](https://vaadin.com/docs/latest/components/context-menu#styling-menu-items) which is currently broken due to the fact that Lumo utility classes are no longer imported globally.

As far as I understand, this is a result of the [Flow change](https://github.com/vaadin/flow/pull/19274) which moved global Lumo imports including `utility-global.js` to the `theme-docs.global.generated.js` and that file isn't currently imported (since `applyTheme` helper used in examples is exported from `theme-docs.generated.js` which no longer contains global modules).


frontend/demo/init.ts
tomivirkki, line 1: I wonder if this should be fixed in the `webpack.dspublisher.js` instead so it would also affect custom themes. Perhaps [this](https://github.com/vaadin/docs/blob/5f5c4c740ac894a3fb96f2209ce966ef6a5cb0ae/webpack.dspublisher.js#L75) line needs an update?

web-padawan: How would you suggest to change it?

BTW, just noticed that `generated-flow-imports.js` includes `color-global.js` and `typography-global.js` but not `utility-global.js`. I guess it's because of the fact that first two are included using `@JsModule` annotation on the `Lumo` class, while `badge-global` and `utility-global` are not.

tomivirkki: Right, I see. It would probably be appropriate to add `import '@vaadin/vaadin-lumo-styles/utility-global.js';` to [`init-browser.ts`](https://github.com/vaadin/docs/blob/5103b9a82ea13e642ae66ee47119ebfb4cd446be/dspublisher/theme/init-browser.ts#L7).

web-padawan: Tested it and it works for docs pages, but not when opening example in a new window.

I tried to add `import '../generated/vaadin'` but that fails with the following error:

```
Failed to compile
../../../../../../vaadin/docs/frontend/generated/index.tsx

Module not found: Can't resolve 'Frontend/routes.js' in '/Users/serhii/vaadin/docs/frontend/generated'
```

tomivirkki: > Tested it and it works for docs pages, but not when opening example in a new window

`import '@vaadin/vaadin-lumo-styles/utility-global.js';` in this file seems to work for both cases but it's not ideal either.

The original fix should be good enough since we can assume "docs" theme to be available in projects using a custom theme also, assuming they set it as the parent theme for "docs" as [documented](https://vaadin.com/docs/latest/tools/dspublisher/custom-theme)...otherwise, some other examples would be missing essential styles also.

tomivirkki (approved):

web-padawan:
Looks like I included this change by accident to #3587. Closing.

PR #3620, opened 2024-08-27
Project structure

peholmst:


peholmst:
@russelljtdyer I will still write an article about multi-module projects, but I would appreciate it if you could review and edit the pages that are already finished (intro, spring boot and single-module projects).

peholmst:
@russelljtdyer Thanks! Yes, I will do quite a lot of more writing before this is published. I'll ping you on Slack when I'm ready.

PR #3627, opened 2024-08-28
docs: document applicationIdentifier property

tltv:
`applicationIdentifier` property was added for 24.5 for DAU License Checking feature but not limited just for it.

### DIFF
--- a/articles/flow/configuration/properties.adoc
+++ b/articles/flow/configuration/properties.adoc
@@ -100,6 +100,10 @@ The following table contains the properties that are defined in the [classname]`
 |===
 |Property Name |Default Value |Description
 
+|`applicationIdentifier`
+|`default-project-id`
+|Application identifier that is by default generated based on the project build settings, e.g. maven's groupId and artifactId, during a production build.
+
 |`brotli`
 |`true`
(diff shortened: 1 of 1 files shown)


articles/flow/configuration/properties.adoc
```diff
@@ -100,6 +100,10 @@ The following table contains the properties that are defined in the [classname]`
 |===
 |Property Name |Default Value |Description
 
+|`applicationIdentifier`
+|`default-project-id`
+|Application identifier that is generated by default based on the project build settings (e.g., Maven's `groupId` and `artifactId`) -- during a production build.
```
mshabarov, line 105: This should also explain a use case: this property is used by the daily active users license model only at the moment, and can be configured by developers to differentiate end-users distribution between same applications (having same group and artifact ids) running in a cluster.

mshabarov (changes requested):

articles/flow/configuration/properties.adoc
tltv: Done

mshabarov (approved):

mshabarov:
This probably needs another language checking after pushing additions @russelljtdyer

PR #3638, opened 2024-09-04
fix: escape special chars correctly in pattern

vursen:
Added missing escaping in Java examples and removed incorrect escaping in React examples. This makes the pattern property work in those examples.

### DIFF
--- a/frontend/demo/component/textfield/react/text-field-constraints.tsx
+++ b/frontend/demo/component/textfield/react/text-field-constraints.tsx
@@ -9,7 +9,7 @@ function Example() {
       required
       minlength={5}
       maxlength={18}
-      pattern="^[+]?[\\(]?[0-9]{3}[\\)]?[\\-s.]?[0-9]{3}[\\-s.]?[0-9]{4,6}$"
+      pattern="^[+]?[\(]?[0-9]{3}[\)]?[\-]?[0-9]{3}[\-]?[0-9]{4,6}$"
       allowedCharPattern="[0-9()+-]"
       label="Phone number"
       helperText="Format: +(123)456-7890"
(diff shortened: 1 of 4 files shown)


frontend/demo/component/textfield/text-field-constraints.ts
```diff
@@ -21,7 +21,7 @@ export class Example extends LitElement {
         required
         min-length="5"
         max-length="18"
-        pattern="^[+]?[\\(]?[0-9]{3}[\\)]?[\\-s.]?[0-9]{3}[\\-s.]?[0-9]{4,6}$"
+        pattern="^[+]?[\\(]?[0-9]{3}[\\)]?[-]?[0-9]{3}[\\-]?[0-9]{4,6}$"
```
vursen, line 24: **note:** The `allowedCharPattern` of this text field rejects spaces and dots, so the pattern doesn't have to allow them either.

frontend/demo/component/textfield/text-field-constraints.ts
```diff
@@ -21,7 +21,7 @@ export class Example extends LitElement {
         required
         min-length="5"
         max-length="18"
-        pattern="^[+]?[\\(]?[0-9]{3}[\\)]?[\\-s.]?[0-9]{3}[\\-s.]?[0-9]{4,6}$"
```
vursen, line 24: **note:** The pattern itself was also modified to disallow spaces and dots, as these characters cannot anyway be entered due to the `allowedCharPattern`.

frontend/demo/component/textfield/react/text-field-constraints.tsx
```diff
@@ -9,7 +9,7 @@ function Example() {
       required
       minlength={5}
       maxlength={18}
-      pattern="^[+]?[\\(]?[0-9]{3}[\\)]?[\\-s.]?[0-9]{3}[\\-s.]?[0-9]{4,6}$"
```
vursen, line 12: JSX doesn't support escaping in static strings.

PR #3640, opened 2024-09-04
docs: explain text field validation

vursen:


articles/components/_input-field-common-features.adoc
```diff
@@ -193,10 +193,16 @@ The pattern is a regular expression used to validate the full value entered into
 [%collapsible]
 ====
 A separate single-character, regular expression can be used to restrict the characters that can be entered into the field. Characters that don't match the expression are rejected.
+
+However, programmatically set values are not subject to this restriction, even if they contain disallowed characters. To validate these values, a pattern constraint should be used additionally.
```
vursen, line 197: I wanted to add this note but I've realized there is a complication. ComboBox and MultiSelectComboBox both have `setAllowedCharPattern`, but they don't have `setPattern` or if they do, it doesn't work, see https://github.com/vaadin/flow-components/issues/6635. 

I'm now not sure. I wonder whether there is actually a use-case for `setAllowedCharPattern` for MultiSelectComboBox. The `required` constraint seems to be only reasonable for this component.

articles/components/_input-field-common-features.adoc
```diff
@@ -158,7 +158,7 @@ CONSTRAINT FEATURES
 .Required
 [%collapsible]
 ====
-Required fields are marked with an indicator next to the label, and become invalid if left empty after having been focused. An error message explaining that the field is required needs to be provided manually.
+Required fields are marked with an indicator next to the label, and become invalid if their value is cleared.
```
rolfsmeds, line 161:
```suggestion
Required fields are marked with an indicator next to the label, and become invalid if their value is first entered and then cleared.
```

articles/components/_input-field-common-features.adoc
rolfsmeds: I guess you could have a CB where you should only be able to enter numbers or something. In any case, both pattern and allowChars are really only meaningful for custom entry. I'd mention it as planned for now and see if we can fix 6635.

rolfsmeds, line 197:
```suggestion
Note, however, that programmatically set values are not subject to this restriction, even if they contain disallowed characters. To validate these values, a pattern constraint can be used in addition.
```

articles/components/date-picker/index.adoc
```diff
@@ -50,42 +50,61 @@ The date can be entered directly using the keyboard in the format of the current
 
 == Validation
 
-To validate a date entered or selected, various ways may be used. For instance, you may want to allow only dates in a certain range, or you might not want to allow certain dates. Below are some of your options for date validation.
+Date Picker provides a validation mechanism based on constraints. Constraints allow you to define criteria that the date must meet to be considered valid. Validation generally occurs when the user initiates a date change, for example by selecting a date from the overlay or through text input followed by Enter press. If the date is invalid, the field is visually highlighted in red, and an error message appears underneath the input.
```
rolfsmeds, line 53:
```suggestion
Date Picker provides a validation mechanism based on constraints. Constraints allow you to define criteria that the date must meet to be considered valid. Validation generally occurs when the user initiates a date change, for example by selecting a date from the overlay or through text input followed by kbd:[Enter]. If the date is invalid, the field is visually highlighted in red, and an error message appears underneath the input.
```

articles/components/date-picker/index.adoc
```diff
@@ -50,42 +50,61 @@ The date can be entered directly using the keyboard in the format of the current
...
+Below is a list of supported constraints with more detailed information:
 
-=== Min & Max Value
+[.example]
+
+[#bad-input]
+.Bad Input
+[%collapsible]
+====
+Bad input refers to any input that cannot be parsed into a valid date. When bad input is applied, the field resets the value to empty and becomes invalid. This constraint is non-configurable and enabled by default.
```
rolfsmeds, line 63:
```suggestion
Bad input refers to any input that cannot be parsed into a valid date. When an unparseble value is entered, the field resets the value to empty and becomes invalid. This constraint is non-configurable and enabled by default.
```

articles/components/email-field/index.adoc
```diff
@@ -75,43 +75,52 @@ endif::[]
 --
 
 
-// Constraints
+== Validation
 
-include::{articles}/components/_input-field-common-features.adoc[tags=constraints-intro;required]
+Email Field provides a validation mechanism based on constraints. Constraints allow you to define criteria that the value must meet to be considered valid. Validation generally occurs when the user initiates a value change, for example by entering input and pressing Enter. If the value is invalid, the field is visually highlighted in red, and an error message appears underneath the input.
```
rolfsmeds, line 80:
```suggestion
Email Field provides a validation mechanism based on constraints. Constraints allow you to define criteria that the value must meet to be considered valid. Validation generally occurs when the user initiates a value change, for example by entering input and pressing kbd:[Enter]. If the value is invalid, the field is visually highlighted in red, and an error message appears underneath the input.
```

articles/components/text-field/index.adoc
```diff
@@ -71,34 +71,45 @@ endif::[]
 --
 
 
-// Constraints
+== Validation
 
-include::{articles}/components/_input-field-common-features.adoc[tags=constraints-intro;required;pattern;min-and-max-length;allowed-chars]
+Text Field provides a validation mechanism based on constraints. Constraints allow you to define criteria that the value must meet to be considered valid. Validation generally occurs when the user initiates a value change, for example by entering input and pressing Enter. If the value is invalid, the field is visually highlighted in red, and an error message appears underneath the input.
```
rolfsmeds, line 76:
```suggestion
Text Field provides a validation mechanism based on constraints. Constraints allow you to define criteria that the value must meet to be considered valid. Validation generally occurs when the user initiates a value change, for example by entering input and pressing kbd:[Enter]. If the value is invalid, the field is visually highlighted in red, and an error message appears underneath the input.
```

rolfsmeds (changes requested):

articles/components/_input-field-common-features.adoc
```diff
@@ -144,7 +144,7 @@ Disabled fields can be useful in situations where they can become enabled based
 
 
 ////
-CONSTRAINT FEATURES
+VALIDATION FEATURES
 ////
 
 // tag::constraints-intro[]
```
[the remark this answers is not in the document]
vursen: I removed this snippet because I considered that it's too small to justify reuse.

vursen:
I decided to exclude Select and RadioButton from the PR scope. These components should have examples demonstrating required validation, probably similar to the Checkbox article. However, creating such examples would require using Binder, as the `validate()` method isn't available yet. I propose to postpone their documentation update until we have the `validate()` method available, see https://github.com/vaadin/flow-components/issues/6634

articles/components/date-time-picker/index.adoc
```diff
@@ -180,10 +192,11 @@ include::{root}/frontend/demo/component/datetimepicker/react/date-time-picker-va
 endif::[]
 --
 
-ifdef::flow,lit[]
+It's important to ensure an appropriate error message is configured for each constraint violation to provide users with clear feedback.
+
 === Custom Validation
 
-If the `min` and `max` values aren't enough, you can also apply custom validation.
+For more complex cases where constraint validation isn't enough, Flow and Hilla offer the Binder API that allows you to define custom validation rules. This is useful, for example, when you want to restrict date selection to weekdays (Monday-Friday) and time selection to opening hours. In the following example, try selecting a date that's on a weekend or a time that's outside opening hours to see a custom validation message.
```
vursen, line 199:
```suggestion
For more complex cases where constraint validation isn't enough, Flow and Hilla offer the Binder API that allows you to define custom validation rules. This is useful, for example, when you want to restrict date selection to weekdays (Monday-Friday) and time selection to only certain opening hours. In the following example, try selecting a date that's on a weekend or a time that's outside opening hours to see a custom validation message.
```

vursen:
Just noticed that there is an issue with missing margin when an accordion isn't the last element in a section:

image

PR #3656, opened 2024-09-12
docs: add more details about applicationIdentifier

tltv:
Explain how default application identifier is also hashed and prefixed with 'app-'.


articles/flow/configuration/properties.adoc
```diff
@@ -102,7 +102,7 @@ The following table contains the properties that are defined in the [classname]`
 
 |`applicationIdentifier`
 |`default-project-id`
-|Application identifier that's generated by default based on the project build settings (e.g., Maven's `groupId` and `artifactId`) -- during a production build. The property is used by the daily active users license model only for the moment. It can be configured by developers to differentiate end-users distribution between the same applications running in a cluster -- ones that have the same group and artifact IDs.
+|Application identifier that's generated by default based on the project build settings (e.g., Maven's `groupId` and `artifactId` are hashed and prefixed with `app-`) -- during a production build. The property is used by the daily active users license model only for the moment. It can be configured by developers to differentiate end-users distribution between the same applications running in a cluster -- ones that have the same group and artifact IDs.
```
mshabarov, line 105:
```suggestion
|Application identifier that's generated by default based on the project build settings (e.g., Maven's `groupId` and `artifactId`, the string value is then hashed and prefixed with `app-`) -- during a production build. The property is used by the daily active users license model only for the moment. It can be configured by developers to differentiate end-users distribution between the same applications running in a cluster -- ones that have the same group and artifact IDs.
```

mshabarov (approved):

PR #3679, opened 2024-09-17
fix: only close Popover dropdown field on non-empty value

### LINKED ISSUE #3678: [popover] closes when selecting value
### Description

When using a select or e.g. datepicker in a popover - the popover closes once a value is selected.

### Expected outcome

Popover stays open.

### Minimal reproducible example

https://vaadin.com/docs/latest/example?embed=popover-dropdown-field-wc.js&import=component/popover/popover-dropdown-field.ts,component/popover/react/popover-dropdown-field.tsx

### Environment

Vaadin version(s): latest docs
OS:

web-padawan:
Fixes #3678

Selecting a date in any of the pickers triggers a value change event in the range selector, which currently unexpectedly closes the popover in the example (unlike Lit / React versions). Updated the listener to ignore empty string values.


src/main/java/com/vaadin/demo/component/popover/PopoverDropdownField.java
```diff
@@ -50,7 +50,9 @@ public PopoverDropdownField() {
         rangeSelector.setItems("Today", "Last week", "Last month",
                 "Year to date", "Last year", "Past 5 years");
         rangeSelector.addValueChangeListener((e) -> {
-            updateRange(e.getValue());
+            if (!e.getValue().equals("")) {
+                updateRange(e.getValue());
+            }
```
vursen, line 55:
```suggestion
            if (e.isFromClient()) {
                updateRange(e.getValue());
            }
```

web-padawan: Thanks, good point. Done.

vursen (approved):

PR #3689, opened 2024-09-19
docs: better documentation of reactEnable

tltv:
Document in more detail how `reactEnable` plugin configuration property and `vaadin.react.enable` configuration property works.

RelatedTo: vaadin/flow#19955

### DIFF
--- a/articles/flow/configuration/properties.adoc
+++ b/articles/flow/configuration/properties.adoc
@@ -194,7 +194,7 @@ The following table contains the properties that are defined in the [classname]`
 
 |`react.enable`
 |`true`
-|Whether to use React Router, add React core dependencies, React integration helpers and Vaadin's provided React components (`@vaadin/react-components`). Fallbacks to `vaadin-router`, excludes all React dependencies and adds `Lit` dependencies, if set to `false`.
+|Whether to use React Router, add React core dependencies, React integration helpers and Vaadin's provided React components (`@vaadin/react-components`). Fallbacks to `vaadin-router`, excludes all React dependencies and adds `Lit` dependencies, if set to `false`. Plugin configuration name for the same property is `reactEnable`. See <<../../upgrading#, Hilla & React Dependencies>> for more information.
 
 |`requestTiming`
 |`true` for development mode; `false` for production mode
--- a/articles/upgrading/index.adoc
+++ b/articles/upgrading/index.adoc
@@ -333,8 +333,11 @@ Vaadin Flow applications don't need any changes regarding React or Hilla depende
 - having Hilla or React dependencies is undesirable; and
 - you need the legacy `vaadin-router` (e.g. if your project uses an add-on that needs it).
 
-To opt-out, use the `reactEnable` configuration parameter. Disabling it excludes React completely from your project and fallbacks to `vaadin-router` and Lit:
+To opt-out, use always the `reactEnable` plugin configuration parameter. Disabling it excludes React completely from your project in production mode and fallbacks to `vaadin-router` and Lit. Use also `vaadin.react.enable` configuration property always when running the application in development mode without Maven or Gradle (E.g. with `java` command or with IDE's Run option --which ignores Maven plugin configurations).
 
+Both, `reactEnable` plugin configuration parameter and `vaadin.react.enable` configuration property should always have the same value to avoid confusion over which one is effective. In development mode, `vaadin.react.enable` configuration property has priority.
+
+Set `reactEnable` plugin configuration parameter in the `vaadin-maven-plugin`. It's effective for builds and in development runtime when run with Maven:
 
 [source,xml]
 .pom.xml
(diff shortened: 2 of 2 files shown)


articles/flow/configuration/properties.adoc
```diff
@@ -194,7 +194,7 @@ The following table contains the properties that are defined in the [classname]`
 
 |`react.enable`
 |`true`
-|Whether to use React Router, add React core dependencies, React integration helpers and Vaadin's provided React components (`@vaadin/react-components`). Fallbacks to `vaadin-router`, excludes all React dependencies and adds `Lit` dependencies, if set to `false`.
+|Whether to use React Router, add React core dependencies, React integration helpers and Vaadin's provided React components (`@vaadin/react-components`). Fallbacks to `vaadin-router`, excludes all React dependencies and adds `Lit` dependencies, if set to `false`. Plugin configuration name for the same property is `reactEnable`. See <<../../upgrading#hilla-react-dependencies, Hilla & React Dependencies>> for more information.
```
tepi, line 197: Maybe change the added part to something like "Vaadin plugin configuration for the reactEnable property needs to be set to match the configuration property value." 

Also, similar mention (and link to upgrading page?) should be added to the maven config page https://vaadin.com/docs/latest/flow/configuration/maven#properties

articles/upgrading/index.adoc
```diff
@@ -333,8 +333,11 @@ Vaadin Flow applications don't need any changes regarding React or Hilla depende
 - having Hilla or React dependencies is undesirable; and
 - you need the legacy `vaadin-router` (e.g. if your project uses an add-on that needs it).
 
-To opt-out, use the `reactEnable` configuration parameter. Disabling it excludes React completely from your project and fallbacks to `vaadin-router` and Lit:
+To opt-out, use always the `reactEnable` plugin configuration parameter. Disabling it excludes React completely from your project in production mode and fallbacks to `vaadin-router` and Lit. Use also `vaadin.react.enable` configuration property always when running the application in development mode without Maven or Gradle (E.g. with `java` command or with IDE's Run option --which ignores Maven plugin configurations).
 
+Both, `reactEnable` plugin configuration parameter and `vaadin.react.enable` configuration property should always have the same value to avoid confusion over which one is effective. In development mode, `vaadin.react.enable` configuration property has priority.
+
+Set `reactEnable` plugin configuration parameter in the `vaadin-maven-plugin`. It's effective for builds and in development runtime when run with Maven:
```
tepi, line 340: Please simplify this a bit. Should be ok to just tell that set both the plugin parameter and the configuration property to false, and if changed later, make sure their values always match.

articles/upgrading/index.adoc
```diff
@@ -348,9 +351,44 @@ To opt-out, use the `reactEnable` configuration parameter. Disabling it excludes
 </plugin>
 ----
 
+Set `vaadin.react.enable` application property in the `application.properties` file when using Spring. It's effective in runtime in development mode. It's also effective in production mode, but only for runtime features. It's not affecting already build production bundle content:
```
tepi, line 354: Consider starting this part with "For Spring projects.." or "When using Spring.." so it's immediately obvious who this part is for.

Also I'd leave "It's effective in runtime in development mode. It's also effective in production mode, but only for runtime features. It's not affecting already build production bundle content" out and replace it with a mention that in addition to the application property you still need to set the plugin configuration to match

articles/upgrading/index.adoc
```diff
@@ -348,9 +351,44 @@ To opt-out, use the `reactEnable` configuration parameter. Disabling it excludes
...
+
+To make it easier to keep both properties same, Spring supports https://docs.spring.io/spring-boot/how-to/properties-and-configuration.html#howto.properties-and-configuration.expand-properties.maven[Automatic Property Expansion with Maven] and https://docs.spring.io/spring-boot/how-to/properties-and-configuration.html#howto.properties-and-configuration.expand-properties.gradle[with Gradle]. Please follow the instructions in the Spring documentation to enable this feature.
+
+
+When running with Maven and using `spring-boot-starter-parent`, property value can have `@reactEnable@` which is replaced with the value of the `reactEnable` Maven project property. Otherwise, please follow the https://docs.spring.io/spring-boot/how-to/properties-and-configuration.html#howto.properties-and-configuration.expand-properties.maven[Spring documentation] to learn how to enable `resources` filtering and set `maven-resources-plugin` configurations.
```
tepi, line 367: Quite a lot of very spring-specific stuff here, but I guess it's ok if it's seen as needed.

articles/upgrading/index.adoc
```diff
@@ -348,9 +351,44 @@ To opt-out, use the `reactEnable` configuration parameter. Disabling it excludes
...
+Set `vaadin.react.enable` application property in the `application.properties` file when using Spring. It's effective in runtime in development mode. It's also effective in production mode, but only for runtime features. It's not affecting already build production bundle content:
+
 [source,terminal]
+.application.properties
 ----
--Dvaadin.react.enable=false
+vaadin.react.enable=false
+----
+
+Alternatively, set `vaadin.react.enable` <<../flow/configuration/properties#system-properties, system property>> or `react.enable` <<../flow/configuration/properties#servlet-initialization-parameters, initialization parameter>> for example when Spring is not used. Notice that system property has priority over application property. Also notice that running Spring application with Maven, you may need to wrap system property in `-Dspring-boot.run.jvmArguments=vaadin.react.enable=false"`.
```
tepi, line 362: The "Alternatively..." part about properties is a bit weird here since I think that was explained on lines 336-340 as the first option, before the Spring part.

tltv:
it would probably make sense to refactor most changes under headers and start from the most important topic, which is how to disable react, and then rest are focusing on telling why and also spring/non-spring specific details

tepi:
> it would probably make sense to refactor most changes under headers and start from the most important topic, which is how to disable react, and then rest are focusing on telling why and also spring/non-spring specific details

That would make more sense (and add a bit more work).

tltv:
> That would make more sense (and add a bit more work).

I'll start to do this and of course will take into account all other comments at the same time.

articles/flow/configuration/properties.adoc
tltv: Done

articles/flow/configuration/maven.adoc
```diff
@@ -86,7 +86,7 @@ Define whether the application is running in production mode. Defaults to `false
 The folder where the [filename]`package.json` file is located. Defaults to `${project.basedir}`.
 
 `reactEnable`::
-Whether to use React Router, add React core dependencies, React integration helpers and Vaadin's provided React components (`@vaadin/react-components`). Fallbacks to `vaadin-router`, excludes all React dependencies and adds `Lit` dependencies, if set to `false`. Defaults to `true`.
+Whether to use React Router, add React core dependencies, React integration helpers and Vaadin's provided React components (`@vaadin/react-components`). Fallbacks to `vaadin-router`, excludes all React dependencies and adds `Lit` dependencies, if set to `false`. Defaults to `true`. Vaadin plugin configuration for the `reactEnable` property needs to be set to match the configuration property value. See <<../../upgrading#hilla-react-dependencies, Hilla & React Dependencies>> for more information.
```
tepi, line 89: Since this is the plugin's `reactEnable` description, I think the added part should refer to `react.enable` or `vaadin.react.enable` property instead of `reactEnable`.

articles/upgrading/index.adoc
```diff
@@ -333,8 +333,11 @@ Vaadin Flow applications don't need any changes regarding React or Hilla depende
 - having Hilla or React dependencies is undesirable; and
 - you need the legacy `vaadin-router` (e.g. if your project uses an add-on that needs it).
 
-To opt-out, use the `reactEnable` configuration parameter. Disabling it excludes React completely from your project and fallbacks to `vaadin-router` and Lit:
+==== Opting-Out the React
```
tepi, line 336: Maybe "Opting out of using React"

articles/upgrading/index.adoc
```diff
@@ -348,10 +351,65 @@ To opt-out, use the `reactEnable` configuration parameter. Disabling it excludes
...
+vaadin.react.enable=false
+----
+
+Disabling it excludes React completely from your project and fallbacks to `vaadin-router` and Lit.
+
+===== Why Are There Two Places for the Same Property?
+
+Two properties exists mainly to support the development and production modes. `reactEnable` is effective for both and `vaadin.react.enable` only for development.
+
+Production package is build by running the Vaadin Maven / Gradle plugin with `build-frontend` goal. Package contains the `reactEnable` plugin configuration parameter value in the build info file inside the package, which is used as a default value in runtime.
```
tepi, line 367: Should package be bundle here as that is referred to later as well

tepi (changes requested):
Much better now! A few comments but mostly ready for language check.

articles/upgrading/index.adoc
tltv: It's same in few other titles which would become a bit longer with this format, so I don't know if it's necessary. If it is, then I'll change it to all like this.

tepi: Ok, if we want to keep it short, how about "Opting-Out of React"?

articles/upgrading/index.adoc
tltv: I use package word here because build info file is actually in the jar/war package. It's, not part of the production bundle.

But for clarity, this could say both here: "Production package and bundle is build by running ...".

tepi: Alright, then it makes sense.

PR #3702, opened 2024-09-19
add new line to the encoded passKey to allow correctly be decoded

manolo:
related to: https://github.com/vaadin/control-center/issues/519

### DIFF
--- a/articles/control-center/getting-started/index.adoc
+++ b/articles/control-center/getting-started/index.adoc
@@ -54,7 +54,7 @@ The wizard requires a passkey to proceed. Run the following command to retrieve 
 .Terminal
 [source,bash]
 ----
-kubectl -n control-center get secret control-center-passkey -o jsonpath='{.data.passkey}' | base64 --decode
+kubectl -n control-center get secret control-center-passkey -o jsonpath='{.data.passkey}{"\n"}' | base64 --decode
 ----
 
 On Windows, you might need to remove `| base64 --decode` and use a separate tool to decode the base64 output.


articles/control-center/getting-started/index.adoc
```diff
@@ -54,7 +54,7 @@ The wizard requires a passkey to proceed. Run the following command to retrieve
 .Terminal
 [source,bash]
 ----
-kubectl -n control-center get secret control-center-passkey -o jsonpath='{.data.passkey}' | base64 --decode
+kubectl -n control-center get secret control-center-passkey -o jsonpath='{.data.passkey}{"\n"}' | base64 --decode
```
heruan, line 57: The suggested change does change the output for me.

heruan: In macOS 15.0 the original command works fine:
```
$ kubectl -n control-center get secret control-center-passkey -o jsonpath='{.data.passkey}' | base64 --decode
290dc60e-26a4-4a2f-969f-3cad21f66840
```

heruan: I think the issue relies on different implementations of `base64`. What about this command?
```
kubectl -n control-center get secret control-center-passkey -o json | jq -r '.data.passkey | @base64d'
```

heruan: Or even this one that only uses `kubectl` without piping to other commands:
```
kubectl -n control-center get secret control-center-passkey -o go-template='{{ .data.passkey | base64decode }}'
```

heruan:
I have tested on another machine. There I get a % char after the decoded passkey. Is that the issue? We could safely strip that server side.

manolo:
Take a look to the attached screenshot to see what happens in my machine:
- macOS: Sonoma 14.5
- docker desktop: 4.33
- k8s: v1.32.0 (delivered with docker desktop)
- iTerm: 3.5.4

As you can see running the full command without the NL base64 does not decodes anything.
Also you can see that running only the kubectl command without the pipe, the output does not have any NL at the end so that is the problem.


Screenshot 2024-09-20 at 08 01 48

Note it's not a problem of `base64`, since the equivalente `openssl base64 -d` has the same error


Screenshot 2024-09-20 at 09 39 02

In any case merging this PR is safe since adding multiples new lines after the encoded string does not affect the result

articles/control-center/getting-started/index.adoc
manolo: That's fine since it does not break anything, it just fixes in systems where it does not work

heruan: Good, we will go with this command then:
```suggestion
kubectl -n control-center get secret control-center-passkey -o go-template='{{ .data.passkey | base64decode }}'
```
We will also strip any `%` from the value before validation, if they happen to be mistakenly copied from the terminal with the passkey. See vaadin/control-center#525

heruan (changes requested):

heruan:
> As you can see running the full command without the NL base64 does not decodes anything.

I think this is a problem from the terminal configuration, that somehow does not show the line. I can't replicate this with the default iTerm configuration. At this point I would go with the [suggested change](https://github.com/vaadin/docs/pull/3702#discussion_r1768147802), so we only use `kubectl` and it already adds a newline to the output.

heruan:
Fixed in #3752

PR #3716, opened 2024-09-24
fix(gradle): replace outdated properties from hilla to vaadin

DuncanCasteleyn:
The currently documented properties are outdated and no longer work


taefi (approved):
The whole document needs to be update with `vaadin` namespace instead of `hilla`, but this change on its own LGTM, and can be merged.
Thanks for the contribution!

Issue #3700, opened 2024-09-19
Docs for Java Records support in Binder


mshabarov:
**Tell us about your request**

Documentation for https://github.com/vaadin/flow/pull/19806 in https://vaadin.com/docs/latest/flow/binding-data/components-binder-beans.

---

If you think this issue is important, add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

PR #3746, opened 2024-09-30
Add binder record methods

tepi:
Fixes https://github.com/vaadin/docs/issues/3700

### DIFF
--- a/articles/flow/binding-data/components-binder-load.adoc
+++ b/articles/flow/binding-data/components-binder-load.adoc
@@ -176,4 +176,25 @@ binder.addStatusChangeListener(event -> {
 ----
 
 
+== Using Java Records with Binder
+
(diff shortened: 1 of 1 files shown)


articles/flow/binding-data/components-binder-load.adoc
```diff
@@ -176,4 +176,25 @@ binder.addStatusChangeListener(event -> {
 ----
 
 
+== Using Java Records with Binder
+
+Instead of Java Beans, it is also possible to use the [classname]`Binder` with Java Records. However, since Java Records are immutable, only manual reading and writing can be used when the data type is a record. This also means that methods relying on bean data type such as `writeBeanAsDraft`, `writeBean` and `setBean` will throw an exception when called for a Binder with a record data type.
```
mcollovati, line 181: I wonder if we should mention that for records, `refreshFields()` will always clear all fields.
It could be obvious reading the method Javadoc, but it could be good to add a note also here.

mcollovati (approved):
Looks good to me.
I added a comment for a potential minor change, but not a blocker.

PR #3778, opened 2024-10-08
fix: add note on access requirement in layout

caalador:
Add note for access controll of
layout when access protection in use.

### DIFF
--- a/articles/flow/routing/layout.adoc
+++ b/articles/flow/routing/layout.adoc
@@ -25,6 +25,10 @@ public class MainLayout extends Div
 }
 ----
 
+[NOTE]
+If the application is using access protection the view access is checked for the Layout as well as the view itself.
+To make the layout accessible add [annotationname]`@AnonymousAllowed` (or other access annotation) on the `MainView` so that the request is not denied.
+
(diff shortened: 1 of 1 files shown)


articles/flow/routing/layout.adoc
```diff
@@ -25,6 +25,10 @@ public class MainLayout extends Div
 }
 ----
 
+[NOTE]
+If the application is using access protection the view access is checked for the Layout as well as the view itself.
+To make the layout accessible add [annotationname]`@AnonymousAllowed` (or other access annotation) on the `MainView` so that the request is not denied.
```
tltv, line 30: Does this actually mean `MainLayout` instead, if it's reference to previous example?

articles/flow/routing/layout.adoc
```diff
@@ -25,6 +25,10 @@ public class MainLayout extends Div
 }
 ----
 
+[NOTE]
```
mshabarov, line 28: I don't think this needs a NOTE tag, can be just a plain text.

articles/flow/routing/layout.adoc
caalador: Yes. Mixing up all the time what is where. Fixed

articles/flow/routing/layout.adoc
caalador: removed note

tltv (approved):

PR #3444, opened 2024-05-22
fix: do not use static drop mode for grids in dnd examples

tomivirkki:
Currently, all the grids in the drag & drop examples section use static drop mode values. This means that the drop indicator is shown for them even while dragging unrelated things like text or rows from another example:

https://github.com/vaadin/docs/assets/1222264/f780e578-fe75-41af-b1ed-4ae7725f8753

This PR fixes the issue by updating the examples to dynamically apply the drop mode when a row drag starts and clear the drop mode when the row drag ends.

Related to https://github.com/vaadin/flow-components/issues/6306

PR #3786, opened 2024-10-09
fix: revert static drop mode for tree grid

vursen:
A while ago, https://github.com/vaadin/docs/pull/3444 updated the grid examples to set the drop mode dynamically. However, in tree grids, setting the drop mode dynamically on drag start causes an immediate reset of the scroll position. This completely disrupts the drag-and-drop process unless the grid is at the top. While losing the scroll position is a known limitation in tree grids, we should avoid recommending configurations that result in a reset _right_ at the start of dragging.

Related to https://github.com/vaadin/flow-components/issues/6662


tomivirkki (approved):
Using a static drop mode also causes some confusing UX, but maybe it's the lesser evil. Ideally tree grid should be fixed to not lose its sroll position on drop mode change.

vursen:
I created https://github.com/vaadin/flow-components/issues/6713 in which I described a solution that could resolve this issue.

PR #3793, opened 2024-10-10
Changed http links to https in the v23 branch

edler-san:


manolo (approved):

articles/configuration/node-js.asciidoc
```diff
@@ -63,8 +63,8 @@ The [filename]`.npmrc` file structure is `ini` (like Java properties files).
 It includes pairs of key-values separated by `=`.
 Here is an example of the content of such a file with proxy settings:
 ----
-proxy=http://myusername:s3cr3tpassw0rd@proxyserver1:8085"
-https-proxy=http://myusername:s3cr3tpassw0rd@proxyserver1:8086"
+proxy=https://myusername:s3cr3tpassw0rd@proxyserver1:8085"
+https-proxy=https://myusername:s3cr3tpassw0rd@proxyserver1:8086"
```
manolo, line 67: I dont think this might work, probably 8085 and 8086 ports are not proper https services but http

manolo: Actually there are multiple `:80*` coincidences from here, they should be kept in http

manolo (changes requested):

articles/configuration/node-js.asciidoc
edler-san: Reverted those

PR #3814, opened 2024-10-16
feat!: align spring boot version with 24.5

knoobie:
Not sure how you handle different minor versions in the docs, that part is up to you :) This PR is here to create a "discussion"

### DIFF
--- a/articles/compatibility.adoc
+++ b/articles/compatibility.adoc
@@ -14,7 +14,7 @@ This page shows the minimum required versions of supported technologies for Vaad
 |Technology|Version
 
 | Java| 17 or later
-| Spring Boot| 3.0 or later
+| Spring Boot| 3.3 or later
 | Node.js| 18 or later
 | Maven| 3.5 or later (except for 3.8.2 and 3.8.3)
 | Gradle| 8.5 or later


mshabarov:
It's not only Spring Boot version that isn't the latest. 
I was going to suggest adding a link to Vaadin Platform release notes, but turned out that it's already in place right in the beginning, see https://vaadin.com/docs/latest/compatibility.

Instead of changing versions in the table, I'd suggest rephrasing this sentence
```
See the [release notes](https://github.com/vaadin/platform/releases/tag/24.0.0) for more details.
```
to something like:
```
Minimum supported versions may vary between Vaadin 24 minor releases. For details on specific minor versions, please refer to the [release notes](https://github.com/vaadin/platform/releases/latest | https://github.com/vaadin/platform/releases).
```

jouni:
Previous minors are only supported for 3 months after the next minor has been released, right? The latest documentation covers the latest supported versions, not all minors for the current major version (V24 today). So if 24.4 and 24.5 require Spring Boot 3.3, then that’s what we should indicate.

Legioth:
Updated based on Vaadin 24.8 in https://github.com/vaadin/docs/pull/4521.

PR #3838, opened 2024-10-22
chore: add Notification to avoid error with Astro DSP

web-padawan:
This is a workaround for a problem with some Notification Java examples in Astro based DSP.

Steps to reproduce the problem:

- Open the "Notification" page
- Go to other component page
- Return to the "Notification" page
- Click a button to show notification

There is a following error in the console originating from `setText()` method:

```
FlowClient.js:74 Uncaught TypeError: Cannot set properties of undefined (setting 'textContent')
    at _Notification.renderer (eval at Ut (FlowClient.js:1049:101), <anonymous>:3:94)
    at _Notification.requestContentUpdate (chunk-TD75QAJX.js:593:10)
    at Object.eval (eval at Ut (FlowClient.js:1049:101), <anonymous>:3:11
```

Seems like an edge case of auto-adding mechanism. When adding to UI manually there is no error.

### DIFF
--- a/src/main/java/com/vaadin/demo/component/notification/NotificationPosition.java
+++ b/src/main/java/com/vaadin/demo/component/notification/NotificationPosition.java
@@ -29,14 +29,17 @@ public class NotificationPosition extends Div {
     // tag::createButton[]
     private Button createButton(Notification.Position position) {
         Button button = new Button(position.getClientName());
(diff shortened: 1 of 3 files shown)


web-padawan:
Closing as this will be fixed in the DSP instead.

PR #3840, opened 2024-10-23
docs: fix broken cross-references in Classic Components

web-padawan:
Discovered these when building docs locally:

```
Broken cross reference /tools/routing/exceptions in C:\Users\serhii\docs\articles\tools\modernization-toolkit\classic-components.adoc
Broken cross reference /tools/create-ui/shortcut in C:\Users\serhii\docs\articles\tools\modernization-toolkit\classic-components.adoc
```


tomivirkki (approved):

jouni:
For future reference: you don't need to use the `{articles}` prefix anymore. Absolute cross ref paths should resolve correctly, using `ARTICLES_PATH` as the base path.

PR #3841, opened 2024-10-23
chore: Mention the readRecord shortcut for Records

mshabarov:
Related with https://github.com/vaadin/flow/pull/20323

### DIFF
--- a/articles/flow/binding-data/components-binder-load.adoc
+++ b/articles/flow/binding-data/components-binder-load.adoc
@@ -155,7 +155,7 @@ binder.writeBeanAsDraft(person, true);
 
 == Tracking Binding Status
 
-[classname]`Binder` tracks which bindings have been updated by the user and which are in an invalid state. When the `readBean` method is invoked, the initial value of each field is stored. 
+[classname]`Binder` tracks which bindings have been updated by the user and which are in an invalid state. When the `readBean` method is invoked, the initial value of each field is stored.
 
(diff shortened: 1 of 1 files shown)


articles/flow/binding-data/components-binder-load.adoc
```diff
@@ -178,15 +178,15 @@ binder.addStatusChangeListener(event -> {
...
+Instead of using Java Beans, it's possible to use the [classname]`Binder` with Java Records. Since Java Records are immutable, only manual reading and writing can be used when the datatype is a record. This means that methods relying on bean datatypes such as `writeBeanAsDraft`, `writeBean`, and `setBean` throw an exception when called for a Binder with a record datatype. Additionally, since records can only be read via `readRecord`, the `refreshFields` method clears all of the Binder fields.
 
-For reading a record, the `readBean` method should be used. It works for both beans and records. Below is an example of this:
+For reading a record, the `readRecord` method should be used. It throws an exception, if used with beans. Below is an example of this:
```
tepi, line 183: Maybe with this change the last two sentences need some work. Now it could be understood so that the example is about throwing the exception, which it is not.

Something like "For reading a record, the `readRecord` method should be used as shown in the example below. This method throws an exception, if used with beans."

mshabarov: Good addition, changed as you suggested.

PR #3862, opened 2024-10-28
[Control Center] Add a section about displaying buttons based on authentication status

MatthewVaadin:
Based on feedback, a new section has been added to the login/logout page to show how to display the correct button based on the authenticated status of the user.


articles/control-center/identity-management/login-logout.adoc
```diff
@@ -1,11 +1,11 @@
 ---
-title: Login and Logout Links
-description: How to create login and logout links or buttons using Control Center identity management security.
+title: Login and Logout
+description: How to create login and logout controls using Control Center identity management security.
```
tamasmak, line 3: Should these changes also be applied to this [section](https://github.com/vaadin/docs/blob/latest/articles/control-center/identity-management/index.adoc?plain=1#L84-L86)?

tamasmak (approved):

PR #3870, opened 2024-10-30
docs: Adding more details about the copilot wrap with... feature.

czp13:
The feature description is currently too short and misses some key details. This PR adds more details and insights to the feature.

Could you please review and let me know if I need to change something! 🙇 

Fixes https://github.com/vaadin/copilot-internal/issues/3601


MarcinVaadin:
Ok for me, @czp13 please fix merge conflict.

czp13:
> Ok for me, @czp13 please fix merge conflict.
I was wondering why was this blocked, thanks for notice @MarcinVaadin. 
Conflict was fixed.

Can we merge it @jouni? 
(As I am not sure what language-unchecked means or we need some proof read from some other reviewer?)

jouni:
The `Language unchecked` label is used to let @russelljtdyer know that he should go through the changes, and check the English language, writing style, as well as the AsciiDoc formatting. You can merge a PR without having that check done, but leave the label there so that Russell can do the check retroactively.

PR #3887, opened 2024-11-07
[Control Center] Improve "manual installation" documentation

MatthewVaadin:
- [x] Renames "Manual Installation" page to "Installation of Dependencies".
- [x] Changes the dependency installation commands to use Helm.
- [x] Moves the admonitions in the "Getting Started" to be above the CC Helm commands.

The Keycloak commands need to remain in the documentation until we update the Helm charts to include the Keycloak Operator (https://github.com/vaadin/control-center/issues/721).

Fixes https://github.com/vaadin/control-center/issues/720

### DIFF
--- a/articles/control-center/getting-started/dependency-installation.adoc
+++ b/articles/control-center/getting-started/dependency-installation.adoc
@@ -1,25 +1,27 @@
 ---
-title: Manual Installation of Dependencies
+title: Installation of Dependencies
 description: Learn how to manually install Control Center's dependencies.
 order: 20
 ---
 
 
-= Manual Installation
+= Installation of Dependencies
 
 This guide explains how to install the required dependencies of Control Center without utilizing cluster-wide administrator permissions.
-
-Control Center must already be installed in your cluster before you can proceed with the installation of dependencies. If you haven't done this yet, see the <<index#,Getting Started>> guide for instructions.
 
 
 == NGINX Ingress Controller
 
-https://docs.nginx.com/nginx-ingress-controller[NGINX] is an ingress controller that manages external access to services in your cluster. You can install it by running the following command:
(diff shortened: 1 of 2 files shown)


articles/control-center/getting-started/dependency-installation.adoc
```diff
@@ -1,25 +1,27 @@
...
 
 .Terminal
 [source,bash]
 ----
-kubectl apply -f https://raw.githubusercontent.com/kubernetes/ingress-nginx/controller-v1.11.3/deploy/static/provider/cloud/deploy.yaml
+helm upgrade --install ingress-nginx ingress-nginx \
+  --repo https://kubernetes.github.io/ingress-nginx \
+  --namespace ingress-nginx \
+  --create-namespace \
+  --version 4.11.3
```
tamasmak, line 24: Should we add the `v` before the version?
```suggestion
  --version v4.11.3
```

tamasmak (commented):
I have a feeling that this [sentence](https://github.com/vaadin/docs/pull/3887/files#diff-15dda06831f3b89eebec430141add4ee277dd82bb29bfec6b0deee37ce769397R22) is standing alone here, there's no any command right after it. Although, it mentions that Control Center is going to be deployed to Kubernetes cluster, which is good I think, just maybe it could be placed somewhere else. What do you think?

articles/control-center/getting-started/dependency-installation.adoc
heruan: The `v` should not be needed.

tamasmak (approved):
Looks good. I tested the new Helm commands and in case the CloudbNativePG is installed with the new Helm command, Control Center requires some rules (see [vaadin/control-center#723](https://github.com/vaadin/control-center/issues/723)) updates to let the setup wizard to finish the install without issues. 
I would not merge this PR until the above mentioned changes are not released.

PR #3889, opened 2024-11-08
feat: Web Push Options

mshabarov:
Documents https://github.com/vaadin/flow/issues/20285.

### DIFF
--- a/articles/flow/configuration/setting-up-webpush.adoc
+++ b/articles/flow/configuration/setting-up-webpush.adoc
@@ -293,4 +293,77 @@ Also, Safari needs the web push notification features enabled. To do this, go to
 Mobile devices require the site to be served through `https` with a TLS/SSL certificate or they won't accept the service worker.
 ====
 
+=== Apply Custom Settings To Notifications
+
+A notification can display additional information, such as an icon, custom actions in a dropdown, or an arbitrary data object.
+The available options are documented https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification[here].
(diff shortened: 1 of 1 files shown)


articles/flow/configuration/setting-up-webpush.adoc
```diff
@@ -293,4 +293,77 @@ Also, Safari needs the web push notification features enabled. To do this, go to
...
+[source,java]
+----
+public record WebPushAction(String action, String title) implements Serializable {
+}
+public record WebPushOptions(String body,
+                             List<WebPushAction> actions,
+                             Serializable data,
+                             String icon) implements Serializable {
+}
+----
```
caalador, line 327: I would perhaps have these in the reverse order so that the record definitions are shown first to not create confusion on where these are available.
So say for the following records [sample records] they can be used as [sample creation]....

caalador (changes requested):

articles/flow/configuration/setting-up-webpush.adoc
mshabarov: I agree, done.

caalador (approved):

jouni:
Could we still rename the file from `setting-up-webpush.adoc` to `webpush.adoc`? Do we foresee the need to reserve that path for some other web-push-related content?

mshabarov:
`webpush.adoc` can be used instead, I see no problems.

jouni:
Thanks. I’ll update it. I assume there shouldn't be too many links to this pages out in the wild, so we won't need any redirect rules for it.

PR #3925, opened 2024-11-19
chore: Remove WebPush feature flag and use WebPushSubscription

mshabarov:
Updates documentation accordingly https://github.com/vaadin/flow/pull/20484.

### DIFF
--- a/articles/flow/configuration/webpush.adoc
+++ b/articles/flow/configuration/webpush.adoc
@@ -10,18 +10,6 @@ description: How to setup, subscribe, and send Web Push notifications from a Vaa
 
 This page explains how to configure a Vaadin Flow application to utilize Web Push. This can enable Web Push for a completed Customer Relationship Management (CRM) application so that users can receive notifications -- even when they're not accessing the application.
 
-[WARNING]
-.Experimental Feature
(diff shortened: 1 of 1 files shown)


articles/flow/configuration/webpush.adoc
```diff
@@ -168,22 +157,22 @@ public class WebPushService {
...
+        endpointToSubscription.put(subscription.endpoint(), subscription);
     }
 
 
-    public void remove(Subscription subscription) {
-        getLogger().info("Unsubscribed {}", subscription.endpoint);
-        endpointToSubscription.remove(subscription.endpoint);
+    public void remove(WebPushSubscription subscription) {
+        getLogger().info("Unsubscribed {}", subscription.endpoint());
+        endpointToSubscription.remove(subscription.endpoint());
```
caalador, line 175: `endpoint()` and `endpoint` are both valid for a record so I would say the main question is which should we use and then use the same way through all the documentation.

caalador (approved):

articles/flow/configuration/webpush.adoc
mshabarov: Record's fields are private and final by default, thus you cannot access them outside of the class, only with the getter methods.
We need `endpoint()` variant because of that.

PR #3936, opened 2024-11-20
chore: Example how to use arbitrary hidden component for setDragImage

mshabarov:
This documents a workaround while the https://github.com/vaadin/flow/pull/20490 is being investigated why it doesn't work on Mac.

### DIFF
--- a/articles/flow/create-ui/dnd/drag-source.adoc
+++ b/articles/flow/create-ui/dnd/drag-source.adoc
@@ -1,13 +1,13 @@
 ---
 title: Drag Source
-description: Making any component draggable, and configuring the drag operation.
+description: Making any component draggable and configuring the drag operation.
 order: 20
 ---
 
 
 = Creating a Drag Source
 
-With the [classname]`DragSource` class, you can make any component draggable by the user, and configure the drag operation. By default, the entire component is made draggable, instead of a small part of it. The [classname]`DragSource` class is a configuration object for the dragged item and contains static methods for configuring the given component instance.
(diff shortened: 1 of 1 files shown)


articles/flow/create-ui/dnd/drag-source.adoc
```diff
@@ -175,7 +175,22 @@ Any optional coordinates define the offset of the pointer location from the top
...
+
+*Example*: Use off-screen component as a drag image.
+[source,java]
+----
+Span dragImage = new Span("Drag Image Component");
+Style dragImageStyle = dragImage.getElement().getStyle();
+dragImageStyle.setPosition(Style.Position.ABSOLUTE);
+dragImageStyle.setTop("-100px");
+dragImageStyle.setLeft("-100px");
+dragImageStyle.setDisplay(Style.Display.NONE);
```
caalador, line 191: This part won't work without the patch that adds the display style change on drag start/end.
Offset image need to be visible and when having multiple images they should not be offset to the same spot.

mshabarov: Oh

mshabarov: Removed display none.

caalador (approved):

caalador (commented):
Looks ok, but now with the feature in you can in the sample have `display: hidden` if you want as it will be toggled automatically.

caalador (approved):

mshabarov:
Sure, added it back.

PR #3938, opened 2024-11-20
feat: add fine-grained security section

taefi:
This covers fine-grained access control via adding operation validators.


cromoteca (changes requested):
Even if that can be learnt by example, I'd give a short definition of the `OperationValidator` API and its `ValidationResult`, to make clear that there are only two possible outcomes.

cromoteca (approved):

PR #3946, opened 2024-11-25
docs: enhance dialog component documentation with positioning and sizing details

DiegoCardoso:
Create new "Position" and "Size" sections and move "Draggable" and "Rezizable" sections to the corresponding ones. 

Also add code snippets showing the new API for programmatically sizing and positioning.

### DIFF
--- a/articles/components/dialog/index.adoc
+++ b/articles/components/dialog/index.adoc
@@ -279,7 +279,7 @@ Modal dialogs don't benefit from being draggable, as their modality curtain (the
 
 == Size
 
-The Dialog size can be set with the width and height on the Dialog itself (__since V24.6 for React and Lit__). You can also set the size of the content of Dialog, whereby the Dialog scales to accommodate it In both cases, the Dialog can also be made resizable.
+The Dialog size can be set with the width and height on the Dialog itself (__since V24.6 for React and Lit__). You can also set the size of the content of Dialog, whereby the Dialog scales to accommodate it. In both cases, the Dialog can also be made resizable.
 
 [.example]
 --


articles/components/dialog/index.adoc
```diff
@@ -199,7 +199,45 @@ endif::[]
 
 Usually, non-modal dialogs should be draggable, so that the user can move them to access the user interface beneath.
 
-== Draggable
+=== Position
```
vursen, line 202:
```suggestion
== Position
```

Otherwise it doesn't appear in the navigation menu:

image

articles/components/dialog/index.adoc
```diff
@@ -199,7 +199,45 @@ endif::[]
...
+
+Dialogs can be positioned in two ways:
+
+. [since:com.vaadin:vaadin@V24.6]##By setting fixed coordinates programmatically##
+. By allowing the user to drag them (see <<Draggable>>)
```
vursen, line 207: This gave me the impression that these two ways are mutually exclusive, but that's not the case, if I understand correctly. You can set the initial position using the `top` and `left` properties and still enable dragging, right? The same goes for resizing.

DiegoCardoso: Yeah, both cases can work at the same time. I don't have the same impression, but I can be a bit biased. I can try to improve it to remove any misunderstanding.

articles/components/dialog/index.adoc
DiegoCardoso: Thanks!

articles/components/dialog/index.adoc
rolfsmeds: Not sure if I'm getting a mutually exclusive impression, but it does feel a bit strange to present those as two parallel options. How about something like this instead?

```suggestion
By default, dialogs open in the center of the viewport. [since:com.vaadin:vaadin@V24.6]##A different positioning can be set programmatically##. Dialogs can also be made draggable, allowing the end user to move them around.
```

rolfsmeds: (not sure I got the version tag syntax right there, so please check)

articles/components/dialog/index.adoc
```diff
@@ -247,7 +285,46 @@ For example, a Dialog for taking notes or for adding widgets to a dashboard by d
 
 Modal dialogs don't benefit from being draggable, as their modality curtain (the dark overlay behind the dialog) obscures the underlying user interface.
 
-== Resizable
+== Size
+
+The Dialog size can be controlled in two ways:
+
+. By setting fixed dimensions using the width and height properties (__since V24.6 for React and Lit__)
+. By making the dialog resizable (see <<Resizable>>)
```
rolfsmeds, line 293: A bit same as above, feels like resizability is presented as an alternative. Also, I think it's best to mention content-based sizing already at the top, like so:

```suggestion
The Dialog size can be set in two ways:

. By setting width and height on the Dialog itself (__since V24.6 for React and Lit__)
. By setting the size of the Dialog's contents, whereby the Dialog scales to accommodate it

In both cases, the Dialog can also be made resizable.
```

articles/components/dialog/index.adoc
```diff
@@ -247,7 +285,46 @@ For example, a Dialog for taking notes or for adding widgets to a dashboard by d
...
+ifdef::react[]
+[source,html]
+----
+<source-info group="React"></source-info>
+<Dialog width="400px" height="200px">...</Dialog>
+----
+endif::[]
+--
+
+The dialog's default size adjusts automatically based on its content. Fixed dimensions could be used when you need precise control over the dialog's size, for example, when displaying a form or grid with specific space requirements.
```
rolfsmeds, line 325: Then we can remove this paragraph.

articles/components/dialog/index.adoc
```diff
@@ -247,7 +285,46 @@ For example, a Dialog for taking notes or for adding widgets to a dashboard by d
...
+
+=== Resizable
 
 A resizable dialog allows the user to resize the Dialog by dragging from the edges of the Dialog with a pointing device.
 Dialogs aren't resizable by default.
```
rolfsmeds, line 330: Let's rewrite this weird old sentence while we're at it 😅 

```suggestion
The Dialog can be configured to allow the end user to resize it by dragging from its edges.
```

rolfsmeds (changes requested):

articles/components/dialog/index.adoc
DiegoCardoso: In this case, the paragraph about content-based sizing after the code snippets should be removed, right?

DiegoCardoso: NVM. I just saw the comment below.

articles/components/dialog/index.adoc
DiegoCardoso, line 325:
```suggestion
```

articles/components/dialog/index.adoc
DiegoCardoso: It's working:
image

rolfsmeds (approved):

articles/components/dialog/index.adoc
```diff
@@ -242,18 +274,51 @@ include::{root}/frontend/demo/component/dialog/react/dialog-draggable.tsx[render
...
 
 Modal dialogs don't benefit from being draggable, as their modality curtain (the dark overlay behind the dialog) obscures the underlying user interface.
 
-== Resizable
 
-A resizable dialog allows the user to resize the Dialog by dragging from the edges of the Dialog with a pointing device.
-Dialogs aren't resizable by default.
+== Size
+
+The Dialog size can be set with the width and height on the Dialog itself (__since V24.6 for React and Lit__). You can also set the size of the content of Dialog, whereby the Dialog scales to accommodate it In both cases, the Dialog can also be made resizable.
```
DiegoCardoso, line 284: Missing dot:
```suggestion
The Dialog size can be set with the width and height on the Dialog itself (__since V24.6 for React and Lit__). You can also set the size of the content of Dialog, whereby the Dialog scales to accommodate it. In both cases, the Dialog can also be made resizable.
```

PR #3947, opened 2024-11-25
Add banner at the end of the toc section

marcushellberg:
TODO:

- [ ] Update image
- [ ] Fix how the banner is displayed on a wide viewport


jouni (approved):
I would’ve expected that you’d want the banner to be visible even if the TOC is collapsed on small viewport sizes. But if this is the behavior you want, I don't have any issue with it.

PR #3949, opened 2024-11-26
docs: upload folder drag and drop

sissbruecker:
Document that upload now supports drag and drop of folders.

### DIFF
--- a/articles/components/upload/index.adoc
+++ b/articles/components/upload/index.adoc
@@ -69,7 +69,7 @@ include::{root}/frontend/demo/component/upload/react/upload-drag-and-drop.tsx[re
 endif::[]
 --
 
-Upload supports uploading folders via drag & drop. This uploads all files from the folder and its sub-folders recursively. Note that the folder structure will be flattened and the names of uploaded files will not indicate in which folder they were in. Uploading folders is currently not supported through the native file picker.
+Upload supports uploading folders via drag & drop. This uploads all files from the folder and its sub-folders recursively. The folder structure is flattened and the names of the uploaded files don't indicate in which folder they were in. Uploading folders is currently not supported through the native file picker.
 
 [.example]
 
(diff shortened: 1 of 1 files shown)


articles/components/upload/index.adoc
```diff
@@ -69,6 +69,9 @@ include::{root}/frontend/demo/component/upload/react/upload-drag-and-drop.tsx[re
 endif::[]
 --
 
+Upload also allows to drag & drop folders. This recursively uploads all files from the folder and its sub-folders. Note that the folder structure will be flattened and the names of uploaded files will not indicate in which folder they were in. Uploading folders is currently only supported when using drag & drop, but not when using the file picker.
```
rolfsmeds, line 72:
```suggestion
Upload can be configured to allow drag & drop uploading of folders. This uploads all files from the folder and its sub-folders recursively. Note that the folder structure will be flattened and the names of uploaded files will not indicate in which folder they were in. Uploading folders is currently only supported by drag & drop, not through the native file picker opened through the upload button.
```

rolfsmeds (changes requested):

articles/components/upload/index.adoc
sissbruecker: @rolfsmeds We don't have a config option for this, it just works out of the box.

rolfsmeds: Ah, indeed, lemme rephrase it.

```suggestion
Upload supports uploading folders via drag & drop. This uploads all files from the folder and its sub-folders recursively. Note that the folder structure will be flattened and the names of uploaded files will not indicate in which folder they were in. Uploading folders is currently not supported through the native file picker.
```

sissbruecker: Thanks!

rolfsmeds (approved):

PR #3974, opened 2024-12-03
[Control Center] Add a page detailing the options for the helm command

MatthewVaadin:
This adds a page explaining the options that can be used when installing Control Center using the helm command.

Fixes https://github.com/vaadin/control-center/issues/759

### DIFF
--- a/articles/control-center/getting-started/helm-values.adoc
+++ b/articles/control-center/getting-started/helm-values.adoc
@@ -1,0 +1,157 @@
+---
+title: Helm Values
+description: Learn how to configure Control Center using Helm values.
+order: 30
+---
+
+
+= Helm Values
+
+This page explains how to configure Control Center using Helm values. The following table lists the available values and their descriptions. You can set these values when you install Control Center using Helm. Below is an example where the domain, user email (both required), and application host (optional) are set:
+
+.Terminal
+[source,bash]
+----
+helm install control-center oci://docker.io/vaadin/control-center \
+    -n control-center --create-namespace \
+    --set domain=mydomain.com \
+    --set user.email=example@mydomain.com \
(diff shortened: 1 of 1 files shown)


heruan (approved):

articles/control-center/getting-started/configure-installation.adoc
```diff
@@ -0,0 +1,173 @@
...
+----
+
+More command options are available for configuring the installation, which are detailed in the Helm documentation https://helm.sh/docs/helm/helm_install/[here].
+
+
+== Values
+
+The following table lists the available values and their descriptions. Each of the values can be set using the `--set` flag when installing Control Center using Helm.
+
+pass:[<!-- vale off -->]
```
jouni, line 38: Why disable Vale completely? Which rules are causing trouble?

MatthewVaadin: There were at least three rules. The primary one was the spelling, since the values column contains many "words" that are not in the dictionary.  There was also `Vale.Terms` for using things like "keycloak" instead of "Keycloak", and `Vaadin.Abbr` for using "TXT", which in this context is not short for "text".

jouni: In case you’re interested, here’s how I fixed the issues, without disabled all Vale rules: https://github.com/vaadin/docs/commit/e5e6af858274aa9a5afedd76488ebba4965035fb

I needed to add two words to the dictionary (keystore and liveness, which apparently are common in our context). The other major change is to render the two table columns with monospace font, which contain programming code.

I ended up editing the page a bit as well, adding another section heading to make it easier to parse the page quickly, and annotating the code example with callouts.

PR #3977, opened 2024-12-04
[Control Center] Add pages for Control Center 1.1 pre-release

MatthewVaadin:
This adds temporary pages for the production and local installation of Control Center 1.1.

Note that the production installation page will be removed but the local installation page will be reused for the released version.


heruan (approved):

jouni:
@tomivirkki, did you remove the `Language unchecked` label on purpose?

jouni:
I think I’ve said this at some point, that I completely trust that Matthew is capable of writing proper English, so it’s not about that. But that label is also partially used to make sure @russelljtdyer (or someone else, perhaps me) takes a look at formatting and other details. Plus, it never hurts to have another pair of eyes check the English language as well.

jouni:
Sure, I understand the content was needed for the hackathon, and merging was very much the right thing to do. But I was just thinking was there are reason to remove the `Language unchecked` label. If not, I’ll add it back, so that Russell can maybe take another look.

tomivirkki:
Could be added back

MatthewVaadin:
@jouni This has been superseded by several other PRs that have been language checked.

#3965
#3976
#4027

PR #3983, opened 2024-12-06
docs: dashboard editing example

sissbruecker:
Adds dashbard example for an editable dashboard that allows persisting and loading its configuration.


frontend/demo/component/dashboard/react/dashboard-editable.tsx
tomivirkki, line 1: If I open the example on a fresh new browser, only the Flow example has content by default. The React and Lit examples are empty until I click "Restore default"

sissbruecker: Fixed by returning `null` from the mock dashboard service.

frontend/demo/component/dashboard/dashboard-editable.ts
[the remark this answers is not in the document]
sissbruecker: It seems like this assumption isn't true for the Lit version:
https://github.com/vaadin/docs/blob/a24ab8a8294d423e2cd652593e6c365f55798d40/frontend/demo/component/dashboard/dashboard-editable.ts#L49-L51

frontend/demo/component/dashboard/react/dashboard-editable.tsx
```diff
@@ -0,0 +1,181 @@
...
+  ];
+
+  return (
+    <>
+      <MenuBar
+        theme="dropdown-indicators"
+        items={menuItems}
+        onItemSelected={(e) => (e.detail.value as CustomMenuItem).action?.()}
+      />
+      <Dashboard
```
tomivirkki, line 129: Item type missing: `<Dashboard<WidgetConfig>`

sissbruecker: It looks like the type is already inferred from `items`, otherwise expressions such as `item.type` wouldn't compile.

tomivirkki: True, so it seems

frontend/demo/component/dashboard/react/dashboard-editable.tsx
```diff
@@ -0,0 +1,181 @@
...
+            // This function is used to render the actual widgets into the dashboard.
+            // It is called by Dashboard once for each config in the widgets array
+            // and should return a React element. In this example all widget types
+            // have the same content, so we can use generic logic to render a widget.
+            <DashboardWidget
+              widgetTitle={widgetTitles[item.type]}
+              style={{
+                '--vaadin-dashboard-item-colspan': item.colspan?.toString() ?? '1',
+                '--vaadin-dashboard-item-rowspan': item.rowspan?.toString() ?? '1',
+              }}
```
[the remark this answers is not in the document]
sissbruecker: Thanks, removed. Also elaborated in the comment that this happens automatically.

frontend/demo/component/dashboard/dashboard-editable.ts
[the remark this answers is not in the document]
sissbruecker: I've added listeners for all modifications to update the widgets state, which solves the issue.

As discussed internally, there is a separate issue with widgets losing focus in move mode, but it doesn't look like it's an issue with the example setup.

frontend/demo/component/dashboard/dashboard-editable.ts
```diff
@@ -0,0 +1,209 @@
...
+  renderDashboard() {
+    return html`
+      <vaadin-dashboard
+        style="--vaadin-dashboard-col-min-width: 150px; --vaadin-dashboard-col-max-count: 3;"
+        .editable="${this.editable}"
+        .items="${this.widgets}"
+        .renderer="${this.renderWidget}"
+        @dashboard-item-moved="${(e: DashboardItemMovedEvent<WidgetConfig>) => {
+          // Store updated widgets after user has modified them
+          this.widgets = e.detail.items as WidgetConfig[];
```
sissbruecker, line 159: Used a cast here so that the example doesn't have to deal with the `Array<DashboardSectionItem | DashboardItem>` type. This needs to happen somewhere in this component, otherwise we'd also need to deal with this structure in the Java service used in the example. Not great, but having to deal with sections in an example that doesn't use them seems worse?

frontend/demo/component/dashboard/react/dashboard-editable.tsx
```diff
@@ -0,0 +1,189 @@
...
+    widgets.value = [...widgets.value, { ...defaultWidgetConfig }];
+  }
+
+  function restore() {
+    // To restore defaults, we just set a copy of the default configuration
+    widgets.value = [...defaultConfig];
+  }
+
+  // Render function should be memoized to avoid unnecessary re-renders
+  const renderWidget = useCallback(({ item }: DashboardReactRendererProps<WidgetConfig>) => {
```
sissbruecker, line 94: As discussed, put the renderer function into a `useCallback` to avoid re-rendering widgets.

frontend/demo/component/dashboard/react/dashboard-editable.tsx
```diff
@@ -0,0 +1,189 @@
...
+// Helper type to allow defining a custom action for a menu item
+type CustomMenuItem = MenuBarItem & {
+  action?(): unknown;
+};
+
+function Example() {
+  useSignals(); // hidden-source-line
+  // Stores the current dashboard configuration. The Dashboard component will
+  // modify this array in place when editing, so there is no need to update it
+  // using events.
```
tomivirkki, line 52: Should be removed

sissbruecker: Done

tomivirkki (approved):

PR #4005, opened 2024-12-16
Replace old hilla.dev links from v23

veerapaananen:
Old hilla.dev links redirect to vaadin.com now. This fix replaces those links in the docs for version 23 with the final addresses they currently redirect to, removing the old links pointing to hilla.dev and removing redirections.

Example:

Old link: https://hilla.dev/
New link: https://vaadin.com/hilla


peholmst:
V23 Hilla links should not point to the latest version of Hilla, but the to the version that works with Vaadin 23. Unfortunately that documentation is currently not available anywhere.

veerapaananen:
Those links redirect now to the latest documentation anyway. Should the links be removed instead? @peholmst

peholmst:
Given the situation right now, they should be removed. But if we end up publishing the documentation for the old Hilla versions somewhere, they would have to be added back.

veerapaananen:
I'll do some changes and update this PR. Let's add the links back if/when the old Hilla docs are published.

peholmst (approved):

Issue #3969, opened 2024-12-02
Routing docs should explicitly warn about duplicate routes


Legioth:
The routing docs don't seem to mention that routes need to be unique.
Since it's quite common to end up with duplicate routes by mistake, especially when playing around with the platform and copypasting examples from docs, I think we should be very clear about it.
https://vaadin.com/docs/latest/flow/routing/route

PR #4070, opened 2025-01-15
docs: add note about nav target path uniqueness

tepi:
Fixes https://github.com/vaadin/docs/issues/3969


articles/flow/routing/route.adoc
```diff
@@ -67,5 +67,8 @@ public class HomeView extends Div {
 }
 ----
 
+[NOTE]
+All navigation target paths in the application must be unique to allow Flow to determine the view to render without any disambiguation. This affects all [annotationname]`@Route`, [annotationname]`@RouteAlias` and [annotationname]`@RoutePrefix` values. When conflicts are detected, the application fails to start.
```
caalador, line 71: Should we point out that it is Route and RouteAlias combined with any RoutePrefix?

caalador: Should this mention that also client side paths are evaluated with the server side routes?

caalador (approved):

PR #4077, opened 2025-01-21
feat: explain how to make disabled buttons accessible

vursen:
Depends on
- https://github.com/vaadin/flow-components/pull/7050

Fixes https://github.com/vaadin/web-components/issues/4585


articles/components/button/index.adoc
```diff
@@ -392,6 +392,27 @@ include::{root}/frontend/demo/component/button/react/button-disabled.tsx[render,
 endif::[]
 --
 
+=== Accessibility
+
+While the default behavior effectively prevents accidental interactions, it has an accessibility drawback: screen readers skip disabled buttons entirely, and users can't see tooltips that might explain why the button is disabled. To address this, an experimental enhancement allows disabled buttons to receive focus and show tooltips, improving their accessibility. This feature can be enabled with the following feature flag:
```
vursen, line 397:
```suggestion
While the default behavior effectively prevents accidental interactions, it has an accessibility drawback: screen readers skip disabled buttons entirely, and users can't see tooltips that might explain why the button is disabled. To address this, an experimental enhancement allows disabled buttons to receive focus and show tooltips, while still preventing other interactions. This feature can be enabled with the following feature flag:
```

rolfsmeds (approved):

Issue #3933, opened 2024-11-20
Finding info about how to build forms is difficult if you don't know our terminology


Legioth:
HTML and most frameworks call forms forms, we talk about binder, data binding. These are not as findable as terms

PR #4082, opened 2025-01-22
docs: make Forms more prominent in documentation

tepi:
Fixes https://github.com/vaadin/docs/issues/3933

### DIFF
--- a/articles/flow/binding-data/components-binder.adoc
+++ b/articles/flow/binding-data/components-binder.adoc
@@ -1,6 +1,6 @@
 ---
 title: Binding Data to Forms
-page-title: How to bind data to forms with field components using binder
+page-title: How to bind data to forms using binder with Vaadin Flow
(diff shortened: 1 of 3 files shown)


articles/flow/binding-data/components-binder.adoc
```diff
@@ -1,6 +1,6 @@
 ---
 title: Binding Data to Forms
-page-title: How to bind data with components using binder | Vaadin
+page-title: How to bind data to forms with field components using binder
```
jouni, line 3: AFAIK the page title should always include "Vaadin", either as a part of the title itself, or as a "suffix" as was before this change.

jouni: https://vaadin.com/docs/latest/contributing/docs#meta-title

tepi: Alright, I'll add it back in some form. But I do think we have a bunch of other pages missing Vaadin as well.

PR #4094, opened 2025-01-28
fix: Update parameters

### LINKED ISSUE #4041: Clarify Vaadin config parameters vs Vaadin modes
**Tell us about your request**

Vaadin can operate in different modes and exposes a lot of configuration parameters.
Despite the modes and parameters are already described, it's sometimes not clear or problematic:

- for which mode/configuration a particular parameter is valid (dev or prod mode, pre-compiled bundles or hot-deploy with Vite, routing with react or with vaadin router), or is it valid for all possible modes of running and configuring an application.
- which parameter could fix what problem, e.g. when one would need to force the prod bundle or disable bundle optimisation.

We may want to have:
- a kind of mapping / table that could list all relevant parameters per mode.
- troubleshooting / FAQ article that recommends a parameter value change vs an example of possible problem.

---

If you think this issue is important, add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

caalador:
Update missing parameters and
add to the table mode the
parameter targets and
system property written out
for clarity

Closes #4041


articles/flow/configuration/properties.adoc
```diff
@@ -95,156 +95,359 @@ When a system property and a servlet parameter have the same name, the system pr
...
 
 == Vaadin Plugin Properties
 
 The following table contains the properties that are used only by the Vaadin Maven and Gradle Plugin, and are not applicable for deployment configuration:
 
-[cols="1,1,3,1"]
+[cols="1,1,4,1,1"]
 |===
-|System Property
 |Plugin Configuration
```
mshabarov, line 368: This table misses many properties that applicable to plugin, e.g. `nodeVersion`, `pnpmEnable` etc. Shall we address this separately or here?

articles/flow/configuration/properties.adoc
```diff
@@ -95,156 +95,359 @@ When a system property and a servlet parameter have the same name, the system pr
 [[properties]]
 == Properties Defined
 
-The following table contains the properties that are defined in the [classname]`com.vaadin.server.DeploymentConfiguration` and [classname]`com.vaadin.flow.server.Constants` classes. They're listed in alphabetical order. If you use Spring Boot, you should add the `vaadin.` prefix to them (e.g., change `brotli` to `vaadin.brotli`).
+The following table contains the properties that are defined in the [classname]`com.vaadin.server.DeploymentConfiguration`, [classname]`com.vaadin.flow.server.Constants` and [classname]`com.vaadin.flow.server.InitParameters` classes. They're listed in alphabetical order. If you use Spring Boot, you should add the `vaadin.` prefix to them (e.g., change `brotli` to `vaadin.brotli`).
```
mshabarov, line 98: Could we give a short definition of each mode? `DAU` and `Runtime` may cause questions I think.

articles/flow/configuration/properties.adoc
```diff
@@ -95,156 +95,359 @@ When a system property and a servlet parameter have the same name, the system pr
...
 |Disables cross-site request forgery protection. The protection is enabled by default. You should keep it enabled -- except for certain types of testing.
+|`false`
+|Runtime
 
 |`eagerServerLoad`
-|`false`
+|`vaadin.eagerServerLoad`
 |Enables the client-side bootstrap page to include the initial JSON data fragment.
+|`false`
+|Runtime
```
mshabarov, line 208: This property is applicable for development as well.
I haven't seen two modes for a single parameter in these tables, but I bet a subset of parameters have effect in both modes.
Shall we double check and `Development, Runtime` ?

mshabarov:
How about "troubleshooting / FAQ article that recommends a parameter value change vs an example of possible problem." ? 
Shall we do it in a separate pull request? I think it's still needed.

articles/flow/configuration/properties.adoc
caalador: added. DAU dropped as it is more a runtime parameter.

articles/flow/configuration/properties.adoc
```diff
@@ -95,156 +95,359 @@ When a system property and a servlet parameter have the same name, the system pr
...
 |Enables live reload. When using a server-side <<live-reload/index#, live reload tool>>, the browser is refreshed after code is rebuilt on the server. Set to `false` to disable automatic reloading of the browser. This applies only to development mode.
+|`true`
+|Development
 
 |`devmode.optimizeBundle`
-|`false`
+|`vaadin.devmode.optimizeBundle`
 |Optimizes frontend resource bundles. All frontend resources in the classpath are included by default in the generated bundle in development mode. When set to `true`, the frontend build creates an optimized bundle by including only frontend resources that are used from the application entry points. It uses bytecode scanning, which increases application start-up time. Set to `false` to skip the optimization in production mode.
+|`false`
+|Development
```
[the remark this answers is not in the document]
caalador: Changed to build.
Yes should have different naming.

articles/flow/configuration/properties.adoc
caalador: Runtime is for application execution indifferent to what mode dev/prod it is running in.

articles/flow/configuration/properties.adoc
caalador: Those are in the properties table as `node.version`, `pnpm.enable`
This explicitly says that `properties that are only used by the plugins`

articles/flow/configuration/properties.adoc
```diff
@@ -95,156 +95,359 @@ When a system property and a servlet parameter have the same name, the system pr
...
-|`vaadin.skip.dev.bundle`
 |`skipDevBundleRebuild`
-|Prevents a frontend development bundle from being re-built, even if Vaadin Flow decides to use an existing compiled development bundle. This is mainly needed when re-bundling checker in Vaadin Flow has problems leading to false re-bundling, and one needs a workaround while it's being resolved.
+|`vaadin.skip.dev.bundle`
+|Prevents a frontend development bundle from being re-built, even if Vaadin Flow decides to use an existing compiled development bundle. This is mainly needed when the re-bundling checker in Vaadin Flow has problems leading to false re-bundling, and one needs a workaround while it's being resolved.
 |`false`
+|Bundle
 
 |`vaadin.clean.build.frontend.files`
 |`cleanFrontendFiles`
```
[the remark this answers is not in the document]
caalador: Fixed, Just a missed change of the column.

articles/flow/configuration/properties.adoc
mshabarov: > Runtime is for application execution indifferent to what mode dev/prod it is running in.

This is a good point. Could we add it as a sentence to this article?

caalador: I added it in the mode text before the tables

caalador:
> How about "troubleshooting / FAQ article that recommends a parameter value change vs an example of possible problem." ? Shall we do it in a separate pull request? I think it's still needed, we can add the cases that we know at the moment and update that article once we have new cases.

Probably a separate PR would be good as this has quite a lot of changes already.

PR #4099, opened 2025-01-31
docs(hilla): remove section about plugin configuration

### LINKED ISSUE #4096: Hilla: Outdated documentations for Parser plugins and options
As of 24.7, there are no more options to set, for example, `exposedPackagesToParser` or `hilla` extension object for Gradle configurations.
The documentations still pointing to those configurations should be refined / updated to reflect the new reality. 

Since we had these possibilities before, is it better to mention how is it better now, or to clarify whether it is a missing feature which going to be added again or not.

I could spot the following places to update, but there might be more:

- https://vaadin.com/docs/latest/hilla/reference/endpoint-generator#adding-a-custom-generator-plugin
- https://vaadin.com/docs/latest/hilla/reference/gradle#_all_options

cromoteca:
Maven/Gradle plugin configuration is no longer available.

Inform users to always instantiate endpoints as Spring beans.

Also fix some other outdated stuff in the page.

Closes #4096.

### DIFF
--- a/articles/hilla/lit/reference/endpoint-generator.adoc
+++ b/articles/hilla/lit/reference/endpoint-generator.adoc
@@ -1,14 +1,18 @@
 ---
-title: Endpoint Generator
-page-title: How to use the endpoint generator in Hilla | Vaadin
-description: The endpoint generator produces TypeScript files based on the information from an OpenAPI document that's generated from Java (or other JVM language) files.
+title: TypeScript Generator
+page-title: How to use the TypeScript generator in Hilla | Vaadin
+description: The TypeScript generator produces TypeScript files that map to Java services annotated with [annotationname]`@BrowserCallable` or [annotationname]`@Endpoint`.
 meta-description: Discover how to generate endpoints for Hilla applications efficiently.
 ---
 // tag::content[]
 
-= Endpoint Generator
+= TypeScript Generator
 :toclevels: 2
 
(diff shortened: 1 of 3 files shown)


articles/hilla/lit/reference/endpoint-generator.adoc
```diff
@@ -1,14 +1,18 @@
...
 // tag::content[]
 
-= Endpoint Generator
+= TypeScript Generator
 :toclevels: 2
 
+The TypeScript generator produces TypeScript files that map to Java services annotated with [annotationname]`@BrowserCallable` or [annotationname]`@Endpoint`.
+Those files retain the properties and types of the Java classes to provide type safety in the frontend.
+
+Those annotated services are commonly named *endpoints*, although there is a single REST endpoint in Hilla, which in turn calls the browser-callable services. The term "endpoints" is used in this context to refer to those annotated Java services.
```
taefi, line 15: Minor: This could be a `[NOTE]` for better visibility.

taefi (changes requested):
Asked for a minor change. Otherwise, looks good!

articles/hilla/lit/reference/endpoint-generator.adoc
cromoteca: I'm not sure about that. Also, I tend to respect https://vaadin.com/docs/latest/contributing/docs/styleguide#admonitions and try not to interrupt the page flow unless necessary.

taefi (approved):
LGTM.

PR #4103, opened 2025-02-04
feat: Card component

jouni:
Missing Java/Flow examples.


frontend/demo/component/card/card-combine-variants.ts
```diff
@@ -0,0 +1,35 @@
+import 'Frontend/demo/init'; // hidden-source-line
+import '@vaadin/avatar';
+import '@vaadin/card';
+import '@vaadin/icon';
+import '@vaadin/vaadin-lumo-styles/icons.js';
+import { html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
+import { applyTheme } from 'Frontend/generated/theme';
+// @ts-expect-error hidden-source-line
+import img from '../../../../src/main/resources/images/lapland.avif?url';
```
web-padawan, line 10: I think this should be fixed by adding the following to `frontend/types.d.ts`

```ts
declare module '*.avif?url' {
  const value: string;
  export = value;
}
```

Example for `.png`:

https://github.com/vaadin/docs/blob/a7bd7ebf22a68312102b248f98687ad84795f9c3/frontend/types.d.ts#L1

Then we can also remove `// hidden-source-file` comment for image imports.

articles/components/card/styling.adoc
```diff
@@ -0,0 +1,72 @@
...
+
+
+=== Parts
+
+Media:: `vaadin-card+++<wbr>+++**::part(media)**`
+Header:: `vaadin-card+++<wbr>+++**::part(header)**`
+Content:: `vaadin-card+++<wbr>+++**::part(content)**`
+Footer:: `vaadin-card+++<wbr>+++**::part(footer)**`
+
+=== Style Variants
```
web-padawan, line 68: Should we also list `cover-media` and `stretch-media` in the table below?

frontend/demo/component/card/react/card-header-suffix.tsx
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react'; // hidden-source-line
+import { Card } from '@vaadin/react-components/Card.js';
+
+function Example() {
+  return (
+    // tag::snippet[]
+    <Card>
+      <div slot="title">Lapland</div>
+      <div slot="subtitle">The Exotic North</div>
+      <span slot="header-suffix" theme="badge success">
```
web-padawan, line 11: Should use the following syntax to fix React example compilation:

```suggestion
      <span slot="header-suffix" {...{ theme: 'badge success' }} >
```

frontend/demo/component/card/card-stretch-media.ts
sissbruecker, line 1: The examples with icons show up like this for me:

Bildschirmfoto 2025-02-04 um 16 21 58

frontend/demo/component/card/card-combine-variants.ts
web-padawan: Added this block and removed `// @ts-expect-error` comments.

frontend/demo/component/card/react/card-header-suffix.tsx
web-padawan: Fixed.

articles/components/card/index.adoc
```diff
@@ -0,0 +1,480 @@
...
+++++
+<style>
+.example:not(.render-only) .render > * {
+  max-width: 300px;
+  margin: auto;
+}
+
+.example.card-grid .render > * {
+  max-width: fit-content;
+  display: grid !important;
```
web-padawan, line 471: These styles are only applied to examples on the Card page, but not when opening in a new window.
This can be unexpected for some users, would it make sense to move related CSS to examples?

jouni: Right, good point, I didn't realize that. Hmm, need to think about that. I thought it would be okay to have these invisible layout styles in the docs page, but I’m not sure about the standalone example pages.

frontend/demo/component/card/card-stretch-media.ts
jouni: Oh, wow. What browser?

articles/components/card/styling.adoc
jouni: Yeah, sure. I think I just forgot to add them. 

Btw, should we use "Style Variants" everywhere (vs "Theme Variants")?

frontend/demo/component/card/card-stretch-media.ts
sissbruecker: Ah, only in Firefox apparently.

jouni: Yeah, I see it too. I’ll investigate.

jouni: https://github.com/vaadin/web-components/issues/8641

articles/components/card/styling.adoc
web-padawan: IIRC we agreed to use "Style Variants" at some point, so let's use it for now.

articles/components/card/index.adoc
web-padawan: I think some users might actually wonder how one would build a "card list" and these styles especially `repeat(auto-fit, minmax(190px, 1fr));` could be beneficial. 

Alternatively we could just put a few utility classes if we don't want to cause distraction but still keep the examples working the same after copying to some Hilla view.

jouni: Yes, I agree. I’m updating the examples right now.

web-padawan (approved):

PR #4105, opened 2025-02-05
Update index.adoc

rkalkowski:
Replaced Tutorial box with the Vaadin Community (Forum)

### DIFF
--- a/articles/index.adoc
+++ b/articles/index.adoc
@@ -36,7 +36,7 @@ image:flow/_images/code-editor-illustration.svg["", opts=inline]
 [.card.large]
 === Vaadin Community
 
-Stuck on something? Need feedback on your implementation? The Vaadin Forum is the perfect place to ask questions, share ideas, and connect with other developers.
+If you’re stuck or need feedback on your implementation, the Vaadin forum is the perfect place to ask questions, share ideas, and connect with other developers.
 
 link:/forum/[Join the forum]
 
--- a/articles/index.adoc
+++ b/articles/index.adoc
@@ -34,11 +34,11 @@ image:flow/_images/code-editor-illustration.svg["", opts=inline]
 
 
 [.card.large]
-=== Tutorial
+=== Vaadin Community
 
-The Vaadin tutorial walks you through building an application from start to production. It covers both Flow and Hilla development.
+Stuck on something? Need feedback on your implementation? The Vaadin Forum is the perfect place to ask questions, share ideas, and connect with other developers.
 
-<<{articles}/getting-started/tutorial#,Start tutorial>>
+link:/forum/[Join the forum]
 
 
 [.card.large]


rkalkowski (commented):
no changes

articles/index.adoc
```diff
@@ -34,11 +34,11 @@ image:flow/_images/code-editor-illustration.svg["", opts=inline]
...
 
 [.card.large]
-=== Tutorial
+=== Vaadin Community
 
-The Vaadin tutorial walks you through building an application from start to production. It covers both Flow and Hilla development.
+Stuck on something? Need feedback on your implementation? The Vaadin Forum is the perfect place to ask questions, share ideas, and connect with other developers.
 
-<<{articles}/getting-started/tutorial#,Start tutorial>>
+link:/forum/[Join the forum]
```
jouni, line 41: Should it be **The Vaadin Forum** or **The Vaadin forum**? In the text, it’s the former, and in the link it’s the latter. Is that on purpose?

articles/index.adoc
```diff
@@ -34,11 +34,11 @@ image:flow/_images/code-editor-illustration.svg["", opts=inline]
 
 
 [.card.large]
-=== Tutorial
+=== Vaadin Community
 
-The Vaadin tutorial walks you through building an application from start to production. It covers both Flow and Hilla development.
+Stuck on something? Need feedback on your implementation? The Vaadin Forum is the perfect place to ask questions, share ideas, and connect with other developers.
```
jouni, line 39: In general, I’d avoid asking questions, and rather use direct statements. Is it good to force the reader to pause and think for an answer to these questions before going to the pitch about our forum? I know it’s subtle, but these question marks always trigger me :) 

```suggestion
If you’re stuck or need feedback on your implementation, the Vaadin forum is the perfect place to ask questions, share ideas, and connect with other developers.
```

articles/index.adoc
```diff
@@ -34,11 +34,11 @@ image:flow/_images/code-editor-illustration.svg["", opts=inline]
 
 
 [.card.large]
-=== Tutorial
+=== Vaadin Community
```
jouni, line 37: If we hide the CTA link, we can formulate this heading to be more actionable.

```suggestion
=== Join the Forum
```

```suggestion
=== Join the Community
```

articles/index.adoc
jouni, line 41: Maybe we could hide the CTA links on this card and the "video course" card as well, like on all the other cards on the landing page.

```suggestion
[.sr-only]
link:/forum/[Join the forum]
```

rkalkowski: I suggest keeping the CTA links, as it's not clear those boxes are clickable.

articles/index.adoc
rkalkowski: I agree. The suggested change makes sense.

jouni (approved):

PR #4125, opened 2025-02-13
docs: layout troubleshooting section

sissbruecker:
Adds a troubleshooting section to horizontal and vertical layout docs explaining common pitfalls and how to fix them. Also documents the layout improvements feature flag for Flow that will be available in 24.7.


articles/components/horizontal-layout/_troubleshooting.adoc
```diff
@@ -0,0 +1,301 @@
...
+  <div style={{ height: '200px' }}></div>
+  <div style={{ flexGrow: 1 }}></div>
+</VerticalLayout>
+----
+endif::[]
+--
+endif::[]
+
+[role="since:com.vaadin:vaadin@V24.7"]
+==== Enable Layout Improvements (Flow only, experimental)
```
web-padawan, line 192: Looks like this heading is repeated twice on the page which looks a bit unexpected at first glance. 
Should it be just a single entry with two bullet points instead?

sissbruecker: These are nested below the respective level 3 headings, so I think it's fine? Moving the section out of the respective sections for the problems that they are supposed to help solving seems also confusing.

web-padawan: Makes sense, thanks.

rolfsmeds (approved):

web-padawan (approved):

PR #4128, opened 2025-02-13
docs: add slot based horizontal alignment examples

web-padawan:
Part of https://github.com/vaadin/platform/issues/5038


articles/components/horizontal-layout/index.adoc
```diff
@@ -124,32 +124,44 @@ endif::[]
...
+[role="since:com.vaadin:vaadin@V24.7"]
+=== Aligning Individual Items
+
+Each item in the layout can be grouped to the start, center and end of the layout.
 
 [.example]
 --
 
 ifdef::lit[]
 [source,typescript]
```
sissbruecker, line 139: Nit: `[source,html]` works better here for the code snippet. But there's several examples like that, so maybe we can do that as a follow up.

sissbruecker (approved):

PR #4148, opened 2025-02-25
Use setItemsPageable in data binding docs

Artur-:


articles/flow/binding-data/data-provider.adoc
```diff
@@ -134,9 +134,36 @@ Currently, only `Grid` and `ComboBox` support lazy data binding. To use it, thou
...
+
+// Assuming a service like
+class ProductService {
+  public List<Product> list(Pageable pageable) { ... }
+}
+----
+
+The `setItemsPageable` method takes care of making the Grid request available in a Spring friendly format, as a `Pageable` instance. This `Pageable` instance contains information about which page (offset/limit) to load and also about how to sort the data (based on what is selected in the Grid). Many spring data sources use `Pageable` so you can directly pass it into e.g. a JPA repository class.
+
+If you Spring service requires additional parameters, you can use the longer version
```
mshabarov, line 156:
```suggestion
If your Spring service requires additional parameters, you can use the longer version
```

articles/flow/binding-data/data-provider.adoc
```diff
@@ -134,9 +134,36 @@ Currently, only `Grid` and `ComboBox` support lazy data binding. To use it, thou
...
+
+[source,java]
+----
+grid.setItemsPageable(pageable -> productService.list(pageable, somethingElse));
+----
```
mshabarov, line 161: Not sure what's the goal of adding this code snippet as it shows how one can modify the service method, whereas the same `setItemsPageable` is used here.

mshabarov: I'd replace `somethingElse` with a meaningful example, e.g. a full text search string filter.

mshabarov (approved):
In general looks good to me, nice work done! The only comment is about `somethingElse` example.

Artur-:
Something is better than nothing

Issue #2412, opened 2023-05-09
Document the vaadin.frontend.hotdeploy.dependencies feature


Artur-:
**Vote to prioritize:** add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

The feature introduced in https://github.com/vaadin/flow/pull/16732 would need some documentation in the live reload chapter. Should take https://github.com/vaadin/flow/pull/16738 into account also which automates the process for Maven multi module projects

PR #4174, opened 2025-03-07
docs: add frontend.hotdeploy.dependencies documentation

tepi:
Fixes https://github.com/vaadin/docs/issues/2412

Part of https://github.com/vaadin/flow/issues/20992


articles/flow/configuration/properties.adoc
```diff
@@ -225,6 +225,12 @@ Bundle parameter controls parts of frontend bundle creation.
 |`false`
 |Development
 
+|`frontend.hotdeploy.dependencies`
+|`vaadin.frontend.hotdeploy.dependencies`
+|A comma-separated list of module directories to watch for frontend file changes when detecting hot-deployable changes. Entries need to point to directories relative to the root project, and can point to the root project itself as well, in case it has frontend files as resources. The value could be for example "./,./component-module-1,./component-module-2". Flow will watch for changes in both `src/main/resources/META-INF/frontend` and `src/main/resources/META-INF/resources/frontend` directories under the given entries.
```
mcollovati, line 230:
```suggestion
|A comma-separated list of module directories to watch for frontend file changes when detecting hot-deployable resources. Entries need to point to directories relative to the root project, and can point to the root project itself as well, in case it has frontend files as resources. The value could be for example "./,./component-module-1,./component-module-2". Flow will watch for changes in both `src/main/resources/META-INF/frontend` and `src/main/resources/META-INF/resources/frontend` directories under the given entries.
```

mcollovati: I think `root project` can be a bit confusing; for example, it may suggest the path is relative to the parent project in a multi-module setup, instead of being relative to the module that starts the application server.
Can we reword it to make it more clear? For example, `relative to the module that starts the Vaadin application` or something similar?

mcollovati (changes requested):

mcollovati (approved):

articles/flow/configuration/properties.adoc
```diff
@@ -225,6 +225,12 @@ Bundle parameter controls parts of frontend bundle creation.
 |`false`
 |Development
 
+|`frontend.hotdeploy.dependencies`
+|`vaadin.frontend.hotdeploy.dependencies`
+|A comma-separated list of module directories to watch for frontend file changes when detecting hot-deployable resources. Entries need to point to directories relative to the module that starts the Vaadin application, and can point to the module itself as well, in case it has frontend files as resources. The value could be for example "./,./component-module-1,./component-module-2". Flow will watch for changes in both `src/main/resources/META-INF/resources/frontend` and the legacy `src/main/resources/META-INF/frontend` directories under the given directory entries.
```
mshabarov, line 230: Could we also mention that if this parameter is not defined, Vaadin watches `src/main/resources/META-INF/` folder of the active/main module by default, whereas, if you define this parameter, you have to add "./" to the value to have it watched, else it is not watched.

tepi: Good point, I added a note about this.

mshabarov (approved):

PR #4178, opened 2025-03-12
[Control Center] Improve the clarity of application deployment documentation

MatthewVaadin:
This expands the description of the image field and updates the corresponding screenshot.

Fixes https://github.com/vaadin/control-center/issues/852


articles/control-center/application-deployment/index.adoc
```diff
@@ -116,10 +116,10 @@ image::images/app-deploy.png[App Deployment]
 In the [guibutton]*Deploy* panel on the right-hand side (as shown in the screenshot below), fill in the following fields:
 
 - *Application Name*: Give your application a meaningful name (e.g., `My App`).
-- *Image*: Enter the tag of the Docker image (e.g., `my-app:dev`).
+- *Image*: Enter the name of the Docker image, including the repository name and version tag (e.g., `repository/app:version`). If your repository is private, you can also provide a secret name containing the credentials to access it by clicking *Needs Pull Secret*.
 - *Replicas*: Specify the number of instances to run. The default is `1`.
 - *Hostname*: Set the external URL to access your application (e.g., `https://app.example.com`). For more information on how to choose a proper hostname, refer to the <<hostname-guidelines#, Hostname Guidelines>>.
-- *Features*: Enable additional features like Identity Management and Localization if required.
+- *Features*: Enable additional features if required.
```
tamasmak, line 122: The `Features` section of the is also mentioned in this list but the new image doesn't contain that part like the old one. Shouldn't we show that also on the image?

MatthewVaadin: The new theme adds whitespace such that the Features section is now off the edge of the iPad screen.

I could add a second image with the panel scrolled down, or reduce the zoom level like in the following screenshot. Do you think it's readable?

control matthew vaadin cc_app_secure-app_continue(iPad Pro)

tamasmak: The size of the previous image also wasn't an iPad screen, it had a custom size. I think it's still readable in the docs. The image can be opened in a new tab/window and there it has the full size.

articles/control-center/application-deployment/index.adoc
```diff
@@ -116,10 +116,12 @@ image::images/app-deploy.png[App Deployment]
 In the [guibutton]*Deploy* panel on the right-hand side (as shown in the screenshot below), fill in the following fields:
 
 - *Application Name*: Give your application a meaningful name (e.g., `My App`).
-- *Image*: Enter the tag of the Docker image (e.g., `my-app:dev`).
+- *Image*: Enter the name of the Docker image, including the repository name and version tag (e.g., `repository/app:version`). If your repository is private, you can also provide a secret name containing the credentials to access it by clicking *Needs Pull Secret*.
```
tamasmak, line 119:
```suggestion
- *Image*: Enter the name of the Docker image, including the repository name and version tag (e.g., `repository/app:version`). If your repository is private, you can also provide a secret name containing the credentials to access it by enabling the [guilabel]*Needs Pull Secret* option.
```

tamasmak (changes requested):
Description about the certificate section is only missing from the list now. Did we skip that intentionally before?

MatthewVaadin:
> Description about the certificate section is only missing from the list now. Did we skip that intentionally before?

I don't think it was intentional. I've added a description for it now.

Issue #4154, opened 2025-02-27
Document the new API for fine-tuning the frontend build in Flow Maven plugin


mshabarov:
After these two [Configurable class loading in Flow Maven plugin](https://github.com/vaadin/flow/issues/21052) and [Add frontendIgnoreVersionChecks Vaadin Maven plugin parameter](https://github.com/vaadin/flow/issues/21049) are done, we need to document this API in online docs.

mcollovati:
`frontendIgnoreVersionChecks` is covered by https://github.com/vaadin/docs/pull/4173

PR #4188, opened 2025-03-19
docs: document maven plugin frontendScanner configuration

mcollovati:
Part of #4154

### DIFF
--- a/articles/flow/configuration/maven.adoc
+++ b/articles/flow/configuration/maven.adoc
@@ -119,6 +119,43 @@ Clears the generated frontend files after building a project for production. It 
 `frontendExtraFileExtensions`::
 Parameter for adding file extensions to a handle when generating bundles.
 
+`frontendScanner`::
+Allows to fine tune frontend resources detection by defining which artifacts should be included or excluded during the class scanning process.
++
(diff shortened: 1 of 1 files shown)


mcollovati:
Related to vaadin/flow#21052 and vaadin/flow#21125.
In draft, because the feature is still under review

mshabarov (changes requested):
Looks good in general, but few suggestions.
Already posted this on the Flow side pull request, but IMO these are important for docs too:
1. Explain when to use `includeOutputDirectory` - when main maven module has no Vaadin frontend resources or has no Vaadin at all, makes sense to exclude from scanning.
2. `includes` - includes what is listed and excludes everything else. Useful when you want to scan a bit and not scan other (say, scan 10% and skip 90%).
3. `exclude` - excludes what is listed and includes everything else. Useful when you want to scan everything except a few parts (say, scan 90% and skip 10%).
4. Optionally we can remind what kind of frontend we mean under scanning - JsModule, JavaScript, CssImport and NpmPackage annotations (what else, PWA, Layout?)
5. I'd reference this in the dev mode article, e.g. here https://vaadin.com/docs/latest/flow/configuration/development-mode#exclude-vaadin-components

mcollovati:
Applied some changes. However, I'm having troubles starting the publisher locally, so I could not double-check the formatting

mshabarov (approved):

mshabarov:
Looks good. Let's wait for a while until we release the feature in some pre-release.

mshabarov:
This is available in 24.8.0.alpha3 now.

PR #4196, opened 2025-03-21
docs: add Control Center Troubleshooting section

sosa-vaadin:
Add a "Troubleshooting" section for Control Center documentation.

Fixes https://github.com/vaadin/control-center/issues/848


articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+
+= Troubleshooting
+
+When installing Control Center or deploying an application using Control Center, you may encounter issues that are not caused by Control Center itself, but by external factors such as cluster resource limits, DNS propagation delays, or misconfigured dependencies.
+
+
+== Insufficient Cluster Resources
+
+When Control Center instance is slow to respond.
```
heruan, line 17: This doesn't sound like a finished sentence. Let's be more discursive and keep a writing style consistent with the other articles in the docs. There's a lot we can add to help users in trouble: insufficient cluster resources could also stop the app from start at all, probes to fail repeatedly, etc.

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+=== Solution
+
+- Check resource allocation:
+
+[source,shell]
+----
+kubectl describe node
+----
```
heruan, line 26: What the user should expect from the output of this command? What should they look for? Let's provide output example and help the user identify possible issues.

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+- Increase CPU/memory limits.
```
heruan, line 28: What could the user do to increase the limits? What about the node specs?

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+== DNS Record Propagation Delays
+
+When Control Center's URL or a deployed application's URL is not resolving after being deployed.
```
heruan, line 32: We can't assume the user knows it all, let's explain when and why there could be a delay. There's no mention of issues with certificate validation, that might happen for the same reason.

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+=== Solution
+
+- Verify DNS records:
```
heruan, line 36: Verifying the records helps to check if propagation has completed, but it's not a solution per se. Setting a local hostname in `/etc/hosts` waiting for the propagation to complete could be one, for example.

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+[source,shell]
+----
+nslookup <domain>
+----
```
heruan, line 41: What the user should expect from the output of this command?

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+- Wait for propagation.
+- Check external DNS settings.
```
heruan, line 44: What settings should a user look for, and what changes can help?

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+== Crash Loop
+
+When a deployed application never becomes available because it keeps crashing on startup.
+This might be caused due to using an invalid docker image, or by Kubernetes killing a deployment that takes too long to startup thinking it is just dead since its liveness probe would not respond successfully.
+
+=== Solution
+
+Double-check the image of the deployed application. If the image is invalid, the deployment won't ever become available.
```
heruan, line 53: How to identify a valid image from an invalid one? Let's explain the correct syntax, the meaning of the tags and instruct the reader how to check the pod status and events that might help debugging the crash (with command and output examples).

articles/control-center/troubleshooting/index.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+Check the application logs and make sure they do not show any problematic message.
+If a specific error is shown in the logs, address it accordingly.
+If no specific error is shown in the logs, increase the startup delay time just enough to give it time to finish.
```
heruan, line 57: How to increase the startup delay? If it's elsewhere in the docs, a link would help. Also, if it's a startup delay issues there should be events of failing probes: better check for those to be sure the solution is the right one.

heruan (changes requested):
The outline is good, but there's plenty of room for improvement on common issue descriptions and solutions. Let's be more discursive, use a consistent language style with the rest of the docs, provide context with examples to help the user go through the hard time they have if they're reading this page.

heruan (approved):

PR #4199, opened 2025-03-21
docs: combo box lazy loading

sissbruecker:
Adds a section for how to implement lazy loading with combo box. This mainly showcases the newly added `setItemsPageable` and `useComboBoxDataProvider` APIs.

### DIFF
--- a/articles/components/combo-box/index.adoc
+++ b/articles/components/combo-box/index.adoc
@@ -339,6 +339,20 @@ TODO
 
 [.example]
 --
+ifdef::lit[]
+[source,typescript]
+----
+include::{root}/frontend/demo/component/combobox/combo-box-lazy-loading.ts[render,tags=snippet,indent=0,group=Lit]
+----
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/combobox/ComboBoxCountryService.java[render,tags=snippet,indent=0,group=Lit]
+----
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/combobox/ComboBoxCountryRepository.java[render,tags=snippet,indent=0,group=Lit]
+----
+endif::[]
 ifdef::react[]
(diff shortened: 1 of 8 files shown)


articles/components/combo-box/index.adoc
```diff
@@ -333,6 +333,59 @@ endif::[]
 --
 
 
+== Lazy Loading
+
+Combo Box supports lazy loading, which is useful when dealing with large datasets. This avoids having to load the whole dataset into memory and will only fetch data for the current filter and viewport.
```
taefi, line 338:
```suggestion
Combo Box supports lazy loading, which is useful when dealing with large datasets. This avoids loading the whole dataset into memory and only fetches data for the current filter and viewport.
```

frontend/demo/component/combobox/react/combo-box-lazy-loading.tsx
```diff
@@ -0,0 +1,22 @@
...
+// tag::snippet[]
+function Example() {
+  // Create a data provider that calls a backend service with a
+  // Spring Data pageable and the current combo box filter
+  const dataProvider = useComboBoxDataProvider(
+    ComboBoxCountryService.list.bind(ComboBoxCountryService)
+  );
+
+  return (
+    <ComboBox dataProvider={dataProvider} label="Country" itemLabelPath="name" itemValuePath="id" />
```
taefi, line 16: I'd suggest converting this into multiline to avoid the horizontal scrolling.

taefi (changes requested):
Generally, LGTM! Just a couple of suggestions:

frontend/demo/component/combobox/react/combo-box-lazy-loading.tsx
sissbruecker: Unfortunately that results in an eslint error from the prettier plugin.

PR #4227, opened 2025-04-03
docs: add instructions to create and/or delete secret with docker credentials

sosa-vaadin:
Add a subsection in the "Deploying an Application with Control Center" section to explain how to create and use a secret that holds user credentials to log in to a Docker registry. This is in order to pull an image from a docker registry as an authenticated user.

Fixes https://github.com/vaadin/control-center/issues/882


heruan (changes requested):
Please address the Vale comment to be consistent with the title case adopted by our docs.

heruan (approved):

PR #4251, opened 2025-04-14
hilla: add example usage of EndpointExposed

### LINKED ISSUE #4202: Add documentation / an example of using `@EndpointExposed` in docs
Currently, there is no mention of `@EndpointExposed` in the docs. Consider adding a paragraph with an example that shows how it works, DOs and DON'Ts.

cromoteca:
Include an example demonstrating how to use the `@EndpointExposed` annotation to control method exposure from superclasses in endpoint classes.

Closes #4202


taefi (approved):
LGTM!

PR #4254, opened 2025-04-15
Note about Vaadin-Refresh

mukherjeesudebi:
This PR adds a small description about Vaadin-Refresh token


articles/flow/advanced/modifying-the-bootstrap-page.adoc
```diff
@@ -202,6 +202,20 @@ public class AppShell implements AppShellConfigurator {
 
 Modifications in [methodname]`AppShellConfigurator.configurePage()` have priority over the equivalent annotations. Annotations don't cover all of the cases that can be achieved when overriding the [methodname]`AppShellConfigurator.configurePage()` method.
 
+==== Vaadin-Refresh[[vaadin-refresh]]
```
jouni, line 205: I feel this heading doesn't really tell me much about the purpose of this feature. Could we come up with a more descriptive heading? Could something like "Forcing a Page Reload" work? I’m not sure when someone would need this information, what problem/issue/use case they might be working with?

mukherjeesudebi: This information would be needed when someone gets the "Invalid JSON from server" error message in the view. So, basically this is a workaround that helps to reload the page. Some more information here https://github.com/vaadin/flow/issues/19514

mukherjeesudebi: I think "Forcing a Page Reload" is more aligned to the reason for adding the note about Vaadin-Refresh

jouni:
```suggestion
==== Forcing a Page Reload After an Invalid Server Response [[vaadin-refresh]]
```
Maybe?

articles/flow/advanced/modifying-the-bootstrap-page.adoc
```diff
@@ -202,6 +202,20 @@ public class AppShell implements AppShellConfigurator {
 
 Modifications in [methodname]`AppShellConfigurator.configurePage()` have priority over the equivalent annotations. Annotations don't cover all of the cases that can be achieved when overriding the [methodname]`AppShellConfigurator.configurePage()` method.
 
+==== Forcing a Page Reload After an Invalid Server Response[[vaadin-refresh]]
+
+If the XHR response can't be parsed as JSON, Vaadin looks for a “Vaadin-Refresh” string anywhere inside the response text. If it's present, Vaadin reloads the page instead of showing an error message. You can add the refresh token as a meta tag.
+
+[source,java]
+----
+public class AppShell implements AppShellConfigurator {
```
Legioth, line 211: Not sure if this code example makes sense since the unparseable response typically comes from some kind of proxy or authentication filter rather than directly from the regular application.

a1psaitz: I have the same issue. Instead of JSON from the Vaadin server, HTML comes from an Microsoft ADFS server, triggered as soon as the authentication token has expired. Now, I added 
`        settings.addMetaTag("refresh", "Vaadin-Refresh"); `
to my overwridden method 
`        public void configurePage(final AppShellSettings settings)`
For whatever reason, I can't see 
`        <meta name="refresh" content="Vaadin-Refresh"> `
in the HTML of my Vaadin pages. For years, I have 
`        settings.addFavIcon("icon", "/icons/icon.png", "210x210");`
`        settings.addLink("shortcut icon", "/icons/favicon.ico");`
in my configurePage() method. These two are in the HTML, but settings.addMetaTag() has no result.

TatuLund: @a1psaitz , " HTML comes from an Microsoft ADFS server" <- You need to add the META tag with "Vaadin-Refresh" string to this file instead of configuring Vaadin's index page.

a1psaitz: I'm not able to refactor the response from ADFS. Microsoft ADFS is a SaaS application for us. It's running in the Microsoft cloud.

TatuLund: @a1psaitz " triggered as soon as the authentication token has expired", I should have read this more carefully, it looks like you are using Microsoft EntraID config, are you using Spring Security? SSO Kit? Anyways, when properly configured this error html page should not be displayed at all, but upon token expiry Spring Security filter should redirect user to login page.

a1psaitz: I would suggest : If the response is HTML instead of JSON, the browser should render this HTML. 
Why : The HTML can't come from the Vaadin server. There must be a reason for that HTML.
How : Perhaps a new setting in application.yaml is required -> vaalin.renderHtmlOnHeartbeat: true

TatuLund: @a1psaitz I have some doubts to what you are proposing, mainly concerning security. My first reaction is, that if it is done that way, it opens an attack vector. @Legioth could comment this as well. I am also thinking that refreshing the page on what ever offending content could be problematic too, however if it is not a default and needs to be enabled by API, it would be then responsibility of you to ensure that in your environment this is safe, or you are willing to carry the potential risks. Idea of the Vaadin-Refresh magic string included in that HTML page is that you had access to do that, and you are this enabling it on this particular page only. Any other tampered or faulty response to the client will still produce an error.

TatuLund (approved):

PR #4257, opened 2025-04-22
chore: Upgrade supported Node.js to 20

mshabarov:
Node.js goes out of support and React 19 needs Node.js 20+

### DIFF
--- a/articles/compatibility.adoc
+++ b/articles/compatibility.adoc
@@ -17,7 +17,7 @@ This page shows the minimum required versions of supported technologies for Vaad
 
 | Java| 17 or later
 | Spring Boot| 3.0 or later
-| Node.js| 18 or later
+| Node.js| 20 or later
 | Maven| 3.5 or later (except for 3.8.2 and 3.8.3)
 | Gradle| 8.5 or later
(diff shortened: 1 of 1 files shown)


articles/compatibility.adoc
```diff
@@ -17,7 +17,7 @@ This page shows the minimum required versions of supported technologies for Vaad
 
 | Java| 17 or later
 | Spring Boot| 3.0 or later
-| Node.js| 18 or later
+| Node.js| 20 or later
```
Artur-, line 20: This seems to be a 24.0 list and now you are changing only one part of it but still linking to 24.0.0 release notes above. It should be made clearer that this is for the latest minor of Vaadin 24, link to that and also update all the other rows from 24.7.0 release notes

peholmst:
Is this PR still relevant?

peholmst:
@mshabarov Understood. If you want this change to show up in the `main` branch as well (currently used for V25), add the `target/main` label to the PR. It will be automatically cherry-picked after merging.

mshabarov:
Yes, I don't even remember why haven't I chosen the `main` branch from the beginning.

PR #4269, opened 2025-04-25
Added information about how to get the WebBrowser object

OlliTietavainenVaadin:
Added information about how to get the `WebBrowser` object


articles/flow/advanced/browser-access.adoc
```diff
@@ -32,6 +32,23 @@ UI.getCurrent().getPage().fetchCurrentURL(currentUrl -> {
 ----
 
 // TODO add documentation on VaadinSession.getCurrent().getBrowser()
+== Accessing browser information
```
mcollovati, line 35:
```suggestion
== Accessing Browser Information
```

articles/flow/advanced/browser-access.adoc
```diff
@@ -32,6 +32,23 @@ UI.getCurrent().getPage().fetchCurrentURL(currentUrl -> {
...
+
+The initial request contains the user-agent information supplied by the browser. These details can be found from the WebBrowser object,
```
mcollovati, line 37: I would rephrase the first statement, since all requests contain user-agent info, not only the initial one.
Maybe something like the following:

During the initial HTTP request, Vaadin reads the user-agent information supplied by the browser and stores them in a [classname]`WebBrowser` instance, which is accessible from the current [classname]`VaadinSession`

mcollovati (changes requested):

articles/flow/advanced/browser-access.adoc
```diff
@@ -32,6 +32,22 @@ UI.getCurrent().getPage().fetchCurrentURL(currentUrl -> {
 ----
 
 // TODO add documentation on VaadinSession.getCurrent().getBrowser()
```
mcollovati, line 34: Should this comment be removed?

PR #4274, opened 2025-04-28
docs: add Grid Pagination chapter and examples

bwajtr:
Added a chapter to the Grid documentation about the differences between Infinite Scrolling and Pagination.
Added pros and cons for both.
Also added Pagination implementation example in Flow, Lit and React. At the end the text is not that long, so I've decided to not add a new tab for this topic, but rather place it under Lazy Loading chapter

image


src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
```diff
@@ -0,0 +1,253 @@
...
+@SuppressWarnings("FieldCanBeLocal")
+@Route("grid-manual-pagination")
+public class GridManualPagination extends VerticalLayout {
+
+    private final PaginationControls paginationControls = new PaginationControls();
+    private final DataSource dataSource = new DataSource();
+    private final TextField searchField = new TextField();
+
+    // tag::snippet[]
+    private final DataProvider<Person, Void> dataProvider = DataProvider.fromCallbacks(query -> {
```
TatuLund, line 37: I would use fromFilteringCallbacks instead

```
    private final CallbackDataProvider<Person, String> pagingDataProvider = DataProvider
            .fromFilteringCallbacks(query -> {
```

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
```diff
@@ -0,0 +1,253 @@
...
+        var limit = paginationControls.getPageSize();
+
+        return dataSource.fetch(searchField.getValue(), offset, limit);
+    }, query -> {
+        // Total count of filtered items
+        var itemCount = dataSource.count(searchField.getValue());
+
+        var offset = paginationControls.calculateOffset();
+        var limit = paginationControls.getPageSize();
+
```
TatuLund, line 55:
```
                // Update total item count here to avoid calling
                // dataSource.count twice
                paginationControls.update(itemCount);
```

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
```diff
@@ -0,0 +1,253 @@
...
+        setPadding(false);
+
+        Grid<Person> grid = new Grid<>(Person.class, false);
+        grid.addColumn(createPersonRenderer()).setHeader("Name").setFlexGrow(0)
+                .setWidth("230px");
+        grid.addColumn(Person::getEmail).setHeader("Email");
+        grid.addColumn(Person::getProfession).setHeader("Profession");
+
+        grid.setAllRowsVisible(true); // this will prevent scrolling in the grid
+        grid.setDataProvider(dataProvider);
```
TatuLund, line 71: var dataProvider = pagingDataProvider.withConfigurableFilter();
        grid.setDataProvider(dataProvider);

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
```diff
@@ -0,0 +1,253 @@
...
+
+        paginationControls.update(dataSource.count(searchField.getValue()));
```
TatuLund, line 73: paginationControls.update can be removed here when dataprovider did the update

frontend/demo/component/grid/react/grid-manual-pagination.tsx
```diff
@@ -0,0 +1,210 @@
...
+  };
+
+  const pageCount = calculatePageCount(totalItemCount, pageSize.value);
+
+  useEffect(() => {
+    // Adjust the current page if it exceeds the new page count as a side effect of the total item count changing.
+    if (currentPage.value > pageCount) {
+      onCurrentPageChanged(pageCount);
+    }
+  }, [totalItemCount]);
```
web-padawan, line 54: Consider also changing `totalItemCount` to `Signal<number>`
Then you could use `useComputed` for `pageCount` e.g. like this:

```ts
const pageCount = useComputed<number>(() => {
    if (totalItemCount.value === 0) {
      return 1; // Display one page even if there are no items
    }
    return Math.ceil(totalItemCount.value / pageSize.value);
  };
```

It might require some additional changes but it feels like signals should be used consistently.

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
bwajtr: Thanks, I've updated it like so

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
bwajtr: Thanks, I've updated it like so

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
bwajtr: Thanks, I've updated it like so

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
bwajtr: Thanks, I've updated it like so

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
```diff
@@ -0,0 +1,253 @@
...
+        var gridWithPaginationLayout = new VerticalLayout(grid, paginationControls);
+        gridWithPaginationLayout.setPadding(false);
+        gridWithPaginationLayout.setSpacing(false);
+        gridWithPaginationLayout.getThemeList().add("spacing-xs");
+
+        searchField.setWidth("50%");
+        searchField.setPlaceholder("Search");
+        searchField.setPrefixComponent(new Icon(VaadinIcon.SEARCH));
+        searchField.setValueChangeMode(ValueChangeMode.EAGER);
+        searchField.addValueChangeListener(e -> {
```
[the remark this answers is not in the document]
bwajtr: Thanks, I've updated it like so

frontend/demo/component/grid/react/grid-manual-pagination.tsx
bwajtr: Thanks a lot. This also simplified the code a bit.
I've also used the computed signals in the main component -> to compute the `gridItems` (after applying filters) from the `allItems`. It kind of made sense to me. Please check if this makes sense to you too :).

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
```diff
@@ -0,0 +1,276 @@
...
+    @NotNull
+    private TextField createSearchField() {
+        TextField searchField = new TextField();
+        searchField.setWidth("50%");
+        searchField.setPlaceholder("Search");
+        searchField.setPrefixComponent(new Icon(VaadinIcon.SEARCH));
+        searchField.setValueChangeMode(ValueChangeMode.EAGER);
+        return searchField;
+    }
+    // end::snippet[]
```
rolfsmeds, line 111: Could the snippet end already before `wrapWithVerticalLayout`, as that, and `createSearchField` are not central to the solution being demonstrated?

rolfsmeds (approved):
Well written, looks good. The only thing that bothers me a bit is that the full code samples for the solution are very long. Could consider splitting the pagination controls implementation into separate files, but that could also be a later refinement.

src/main/java/com/vaadin/demo/component/grid/GridManualPagination.java
bwajtr: Sure, I've shortenet the snippet this way.

PR #4280, opened 2025-04-30
[Control Center] Add steps to install Control Center in WSL

tamasmak:


jouni:
> Ensure that the section "Install Control Center in WSL" maintains consistent spelling and grammar. For example, "First you need to install the WSL" could be rephrased for clarity, such as "First, you need to install WSL."

Do I understand this correctly, that this review is only pointing out just one example of spelling and grammar issues that should be fixed, instead of all issues?

articles/control-center/getting-started/local-environment.adoc
```diff
@@ -119,3 +119,63 @@ To start deploying your Vaadin applications and take full advantage of Control C
...
+[source,shell]
+----
+wsl --install -d <Distribution Name>
+----
+
+One option is to use the Ubuntu distribution:
+[source,shell]
+----
+wsl --install -d Ubuntu
+----
```
MatthewVaadin, line 137: Let's shorten this section. It's fairly obvious which part can be changed from the second example. Also slightly improved grammar for readability.
```suggestion
One solution to install Control Center on Windows in a local environment is to use https://www.docker.com/products/docker-desktop[Docker Desktop] and https://learn.microsoft.com/en-us/windows/wsl[WSL] (Windows Subsystem for Linux).

First, you need to https://learn.microsoft.com/en-us/windows/wsl/install[install] WSL. It is important to choose a distribution that allows installation of Helm, for example Ubuntu:

[source,shell]
----
wsl --install -d Ubuntu
----
```

articles/control-center/getting-started/local-environment.adoc
```diff
@@ -119,3 +119,63 @@ To start deploying your Vaadin applications and take full advantage of Control C
...
+
+Then you need to enable the Kubernetes and WSL built-in engines in Docker Desktop:
```
MatthewVaadin, line 139: Simplify.
```suggestion
Then, you need to enable Kubernetes and WSL in Docker Desktop:
```

articles/control-center/getting-started/local-environment.adoc
```diff
@@ -119,3 +119,63 @@ To start deploying your Vaadin applications and take full advantage of Control C
...
+
+image::images/enable_kubernetes.png[Enable Kubernetes]
+
+image::images/enable_wsl.png[Enable WSL]
+
+You also need to enable the integration with the Ubuntu distribution in Docker Desktop:
```
MatthewVaadin, line 145: Simplify.
```suggestion
You also need to enable integration with the Ubuntu distribution in Docker Desktop:
```

articles/control-center/getting-started/local-environment.adoc
```diff
@@ -119,3 +119,63 @@ To start deploying your Vaadin applications and take full advantage of Control C
...
+
+image::images/enable_ubuntu_distribution.png[Enable Ubuntu Distribution]
+
+Then you can start the WSL with the following command using the distribution you chose before, in this case the Ubuntu:
```
MatthewVaadin, line 149:
```suggestion
Then, you can start the WSL with the following command, using the distribution you chose before (in this case Ubuntu):
```

articles/control-center/getting-started/local-environment.adoc
```diff
@@ -119,3 +119,63 @@ To start deploying your Vaadin applications and take full advantage of Control C
...
+
+[source,shell]
+----
+wsl.exe -d Ubuntu
+----
+
+After you need to use the WSL terminal to install Helm and Control Center. First,  you need to https://helm.sh/docs/intro/install/#from-apt-debianubuntu[install] Helm before you can start to xref:../getting-started/local-environment#deploying_control_center[install] Control Center in the local environment with the command:
```
MatthewVaadin, line 156:
```suggestion
After this, you need to use the WSL terminal to install Helm and Control Center. First,  you need to https://helm.sh/docs/intro/install/#from-apt-debianubuntu[install] Helm before you can xref:../getting-started/local-environment#deploying_control_center[install] Control Center in the local environment with the command:
```

MatthewVaadin, line 156: Do we need to reiterate the install command and adding entries to host files? This is covered in the linked installation guide. Also, it gives us another place to change when updating these processes (e.g. with `local.gd` we can skip editing the hosts file).

MatthewVaadin (changes requested):

MatthewVaadin:
> ## AI Language Review
> In the file "Getting Started in a Local Environment":
> ...
> 2. In the command "wsl.exe -d Ubuntu," the use of lowercase "e" in ".exe" may lead to confusion, as it is generally uppercase in Windows contexts.

I would ignore this advice. As far as I have experienced, ".exe" is always written in lowercase. This could be changing in the operating system (although not that I have noticed), but in the context of the command line, lowercase is fine.

articles/control-center/getting-started/local-environment.adoc
tamasmak: You are right, it is not needed to mention the install command again. I removed it. However, I would keep the host file part as there are automatic changes by Docker Desktop what is worth to point out.

tamasmak:
> > Ensure that the section "Install Control Center in WSL" maintains consistent spelling and grammar. For example, "First you need to install the WSL" could be rephrased for clarity, such as "First, you need to install WSL."
> 
> Do I understand this correctly, that this review is only pointing out just one example of spelling and grammar issues that should be fixed, instead of all issues?

It looks like that, it mentioned only one example. There were also other changes suggested, but not strictly connected to grammar and spelling, which were resolved by pushing the changes for the review comments.

articles/control-center/getting-started/local-environment.adoc
jouni: > it gives us another place to change when updating these processes

That could be alleviated by `include::[]`:ing that content from the installation guide. No need to copy paste it.

MatthewVaadin (approved):

PR #4294, opened 2025-05-08
docs: refactor impersonation example

mcollovati:
Refactors impersonation example to get the correct security context holder as a parameter, to avoid wiring the dependency by name.


mshabarov (approved):

articles/flow/security/enabling-security.adoc
```diff
@@ -587,7 +587,7 @@ To add impersonation for a Vaadin application, create the [classname]`SwitchUser
 ----
 
 [NOTE]
-The bean should depend on `VaadinSecurityContextHolderStrategy` bean. If the [classname]`SwitchUserFilter` is initialized first, the wrong security holder is used and the feature won't work.
+The bean should use `VaadinSecurityContextHolderStrategy` bean to work properly. If the [classname]`SwitchUserFilter` is initialized differently, the wrong security holder is used and the feature won't work.
```
heruan, line 590: I would suggest to inject the strategy as the interface `SecurityContextHolderStrategy`, in case the user provides their own strategy bean.

mcollovati: The filter needs `VaadinSecurityContextHolderStrategy` to work properly with Vaadin. A different strategy (e.g. `ThreadLocal`) will potentially fail.

mshabarov (approved):

PR #4302, opened 2025-05-12
Add better description how header title and content slots are aligned

TatuLund:


articles/components/dialog/index.adoc
```diff
@@ -63,6 +63,8 @@ The Dialog component has static header and footer areas, and a scrolling content
 
 The header contains an optional title element, and a slot next to it for custom header content, such as a close button.
 
+When there is title text or element set via API, like in the example below, the content slot is automatically right aligned, otherwise it is left aligned.
```
rolfsmeds, line 66:
```suggestion
If the dialog has a title (set via text or custom element), the header content slot is right-aligned. If no title is present, the header content slot is left-aligned.
```

rolfsmeds (approved):

PR #4315, opened 2025-05-15
docs: add MessageList Markdown example

tomivirkki:
Add documentation for the `MessageList` component Markdown support

Related to https://github.com/vaadin/platform/issues/7601


articles/components/message-list/index.adoc
```diff
@@ -82,13 +82,11 @@ In versions prior to 24.3, theme names must be used instead of class names (`the
 [role="since:com.vaadin:vaadin@V24.8"]
 == Markdown
 
-You can configure the Message List to format the messages in Markdown instead of plain text.
+Message List can be configured to support formatting messages in Markdown instead of plain text.
```
jouni, line 85: Why the change to passive style, which we specifically discourage?

tomivirkki: Wasn't sure so took some inspiration from some of the most recently added [documentation](https://vaadin.com/docs/latest/components/master-detail-layout#stack-mode). I can revert the style change.

rolfsmeds: _Do_ we specifically discourage passive style? Technical documentation is commonly written in passive, after all, and the alternative tends to lead to half of all sentences starting with _"You"_

rolfsmeds (approved):

articles/components/message-list/index.adoc
jouni: - https://vaadin.com/docs/latest/contributing/docs/styleguide#writing-style
- https://github.com/vaadin/docs/blob/latest/.github/styles/Vaadin/Passive.yml

Mainly just following what the big teams are doing: https://developers.google.com/style/voice

jouni: > took some inspiration from some of the most recently added documentation

That’s slightly problematic, since we don't have anyone anymore actively reviewing the style, only the LLM review that tends to be quite “permissive”.

jouni: @alvarezguille, one data point to consider when improving the LLM review.

PR #4325, opened 2025-05-22
docs: Copilot privacy concerns

MarcinVaadin:
Added Privacy Concerns section with more detailed information about what is being sent and how to opt out.


emarc:
Looks goo, I suggest the following adjustments:

1. Name the section just "Privacy" (drop "concerns")
2. "User is asked before first operation if AI usage is allowed." -> "User asked for permission before operation. This can be changed via Settings > AI Usage"
3. "Command Window prompt submitted": add "User asked for permission before operation (same as above.)"
4. "User Activity" -> "Copilot usage" (not super good, but works well with the next one)
5. "Copilot error happened" -> "Copilot error"

PR #4328, opened 2025-05-22
docs: card docs update

rolfsmeds:
- Removed Preview
- Updated Since badge to 24.8
- Added a11y stuff
- Moved style variants to Styling page
- Combined Outlined and Elevated to single sample
- Added source and API links


articles/components/card/index.adoc
```diff
@@ -4,18 +4,16 @@ page-title: Card component | Vaadin components
 description: The Card component is a versatile container for grouping related content and actions, with several customization options for layout and appearance.
 meta-description: The Card component is a versatile container for grouping related content and actions, with several customization options for layout and appearance.
 section-nav: badge-preview
```
sissbruecker, line 6: The badge can be removed as well

articles/components/card/index.adoc
sissbruecker, line 1: Can also remove the feature flag from `vaadin-featureflags.properties` as is done here: https://github.com/vaadin/docs/pull/4333/files#diff-b44159d3b4e58bb46772d00323f0b89630c5718e1fd9dba4a95b415f5568d980

But then the PR can only be merged after docs have been updated to a version where the flag has been removed from components.

articles/components/card/index.adoc
rolfsmeds: Good catch. Removed now.

articles/components/card/index.adoc
rolfsmeds: Indeed! Might as well wait for that then.

PR #4320, opened 2025-05-20
[Control Center] Add database feature section

MatthewVaadin:
Adds a new section to the Control Center documentation for the Database feature.

Part of https://github.com/vaadin/control-center/issues/887
Fixes https://github.com/vaadin/control-center/issues/889

PR #4332, opened 2025-05-23
[Control Center] Add a page for Flyway migrations

MatthewVaadin:
Fixes #889


MatthewVaadin:
Adding content to #4320

PR #4352, opened 2025-05-30
Explain how to install JBR manually

peholmst:


MarcinVaadin:
Is this guide dedicated for Visual Studio Code or generic? If for VS Code then easiest option is to use Vaadin extension and it will do everything for user. If it is supposed to be generic tutorial, then we should remove VS Code references.

manolo:
>If for VS Code then easiest option is to use Vaadin extension and it will do everything for user
@MarcinVaadin

peholmst:
I've been trying to create a generic instruction that only uses Maven and a text editor. As the classpath is huge, I've tried to use `./mvnw` to start the application. I can see from the logs that hotswap agent is activated, but it does not detect when I recompile the project. Is this something that requires an IDE and special tool support?

peholmst:
Answering my own question: I guess Maven does not do incremental builds.

PR #4353, opened 2025-05-30
feat: Add Flow Signal documentation

mshabarov:


articles/flow/advanced/signals.adoc
```diff
@@ -0,0 +1,429 @@
...
+----
+
+=== Cleanup Effects When No Longer Needed
+
+Store the returned [classname]`Runnable` from [methodname]`Signal.effect()` and call it to clean up the effect when it's no longer necessary.
+
+[source,java]
+----
+Runnable cleanup = Signal.effect(() -> {
+    // Update UI
```
Legioth, line 377: Same about avoiding examples that "use" `effect` together with anything related to the UI

articles/flow/advanced/signals.adoc
```diff
@@ -0,0 +1,429 @@
...
+});
+
+// Later, when the effect is no longer needed
+cleanup.run();
+----
+
+== Advanced Topics
+
+=== Computed Signals
```
Legioth, line 386: Doesn't seem like this belongs in the "advanced" section since it shows exactly the same thing that was already shown in the very beginning of the page.

articles/flow/advanced/signals.adoc
```diff
@@ -0,0 +1,429 @@
...
+ValueSignal<String> firstName = SignalFactory.IN_MEMORY_SHARED.value("firstName", String.class);
+ValueSignal<String> lastName = SignalFactory.IN_MEMORY_SHARED.value("lastName", String.class);
+
+Signal<String> fullName = Signal.computed(() ->
+    firstName.value() + " " + lastName.value());
+----
+
+=== Signal Mapping
+
+You can transform a signal's value using the map() method.
```
Legioth, line 401: Could mention that this is just a shorthand for creating a computed signal

articles/flow/advanced/signals.adoc
```diff
@@ -0,0 +1,429 @@
...
+
+=== Read-Only Signals
+
+You can create read-only versions of signals that don't allow modifications.
+
+[source,java]
+----
+ValueSignal<String> name = SignalFactory.IN_MEMORY_SHARED.value("name", String.class);
+ValueSignal<String> readOnlyName = name.asReadonly();
+----
```
Legioth, line 418: Could clarify that changes can still be made through `name` and that such changes will also be visible through `readOnlyName`

articles/flow/advanced/signals.adoc
```diff
@@ -0,0 +1,429 @@
...
+
+=== Untracked Signal Access
+
+You can access a signal's value without registering a dependency, i.e., without triggering reactive effect functions.
+It's also possible to pick the confirmed value of a signal without triggering the effect function.
+
+[source,java]
+----
+String name = nameSignal.peek(); // Won't trigger Signal.effect()
```
Legioth, line 427: Might be easier to understand how this works if the code would be inside an actual `effect` callback.

articles/flow/advanced/signals.adoc
```diff
@@ -65,237 +69,212 @@ Transactions allow grouping multiple signal operations into a single atomic unit
...
-            user.update(u -> new User(u.getName(), event.getValue())));
+        field.addValueChangeListener(event -> {
+            // Only update signal if value has changed to avoid triggering infinite loop detection
+            if (!event.getValue().equals(value.peek())) {
+               value.value(event.getValue());
+            }
+        });
 
-        add(nameField, ageField);
+        add(field);
```
mshabarov: Now looking at this (and older) example I'm thinking about adding one more component which value depends on the text input. Otherwise the code looks very synthetic: you input a value, it updates the signal and then it updates the value back to the same value, so visually Ui looks like as you just type a text. But from the other side, the example is anyway extremely simple, and rather should show the principle.

mshabarov (commented):
These additions look good to me. Eager to merge 👍

PR #4373, opened 2025-06-09
Add note about Kubernetes Kit not being compatible with Spring Session

TatuLund:
See: https://github.com/vaadin/flow/issues/17799


TatuLund:
@mcollovati & @heruan , should the right place be around these hoods: https://vaadin.com/docs/latest/flow/integrations/spring , e.g. adding there mentioning that for session serialization one needs to use Kubernetes Kit instead of Spring Session?

mcollovati:
@TatuLund https://vaadin.com/docs/latest/flow/integrations/spring seems a good place to put a note.
But instead of saying that `one needs to use Kubernetes Kit instead of Spring Session`, I would say something like that currently Vaadin does not support Spring Session out-of-the-box but Kubernetes Kit can be used as an alternative, stating that it is a commercial component.

I wonder if we should link or provide a brief summary of the following comments to explain the reasons:

https://github.com/vaadin/kubernetes-kit/issues/4#issuecomment-1276004678
https://github.com/vaadin/kubernetes-kit/issues/10

peholmst:
What's the status of this PR?

TatuLund:
@heruan I added it here now.

https://github.com/vaadin/docs/pull/4703

PR #4393, opened 2025-06-23
(hilla) improve routing documentation clarity for routes.tsx usage

cromoteca:
Enhance the documentation to clarify the usage of the `routes.tsx` file, including its impact on routing behavior and the implications of using a custom file.


taefi (approved):
LGTM!

PR #4396, opened 2025-06-24
mark Control Center backups as a commercial feature

sosa-vaadin:
Also recommend users use files to create secrets with sensitive data.

### DIFF
--- a/articles/control-center/database/backups.adoc
+++ b/articles/control-center/database/backups.adoc
@@ -3,9 +3,13 @@ title: Backups
 page-title: Database backups in Control Center
 description: How to manage backups in Control Center
 meta-description: Create, schedule, or delete database backups in Control Center
+section-nav: commercial
 ---
 
 = Backups
+
+:commercial-feature: Backups
+include::{articles}/_commercial-banner.adoc[opts=optional]
 
 Control Center manages databases using https://cloudnative-pg.io[CloudNativePG], an open-source Kubernetes operator to manage PostgreSQL on any supported Kubernetes cluster.
 PostgreSQL natively provides https://www.postgresql.org/docs/current/backup.html[backup and recovery] capabilities based on file system level (physical) copy.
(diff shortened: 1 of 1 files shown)


articles/control-center/database/backups.adoc
```diff
@@ -104,12 +108,14 @@ You will need the following information about the object store:
 - `ACCESS_KEY_ID`: the ID of the access key that will be used to upload files into S3
 - `ACCESS_SECRET_KEY`: the secret part of the access key mentioned above
 
+WARNING: To prevent the sensitive values of the secrets to be *stored in the shell's history*, write down the values in files *without an extra newline character* at the end of the text and pass those files to `kubectl` to save their content as the secret's values.
```
MatthewVaadin, line 111:
```suggestion
WARNING: To prevent the sensitive values of the secrets being *stored in the terminal command history*, write down the values in files (*without* an extra newline character at the end of the text) and pass those files to `kubectl` to save their content as the secret's values.
```

articles/control-center/database/backups.adoc
```diff
@@ -193,14 +199,16 @@ In order to access your storage account for backup and recovery of CloudNativePG
 When using either *Storage account access key* or *Storage account SAS Token*, the credentials need to be stored inside a Kubernetes Secret, adding data entries only when needed.
 The following command performs that:
 
+WARNING: To prevent the sensitive values of the secrets to be *stored in the shell's history*, write down the values in files *without an extra newline character* at the end of the text and pass those files to `kubectl` to save their content as the secret's values.
```
MatthewVaadin, line 202:
```suggestion
WARNING: To prevent the sensitive values of the secrets being *stored in the terminal command history*, write down the values in files (*without* an extra newline character at the end of the text) and pass those files to `kubectl` to save their content as the secret's values.
```

MatthewVaadin (changes requested):

articles/control-center/database/backups.adoc
```diff
@@ -35,14 +39,32 @@ IMPORTANT: Backups are configured during installation of Control Center.
...
 The available methods for backups are:
 
 - <<Volume Snapshots>>
 - <<Barman Object Stores>>
 
 To decide which method to use, it is recommended to read https://cloudnative-pg.io/documentation/1.25/backup/#object-stores-or-volume-snapshots-which-one-to-use[CloudNativePG's official documentation].
 
+=== Create License Secret
+
+Since backups are a commercial feature, you must provide a valid Vaadin pro license key to enable the feature.
```
MatthewVaadin, line 52: There's a couple of references to "pro" key in this section, but this is not synonymous with commercial license. Are only Pro keys allowed? If not, then we should make this wording generic.

Let's also mention that if they are using an offline key, it might need to be downloaded from their Vaadin account. We could link to the licenses page (https://vaadin.com/docs/latest/flow/configuration/licenses), where this is explained.

MatthewVaadin (changes requested):

PR #4402, opened 2025-06-26
Change deprecated StreamResource to DownloadHandler in code example to load image 

mukherjeesudebi:


articles/flow/application/resources.adoc
```diff
@@ -54,9 +54,9 @@ TODO
 
 === Class Resources
 
-You can use [classname]`StreamResource` to load images from the Java class path.
+You can use [classname]`DownloadHandler` to load images from the Java class path.
 In this case, the images need to be below `src/main/resources` in the Java compilation path.
-The path for [methodname]`getResourceAsStream()` is relative to that path.
+The path for [methodname]`forServletResource()` is relative to that path.
```
mcollovati, line 59: I think `forServletResouce` is not working for resources in `src/main/resources`.
It would work if you put the resource in the webapp root folder or `META-INF/resources/` or `src/main/resources/META-INF/resources/`.

For `src/main/resources` the `forClassResource()` method should be used, also for absolute paths (e.g. `DownloadHandler.forClassResource(getClass(), "/images/myimage.png")`.

We coudl probably add a `DownloadHandler.forClasspathResource(String path)` helper.

mukherjeesudebi: yes, you are right. I mixed up forClassResource and forServletResouce paths. 
So forServletResouce  we need the image to be at `src/main/resources/META-INF/resources/` and forClassResource() also works for absolute paths.

Thanks for rectifying

mcollovati (approved):

mcollovati (approved):

PR #4404, opened 2025-06-27
Add note and code example of disabling server modality

TatuLund:


articles/components/dialog/index.adoc
```diff
@@ -199,6 +199,21 @@ endif::[]
 
 Usually, non-modal dialogs should be draggable, so that the user can move them to access the user interface beneath.
 
+[NOTE]
+In some use cases you may want to disable the server side modality. This can be done after Dialog has been opened.
```
jouni, line 203: Maybe a new section (e.g., `=== Disabling Modality`) instead of an admonition? What are those cases, could we offer an example? And why do I have to do it after the dialog has been opened?

mcollovati (approved):
The example code is OK.
I have no strong opinion on Jouni's comment

PR #4422, opened 2025-07-04
Add info about files behavior

TatuLund:
See also: https://github.com/vaadin/flow-components/issues/1289


articles/components/upload/index.adoc
```diff
@@ -567,6 +567,11 @@ include::{root}/frontend/demo/component/upload/react/upload-error-messages.tsx[r
 endif::[]
 --
 
+=== Component state
+
+In order to conserve resources the files themselves are not part of the component server state. This has significant implications in component behavior. Removing the Upload component from the layout will mean that files in browser memory will be lost. The same applies to refreshing of the page. Check more details about [file handling](https://vaadin.com/docs/latest/components/upload/file-handling) on a separate document page.
```
jouni, line 572:
```suggestion
To conserve resources, the files themselves aren't part of the component's server-side state. This has significant implications for component behavior. Removing the Upload component from the layout means that the files in the browser memory are lost. The same applies when reloading the page. Read more about <</components/upload/file-handling#, file handling>>.
```

articles/components/upload/index.adoc
```diff
@@ -567,6 +567,11 @@ include::{root}/frontend/demo/component/upload/react/upload-error-messages.tsx[r
 endif::[]
 --
 
+=== Component state
```
jouni, line 570:
```suggestion
=== Component State
```

articles/components/upload/index.adoc
```diff
@@ -567,6 +567,11 @@ include::{root}/frontend/demo/component/upload/react/upload-error-messages.tsx[r
...
+
+In order to conserve resources the files themselves are not part of the component server state. This has significant implications in component behavior. Removing the Upload component from the layout will mean that files in browser memory will be lost. The same applies to refreshing of the page. Check more details about [file handling](https://vaadin.com/docs/latest/components/upload/file-handling) on a separate document page.
+
+The UI including Upload component should take this into account. For example if refreshing of the page is expected, you can display a message to the user that files needs to be reselected if this was not the first attach.
```
jouni, line 574:
```suggestion
The user interface, including the Upload component, should take this into account. For example, if you expect that the user will refresh the page, you can display a message to the user that files needs to be selected again, if this wasn't the first time the page was rendered.
```

Issue #4419, opened 2025-07-03
Explain how to use `MasterDetailLayout` with Hilla file router


cromoteca:
As discussed in https://vaadin.com/forum/t/master-detail-layout-component-with-router-integration/178169, `MasterDetailLayout` should be used in a layout view in Hilla File Router. This is not currently mentioned in the documentation.

See the discussion for all the details about navigation structure.

I'm not sure where this information belongs though.

rbrki07:
Hi @cromoteca and @rolfsmeds,
I added a draft for a pull request (#4439) that includes some minor additions to the existing documentation and code example. Basically, these additions seam to be relevant to make the code example in the documentation work. The additions are based on @cromoteca input in the Vaadin forum thread mentioned above.

rolfsmeds:
Thanks @rbrki07, much appreciated!

cromoteca:
Done in https://github.com/vaadin/docs/pull/4439

PR #4439, opened 2025-07-23
docs: improve mdl hilla code example

rbrki07:
Suggestion to improve documentation as requested in #4419.


sissbruecker (approved):
Thanks! There is an [issue](https://github.com/vaadin/hilla/issues/3344) that would ideally make the dummy index file unnecessary, but until then it makes sense to have this in the docs.

PR #4443, opened 2025-07-25
docs: use addSeparator for adding dividers

mstahv:
... and tiny code polishing. Based on comment by Jan Heinrich 🏅

### DIFF
--- a/src/main/java/com/vaadin/demo/component/menubar/MenuBarDividers.java
+++ b/src/main/java/com/vaadin/demo/component/menubar/MenuBarDividers.java
@@ -18,7 +18,7 @@ public class MenuBarDividers extends Div {
         subMenu.addItem("Facebook");
         subMenu.addItem("Twitter");
         subMenu.addItem("Instagram");
-        subMenu.add(new Hr());
+        subMenu.addItem(new Hr());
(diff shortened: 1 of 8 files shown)


src/main/java/com/vaadin/demo/component/menubar/MenuBarDividers.java
```diff
@@ -18,7 +18,7 @@ public MenuBarDividers() {
         subMenu.addItem("Facebook");
         subMenu.addItem("Twitter");
         subMenu.addItem("Instagram");
-        subMenu.add(new Hr());
+        subMenu.addItem(new Hr());
```
sissbruecker, line 21: `addComponent` should be used for adding generic content that doesn't get wrapped in a menu item. Which is what is used three lines below.

`addSeparator` would be an alternative.

mstahv: Well that is a nice API, especilly if it was documented 😁 FYI: no need to make me to land this kind of improvements, feel free to make this youself and just land it. I don't need the credit, I want better documentation 🧸 Opened the project, found out the branch, did the change back here...

jg9431949-pixel:
###

Issue #4456, opened 2025-08-04
Update overlay component examples to use `closed` event instead of `opened-changed`


sissbruecker:
Documentation examples for overlay-based components currently use the `opened-changed` event with the only purpose of reacting to the component closing on its own.

In v25 this event has a behavior change, in that removing a component from the DOM when `opened` changes to `false` (so the overlay closes) will not trigger the closing animation anymore. Instead you would need to listen for the `closed` event before you can remove it.

While we don't seem to have any examples of conditional rendering of overlay components, we should still update the examples to encourage using `closed` over `opened-changed` for this purpose so that developers don't break the closing animation when they start using conditional rendering.

As such all usages of `opened-changed` should be removed from the examples.

PR #4471, opened 2025-08-07
docs: use overlay closed event (main)

sissbruecker:
Closes https://github.com/vaadin/docs/issues/4456


sissbruecker:
There is one usage left for Context Menu which is used in some workaround when using it in grids. However using `opened-changed` is necessary for the workaround.

PR #4472, opened 2025-08-07
chore: upgrade vaadin 25.0.0-alpha3

ZheSun88:


ZheSun88:
this update is not enough yet. as there are errors when build the project.

Issue #4221, opened 2025-03-28
Add how-to guide for creating custom fields


peholmst:
The current documentation for creating custom fields is scattered across multiple pages and are also more conceptual than hands-on.

peholmst:
Work on this article goes into https://github.com/vaadin/docs/tree/custom-field-howto. This branch is not based on #4204, which means some work may be needed to merge these two branches once ready. I'll take care of that, though.

PR #4522, opened 2025-08-21
docs: Guide on creating Custom Field in flow

HerbertsVaadin:
Was assigned to create this guide by Petter.
Guide only covers Flow part. 
Not sure what to do about the other stubs. 

Covers flow version of #4221


peholmst (commented):
Very good! That `package-lock.json` file looks like it was included by mistake, though?

peholmst (approved):

PR #4532, opened 2025-08-25
fix: make docs compile with spring-boot 4

web-padawan:


web-padawan:
Closing in favor of #4531

PR #4537, opened 2025-08-26
Make the documentation promote serving through web servers

mstahv:
Vaadin is not a generic web server and is NOT good at generic file serving. It is better to let the servlet container or Spring MVC tackle these tasks than to guide users to more complex API that offers worse performance.


mshabarov (approved):
Looks reasonable to me indeed, thanks!
Sorry for the long review, I've lost it from my radar for a long time.

PR #4549, opened 2025-08-27
chore: remove deprecated VaadinWebSecurity examples

tltv:
VaadinWebSecurity is deprecated for removal and will be removed in V25. This change removes all example codes using VaadinWebSecurity.

RelatedTo: vaadin/flow#22129


mcollovati (approved):
LGTM.
The only part missing is the migration guide from `VaadinWebSecurity` to `VaadinSecurityConfigurer` that should probably be handled in a separate PR.

PR #4552, opened 2025-08-28
fix: password in console

sosa-vaadin:
There's an instruction for upgrading Control Center in which we recommend writing a password value into the terminal. This PR changes this insecure practice to use a reference to a file containing the password instead.


articles/control-center/getting-started/upgrade.adoc
```diff
@@ -62,13 +62,16 @@ kubectl -n control-center edit configmap control-center
...
 
 [source,bash]
 ----
 kubectl -n control-center create secret generic control-center-user \
 --from-literal=email=user@example.com \
---from-literal=password=examplepassword123
+--from-file=password=/path/to/password.txt
 ----
 
 You'll have to adjust these values, though, for your email address and password.
```
gtzluis, line 77: Suggestion for a more precise and neutral phrase:
> Replace the example values with your own email address and the path to your password file.

PR #4565, opened 2025-08-29
Usage statistics info and opt-out

mshabarov:
Provides general info about usage statistics and ways to opt-out.

Related-to https://github.com/vaadin/flow/pull/21217

### DIFF
--- a/articles/flow/testing/ui-unit/getting-started.adoc
+++ b/articles/flow/testing/ui-unit/getting-started.adoc
@@ -36,17 +36,6 @@ JUnit 5 support is added in Vaadin 24.
 ----
 --
 
-[NOTE]
-As UI Unit Test has a tester for the Chart component, it is required on the class path. If your application is not using the Chart component, you need to add a test scope dependency to it, otherwise a [classname]` ClassNotFoundException` will be thrown.
-[source,xml]
-----
-    <dependency>
-      <groupId>com.vaadin</groupId>
-      <artifactId>vaadin-charts-flow</artifactId>
-      <scope>test</scope>
-    </dependency>
-----
-
 
 == First UI Unit Test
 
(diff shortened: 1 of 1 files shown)


articles/flow/configuration/development-mode/index.adoc
```diff
@@ -202,6 +202,19 @@ When the property is set to `true`, Vaadin's professional and core components ar
...
+This helps us understand which Vaadin features are used in practice and guides future improvements.
+No personal or application-specific data is collected.
+
+You can disable statistics collection on the server by setting either:
+- the `vaadin.devmode.usageStatistics.enabled` system property, or
+- the `VAADIN_USAGE_STATS_ENABLED` environment variable
+
+to `false`.
+
+Follow the instructions in https://github.com/vaadin/vaadin-usage-statistics[usage statistics collector repository] to disable usage statistics on the client-side.
```
tltv, line 217: `vaadin-usage-statistics` component is imported by default (at least in development mode). If I read this first time, my first assumption would be that setting `vaadin.devmode.usageStatistics.enabled` or `VAADIN_USAGE_STATS_ENABLED` would automatically also stop importing `vaadin-usage-statistics` (or disable it) automatically. 
Now disabling requires two steps. And seems that the instructions in `vaadin-usage-statistics` regarding global disabling with a [script is broken](https://github.com/vaadin/vaadin-usage-statistics/issues/81). 
And additionally Flow projects may not have `package.json` to put `disableUsageStatistics` flag. Which would make opting-out not possible by following the current instruction.

mshabarov: Can we make it in steps: add the server-side opt-out docs and option for environment variables, then take a look at the client-side parameter, what is broken and add docs later?

PR #4605, opened 2025-09-18
Add identity-management

heruan:


MatthewVaadin:
There are several parts that are specific to Version 2. Should we add a version badge for those sections?
If so, we should also have an alternate version of line 21 as the default location of Keycloak is different in V1.

PR #4635, opened 2025-10-02
chore: replace theme annotation with custom injection mechanism

sissbruecker:
Remove use of the deprecated theme mechanism for injecting example styles into example components:
- removes the `@Theme` annotation from exported web components
- Lumo theme is imported directly from main stylesheet
- adds a custom `applyTheme` function to inject styles into example components, and use it from Flow, Lit and React examples


vursen (approved):
I manually checked the component pages and everything looked fine.

PR #4650, opened 2025-10-10
docs: update articles for production build change

tltv:
Removed production profile from articles that use example apps that build or will build production package without any production profile.

Part of vaadin/flow#22464

WIP: starters are not updated yet and they are pending on new Spring Boot 4 release that enables removal of the production profile.


articles/flow/production/production-build.adoc
```diff
@@ -146,6 +199,23 @@ The Vite server integration and live reload features -- which are available only
...
+    <dependency>
+        <groupId>com.vaadin</groupId>
+        <artifactId>vaadin-dev</artifactId>
+        <optional>true</optional>
+    </dependency>
+    ...
+</dependencies>
+----
+
+
```
mshabarov: I'd move this snippet to up (before a snippet with profile) so as to make it a primary recommended way for spring apps.

PR #4654, opened 2025-10-15
docs: add tree grid scroll to item section

ugur-vaadin:
Adds `scrollToItem` feature to Tree Grid "Programmatic Scrolling" section.

### DIFF
--- a/articles/components/tree-grid/index.adoc
+++ b/articles/components/tree-grid/index.adoc
@@ -146,8 +146,8 @@ include::{root}/src/main/java/com/vaadin/demo/component/treegrid/TreeGridScrollT
 
 `TreeDataProvider` supports this feature out of the box. For other data providers, you need to override and implement the following methods of [classname]`HierarchicalDataProvider`:
 
-- [methodname]#getParent(T item)#
-- [methodname]#getItemIndex(T item, HierarchicalQuery query)#
+- [methodname]`getParent(T item)`
+- [methodname]`getItemIndex(T item, HierarchicalQuery query)`
 
 === Scrolling to an Item by Path
 


articles/components/tree-grid/index.adoc
```diff
@@ -123,9 +123,34 @@ endif::[]
...
+----
+include::{root}/src/main/java/com/vaadin/demo/component/treegrid/TreeGridScrollToItem.java[render,tags=snippet,indent=0]
+----
+--
+
+`TreeDataProvider` supports this feature out of the box. For other data providers, you will need to implement [methodname]#getParent(T item)# and [methodname]#getItemIndex(T item, HierarchicalQuery query)#.
+
+=== Scrolling to an Item by Path
+
+To use this feature, you need to specify the path of the item you want to view. The scroll position of the grid is adjusted to bring that item into view.
```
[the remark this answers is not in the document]
ugur-vaadin: Applied suggestion.

src/main/java/com/vaadin/demo/component/treegrid/TreeGridScrollToItem.java
```diff
@@ -0,0 +1,61 @@
...
+import com.vaadin.flow.router.Route;
+
+import java.util.ArrayList;
+
+@Route("tree-grid-scroll-to-item")
+public class TreeGridScrollToItem extends Div {
+    private final TreeGrid<Person> treeGrid = new TreeGrid<>();
+
+    public TreeGridScrollToItem() {
+        var treeData = new TreeData<Person>();
```
sissbruecker: Nit: I'd prefer if we do not use `var` in example code as a general rule. When assigning a `var` from a method call you have no context on what type that returns, and since you're not in an IDE you can not easily check. While there are no such examples of that in this code it might still be a good practice to adhere to.

articles/components/tree-grid/index.adoc
```diff
@@ -123,11 +123,37 @@ endif::[]
...
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/treegrid/TreeGridScrollToItem.java[render,tags=snippet,indent=0]
+----
+--
+
+`TreeDataProvider` supports this feature out of the box. For other data providers, you need to override and implement the following methods of [classname]`HierarchicalDataProvider`:
+
+- [methodname]#getParent(T item)#
+- [methodname]#getItemIndex(T item, HierarchicalQuery query)#
```
sissbruecker:
```suggestion
- [methodname]`getParent(T item)`
- [methodname]`getItemIndex(T item, HierarchicalQuery query)`
```

Looks like other usages of `[methodname]` use back ticks, so let's apply those here as well.

PR #4656, opened 2025-10-15
docs: side nav auto expand

sissbruecker:
Add a section about the side nav's auto expand feature and how to disable it.

Part of https://github.com/vaadin/web-components/issues/9421


ugur-vaadin (approved):
Nit: I was wondering whether this section would be better off being just below "Nested Matching", but I don't have a strong opinion.

Issue #4222, opened 2025-03-28
Refactor Gradle article


mshabarov:
**Tell us about your request**

The article about Gradle https://vaadin.com/docs/latest/getting-started/start/alternatives/gradle is too hidden from reader and has too many content for a "get started" article.

I propose to move the article with the whole content next to https://vaadin.com/docs/latest/flow/configuration/maven article (same level), and leave only "get started" and general information in https://vaadin.com/docs/latest/getting-started/start/alternatives/gradle.

---

If you think this issue is important, add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

PR #4669, opened 2025-10-17
Update gradle documentation, split it into two pages.

HerbertsVaadin:
Solves #4222

Changes:
* Removed the "Developing in Eclipse IDE" section - unnecessary, not Vaadin specific information, if you use Gradle, and have chosen to use Gradle, and use Eclipse, you probably know how to import and run gradle application in Eclipse IDE.
* Reduced requirements section to a note - unecessary to have so much detail.
* Reduced starter project contents to only gradle-related files
* Reduced section "The most important files and folders are as follows" to only reference gradle related files, since the rest is the same between other build tools.
* Simplified compile section, referenced configuration page
* Removed warning about Gretty + jcenter, as it is no longer relevant.
* Created new page under flow/configuration specific to Gradle configuration properties.
* Refactored Configuring repositories section, removed focus from pre-releases.
* Removed the build file "Other Configuration" section, as its no longer valid.

### DIFF
--- a/articles/flow/configuration/gradle.adoc
+++ b/articles/flow/configuration/gradle.adoc
@@ -49,8 +49,8 @@ The `repositories` section defines the locations to search for packages. The rep
 
 ----
 repositories {
-    mavenCentral()
     mavenLocal()
+    mavenCentral()
     maven { url = "https://maven.vaadin.com/vaadin-addons" }
     maven { url "https://maven.vaadin.com/vaadin-prereleases" }
     maven { url "https://plugins.gradle.org/m2/" }
@@ -59,8 +59,8 @@ repositories {
 
 The example above shows configured repositories in the following order:
 
- * *Maven central* (required) - contains Vaadin and a lot of other public libraries
  * *Maven local* - Repository which looks in the local Maven cache for dependencies.
+ * *Maven central* (required) - contains Vaadin and a lot of other public libraries
  * repository for *Vaadin add-ons*
  * repository for *Vaadin pre-releases* - not recommended for production environments
  * *Gradle plugin repository* - only necessary when using Gradle community plugins, which are not available through Maven central


articles/flow/configuration/gradle.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+This page describes how to configure a Vaadin application using the Vaadin Gradle plugin.
+
+See <<{articles}/getting-started/start/alternatives/gradle#,Alternatives - Gradle>> page for a guide on creating a Vaadin project with Gradle.
+
+For information about using Gradle, see the link:https://docs.gradle.org/current/userguide/userguide.html[Gradle User Manual].
+
+.Requirements
+[NOTE]
+The Vaadin Gradle plugin requires Java SDK 17 or later, Gradle 8.5 or later.
```
mcollovati: Looking at the Vaadin 24.8 and 24.9 release notes, the minimum Gradle version is 8.7
```suggestion
The Vaadin Gradle plugin requires Java SDK 17 or later, Gradle 8.7 or later.
```

articles/flow/configuration/gradle.adoc
```diff
@@ -0,0 +1,329 @@
...
+It will install Node.js and npm automatically when running the build for the first time.
```
mcollovati: It could be worth it adding that install happens only if Node is not already installed or if it does not match the minimum version required by Vaadin.

articles/flow/configuration/gradle.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+
+[[build-file.vaadin-options]]
+=== Vaadin Plugin Configuration
+
+Vaadin Gradle plugin options are configured in a `vaadin` block. For development, the block is usually like this:
+
+----
+vaadin {
+    optimizeBundle = false
```
mcollovati: I'm not sure this option is needed for development. Ideally, for development there should be no need for any configuration.
What about changing the `For development, the block is usually like this` with something like `An example configuration could be like this` and use `frontendHotdeploy = true`  in the snippet?

articles/flow/configuration/gradle.adoc
```diff
@@ -0,0 +1,329 @@
...
+}
+----
+
+If the parameter is `true`, the frontend bundle is optimized for all supported browsers, but the compilation is much slower. For configuration options, see <<all-options,plugin configuration options>>
```
mcollovati: `optimizeBundle` is to enable the byte code scanner and make the bundle only contain frontend stuff that is actually used in the application. It has nothing to do with browser optimizations.

articles/flow/configuration/gradle.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+You'll need to add the `vaadin-core` or `vaadin` library as a Java dependency. You would do that like so:
+
+----
+dependencies {
+    implementation "com.vaadin:vaadin-core:24.+"
+}
+----
+
+When you specify a version of `24.+`, you're choosing to use the latest version of Vaadin. However, you can also specify the exact version. See https://docs.gradle.org/current/userguide/declaring_dependencies.html[Declaring Dependencies] in the Gradle documentation for more details.
```
mcollovati: nit: I'm not sure if we should suggest using an unpinned version. IMO, from a security point of view, I would always pin the exact version I want to use.

But this is not a blocker.

articles/flow/configuration/gradle.adoc
```diff
@@ -0,0 +1,329 @@
...
+
+.Define the snapshot plugin
+----
+buildscript {
+    ...
+    dependencies {
+        classpath group: 'com.vaadin',
+                name: 'vaadin-gradle-plugin',
+                version: '24.0-SNAPSHOT'
+    }
```
mcollovati: I'm not so skilled in Gradle, but at least our starter seems to provide a different plugin configuration.

It has the version pinned in `settings` file (https://github.com/vaadin/base-starter-gradle/blob/9a8284e07d02e0f6f346b49accbebd902e50865a/settings.gradle#L7) and only definition in the build script (https://github.com/vaadin/base-starter-gradle/blob/9a8284e07d02e0f6f346b49accbebd902e50865a/build.gradle#L15), no `apply` at all.

So the question is if it would be enough to set the snapshot version in settings in combination with the configuration of the pre-release repository.

I'm not sure what we should suggest in the docs. Maybe @mvysny can provide some insight

articles/flow/configuration/gradle.adoc
HerbertsVaadin: Thanks, replaced it with description for frontendHotdeploy, as optimizeBundle is no longer used.

articles/flow/configuration/gradle.adoc
```diff
@@ -0,0 +1,329 @@
...
+}
+----
+
+The example above shows configured repositories in the following order:
+
+ * *Maven central* (required) - contains Vaadin and a lot of other public libraries
+ * repository for *Vaadin add-ons*
+ * repository for *Vaadin pre-releases* - not recommended for production environments
+ * *Gradle plugin repository* - only necessary when using Gradle community plugins, which are not available through Maven central
+
```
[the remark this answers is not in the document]
HerbertsVaadin: thanks, added

articles/flow/configuration/gradle.adoc
HerbertsVaadin: I've updated to a modernized version, seems like the "apply" approach is historic.  Please check.

mcollovati (approved):

Issue #4432, opened 2025-07-18
Highlight that `vaadin-maven-plugin` must always be defined after `maven-compiler-plugin`


mcollovati:
There are several forum posts and GH tickets mentioning that sometimes when running an application in production mode results in error messages about missing resources (javascript, css, etc.).

An example:  https://vaadin.com/forum/t/urgent-production-build-failed/178117

One of the causes is that the `build-frontend` goal gets executed before the Java classes are compiled by the maven compiler plugin (holds also for the kotlin compiler plugin), so the bytecode scanner cannot find any class to analyze.

The documentation should explicitly state that `vaadin-maven-plugin` must always be defined after `maven-compiler-plugin`.

Potential pages to update could be [production build](https://vaadin.com/docs/latest/flow/production/production-build) and  [troubleshooting](https://vaadin.com/docs/latest/flow/production/troubleshooting).

---

If you think this issue is important, add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

PR #4676, opened 2025-10-21
 Highlight that vaadin-maven-plugin must always be defined after maven-compiler-plugin

HerbertsVaadin:
Solves: #4432

Production Build:
image

Troubleshooting Production Mode:
image


HerbertsVaadin:
I disagree with AI language Review suggestions, would prefer a human review.

mcollovati (approved):

Issue #2124, opened 2023-02-10
Document FlexLayout


OlliTietavainenVaadin:
Currently, the documentation mentions FlexLayout only indirectly. It should be documented, especially the features that aren't available in basic layouts such as wrapping.

**Vote to prioritize:** add a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to help the community and maintainers prioritize this issue.

HerbertsVaadin:
Flex Layout is documented here now:
https://vaadin.com/docs/next/components/html-elements

PR #4700, opened 2025-10-28
Add Flex Layout documentation

HerbertsVaadin:
Solves 
#2124

image

image

### DIFF
--- a/articles/components/html-elements/index.adoc
+++ b/articles/components/html-elements/index.adoc
@@ -83,12 +83,11 @@ Also, the HTML fragment must have exactly one root element. If the fragment cont
 
 Last thing to consider is that the `Html` component doesn't support SVG elements as a root node. For SVG content, the `Svg` component should be used instead.
 
-[discussion-id]`6774751B-921E-4B79-941E-830D9C3532B4`
 
 == Flex Layout Component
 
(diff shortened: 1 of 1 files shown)


rolfsmeds:
We're actually trying to get away from the need for FlexLayout entirely, so I'm not sure we want to now start promoting it like this. HL now has native support for wrapping, so the only feature it has that HL and VL lacks is `setDirection`.

HerbertsVaadin:
> We're actually trying to get away from the need for FlexLayout entirely, so I'm not sure we want to now start promoting it like this. HL now has native support for wrapping, so the only feature it has that HL and VL lacks is `setDirection`.

@rolfsmeds I guess for HL and VL it would only make sense to have "reverse" direct changes, but a use case where you'd want to switch between HL and VL isn't covered I guess. 

Wondering if it would be useful to add a disclamer on top, that the user should prefer to use HL or VL instead of this component, unless they have some specific scenario where the direction change is necessary? 
Just brainstorming, that "Flex" is more recognizable than "horizontal layout" or "vertical layout" for outside people, so a page for it could funnel the user to the right component? 

Since its not a deprecated component, feels like it should be documented.

HerbertsVaadin:
@rolfsmeds 
Sure, HTML Elements page makes sense, I can rework this PR into that. 
I could write up a much smaller summary, as a separate section that page, similar to how HTML Container has a separate section. Or should it be only only a table entry?

PR #4707, opened 2025-10-29
docs: add-on upgrade instructions

sissbruecker:
Adds a section with common challenges for add-on authors to the upgrade guide.


articles/upgrading/index.adoc
```diff
@@ -872,3 +872,126 @@ Vaadin uses the [filename]`{project directory}/src/main/frontend/` directory as
...
+    ];
+  }
+}
+----
+
+This would result in Lumo styles being applied to text nodes and basic HTML elements (headings, links) in the component’s shadow root.
+
+This is not possible anymore in v25, as the respective Typography module has been converted into a CSS file. If you need to apply Lumo typography styles to basic HTML elements in your component’s shadow root then you can copy the relevant styles into your component’s styles.
+
+==== Material Theme
```
web-padawan: Do we need a dedicated heading for Material here? There is one already under "Themes and Styling".

sissbruecker: Probably not necessary, removed it.

PR #4761, opened 2025-11-13
docs: new styling docs / styling components

sissbruecker:
Adds the "Styling Vaadin Components" article for the new styling docs.

> [!NOTE]
> The PR targets a base branch for the new styling docs. That branch has the old styling docs removed and new articles are added gradually until the new section is ready. Cross-references between new articles will be filled in later, respective sections have been marked with TODOs.

### DIFF
--- a/articles/styling/styling-components.adoc
+++ b/articles/styling/styling-components.adoc
@@ -233,7 +233,7 @@ vaadin-checkbox > label {
 
 ==== Pseudo-Elements
 
-The pseudo-elements `::before` and `::after` are elements generated with CSS. They are commonly used for font-based icons and highlight effects for states like hover and focus, often on one of the shadow parts rather than the root element. As an example, the Date Picker’s icon is a pseudo-element in the `toggle-button` shadow part, and its required indicator is a pseudo-element on the `required-indicator` shadow part:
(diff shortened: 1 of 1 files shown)


articles/styling/styling-components.adoc
```diff
@@ -0,0 +1,459 @@
...
+<source-info group="React"></source-info>
+<Button theme="primary">Save</Button>
+----
+--
+
+== Recommended Styling Approach
+
+The following layered approach to styling Vaadin components is recommended to minimize the amount of CSS, avoid breaking components and protect your styling from getting broken in future versions.
+
+. *Theme*: choose the <<{articles}/styling/themes#,theme>> that is the best match to your desired look and feel for components, or stick to the components’ base styles if neither theme is a good match;
```
web-padawan: We generally use `*` for bullet lists in other articles, should it be used here as well?

articles/styling/styling-components.adoc
```diff
@@ -0,0 +1,459 @@
...
+vaadin-text-field.bordered::part(input-field) {
+  background: white;
+  border: 1px solid gray;
+}
+----
+
+.CSS Class Names
+[NOTE]
+====
+CSS class names are identifier-attributes applied to HTML elements to scope CSS styling to them. The same class name can be applied to multiple elements, and each element can have multiple class names. They have nothing to do with Java or TypeScript classes.
```
web-padawan: Let's align wording with the other similar admonition above (maybe we don't need two of them):

```suggestion
CSS class names are identifier-attributes applied to HTML elements in order to scope CSS styling to them. The same class name can be applied to multiple elements, and each element can have multiple class names. They have nothing to do with Java or TypeScript classes.
```

PR #4799, opened 2025-11-20
docs: Spring Security with StyleSheet imports

tltv:
Add section to upgrading article to note how Spring Boot security configurations must be configured for custom static resources in StyleSheet.

Fixes: vaadin/flow#22476


mshabarov:
Would be worth to include a similar instruction to the styling article once it's merged - https://github.com/vaadin/docs/blob/docs/new-styling-docs-base/articles/styling/stylesheets.adoc.

PR #4810, opened 2025-11-24
Removed Board page, samples, and references to it

rolfsmeds:


ugur-vaadin:
It seems like there is a reference to Board in [MPR limitations page](https://github.com/vaadin/docs/blob/v24/articles/tools/mpr/configuration/limitations.adoc?plain=1#L115-L117).

rolfsmeds:
...aaand then the entire page where that reference was, was removed. Reverted the commit.

PR #4816, opened 2025-11-24
docs: replace Lumo utility in various examples

web-padawan:
Replaced usage of LumoUtility in some Grid, LoginOverlay and ProgressBar examples.


frontend/demo/component/grid/grid-manual-pagination.ts
sissbruecker: Grid has over two dozen other usages of Lumo props. If the plan is to switch examples to Aura I guess we would do this as part of that task? Then again, I wonder if we would use some Aura specific variables for the font size as well.

PR #4819, opened 2025-11-25
docs: loading additional stylesheets with Spring Security

sissbruecker:
Document how to set up Spring Security to load additional stylesheets.


articles/styling/stylesheets.adoc
```diff
@@ -96,4 +96,51 @@ CSS class names are identifier-attributes applied to HTML elements in order to s
 
 Stylesheets can also be <<{articles}/styling/advanced/dynamic-stylesheets#, loaded and unloaded dynamically>> through Java code.
 
+== Spring Security
+
+If you are using <<{articles}/flow/security/enabling-security#, Spring Security>> in your Vaadin application, you need to ensure that your stylesheets are accessible to unauthenticated users. The default Vaadin Spring Security configuration only permits access to `src/resources/META-INF/resources/styles.css` as well as the Aura and Lumo theme stylesheets.
```
mshabarov:
```suggestion
If you are using <<{articles}/flow/security/enabling-security#, Spring Security>> in your Vaadin application, you need to ensure that your stylesheets are accessible to unauthenticated users. The default Vaadin Spring Security configuration only permits access to `src/main/resources/META-INF/resources/styles.css` as well as the Aura and Lumo theme stylesheets.
```

articles/styling/stylesheets.adoc
```diff
@@ -96,4 +96,51 @@ CSS class names are identifier-attributes applied to HTML elements in order to s
...
+// end::snippet[]
+
+        return http.build();
+    }
+
+    ...
+}
+----
+
+This allows access to stylesheets placed in the `src/resources/META-INF/resources/css` folder. For example, if you place a stylesheet in `src/resources/META-INF/resources/css/additional-styles.css`, you can import it in your main stylesheet like so:
```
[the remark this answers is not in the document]
sissbruecker: I've rephrased it like so:
> This allows access to common locations for static resources, one of which is the `css` path. To make stylesheets available under that path, place them in the `src/main/resources/META-INF/resources/css` folder. For example, if you place a stylesheet in `src/main/resources/META-INF/resources/css/additional-styles.css`, you can import it in your main stylesheet like so:

articles/styling/stylesheets.adoc
sissbruecker: Thanks, there were actually several instances in the rest of the document as well. There are more in other documents, will create a separate PR for that.

PR #4845, opened 2025-11-27
refactor: update Accordion article to Aura theme

sissbruecker:
Updates the Accordion page and examples to use the Aura theme.

Noteworthy changes:
- Simplified accordion summary example to only show an icon in the accordion heading once the continue button is clicked. The Aura styles do not stretch the heading content, thus the respective panel details were glued to the panel title and it would have required some CSS to make this work. IMO this works out fine, in general the example felt too complex and the Binder setup was also distracting.
- Updated accordion content example to have colored links with underline
- Removed small theme variant example as it isn't supported by Aura


articles/components/accordion/styling.adoc
```diff
@@ -7,6 +7,92 @@ order: 50
 ---
 = Accordion Styling
 
+== Style Variants
```
sissbruecker: There probably should be a paragraph about how to apply style variants with a code snippet, especially since the terminology is now different from the `theme` attribute that you need to use. I think we can separately add a common section for this in a shared asciidoc and then include it in the styling pages that have a style variants section.

PR #4848, opened 2025-11-27
refactor: update Avatar page to Aura theme

sissbruecker:
Updates the Avatar page and examples to use the Aura theme.

Noteworthy changes:
- Removed avatar size variant example as it isn't supported by Aura
- Removed avatar menu bar example, included similar popover user menu example instead. That one is quite broken, but should be cleaned up when updating popover examples.


articles/components/avatar/styling.adoc
```diff
@@ -7,25 +7,69 @@ order: 50
...
 |Avatar Group overlap
 |`--vaadin-avatar-group-overlap`
-|`8px`
+|Aura, Lumo
 
-|Avatar Group overlap border
+|Avatar Group overlap border width
 |`--vaadin-avatar-group-overlap-border`
-|`2px`
+|Lumo
```
web-padawan: Should we remove or change "User Color Properties" below as that table lists Lumo default values?

web-padawan: Makes sense. Also, in Lumo we only have 7 colors (0-6) but in base styles we have 10 (0-9).

Issue #4801, opened 2025-11-21
Aura Theme in the Documentation


yuriy-fix:
Documentation need to updated in order to showcase the Aura theme.
The following actions need to be proceeded with:
- [ ] Create a new branch for applying Aura, apply Aura to all samples
- [ ] Go through the samples and fix / replace:
  - [ ] List the samples that require updates in the comment below
  - [ ] The `LumoUtility` usages either with inline styles or with stylesheet (dev judgment per case)
  - [ ] `lumo_*` variant names
  - [ ] `--lumo-*` properties
- [ ] Restructure Style Variants and Properties
  - [ ] Style Variants
    - [ ] Move `Style Variants` section under the `Styling` tab
    - [ ] Introduce a new table with listing the theme variants and `Theme Support` column
    - [ ] For the samples use only Aura supported variants
  - [ ] Properties 
    - [ ] Remove `Default Value` column
    - [ ] Introduce `Theme Support` column with listing (text for now) supported themes per the properties that don't support all themes
 - [ ] Research workaround for Badge component (use images or assess the feasibility of using Lumo for this page only) [2 days]

Other topics:
- [ ] Verify Spreadsheet docs page to not include Lumo samples
- [ ] Simplify the Board docs page by referencing the 24 docs with keeping the admonition about removal in 26.

PR #4870, opened 2025-12-02
docs: RTE style variants and props

DiegoCardoso:
Move style variants to styling and add props listing.

### DIFF
--- a/articles/components/rich-text-editor/styling.adoc
+++ b/articles/components/rich-text-editor/styling.adoc
@@ -121,7 +121,13 @@ include::../_styling-section-theming-props.adoc[tag=style-properties]
 | `--vaadin-rich-text-editor-toolbar-button-background`
 | Aura, Lumo
 
+| `--vaadin-rich-text-editor-toolbar-button-border-color`
+| Aura, Lumo
+
 | `--vaadin-rich-text-editor-toolbar-button-border-radius`
+| Aura, Lumo
+
+| `--vaadin-rich-text-editor-toolbar-button-border-width`
 | Aura, Lumo
 
 | `--vaadin-rich-text-editor-toolbar-button-padding`


articles/components/rich-text-editor/styling.adoc
```diff
@@ -7,6 +7,117 @@ order: 50
 ---
 = Rich Text Editor Styling
 
+== Theme Variants
```
sissbruecker: Let's use "Style Variants" and have a small intro text, same as we have for other components:
https://github.com/vaadin/docs/blob/c9f3ed544e37f88e07d168fbe83d0975809597d9/articles/components/dialog/styling.adoc#L13-L15

articles/components/rich-text-editor/styling.adoc
```diff
@@ -7,6 +7,117 @@ order: 50
...
+=== Compact
+
+[cols="1,3,1"]
+|===
+| Variant | Description | Supported by
+
+|`compact`
+| Makes the toolbar more compact
+| Lumo
+|===
```
sissbruecker: For consistency with other component pages, let's have a single table with all variants under "Style Variants". IMO we can just keep the existing descriptions from the sub-sections even if the content somewhat duplicates what is already explained in the table.

articles/components/rich-text-editor/styling.adoc
```diff
@@ -7,6 +7,117 @@ order: 50
...
+----
+include::{root}/frontend/demo/component/richtexteditor/react/rich-text-editor-theme-no-border.tsx[render,tags=snippet,indent=0,group=React]
+----
+endif::[]
+--
+
+include::../_styling-section-intros.adoc[tag=properties]
+
+
+[cols="1,3,1"]
```
sissbruecker: The layout is broken as there is only content for two columns.
```suggestion
[cols="3,1"]
```

articles/components/rich-text-editor/styling.adoc
DiegoCardoso: Done.

PR #4885, opened 2025-12-03
fix gha checking fail on too long arguements

ZheSun88:


ZheSun88:
normally this checking is only triggered when there are words changing in docs.
ideally, we wanted this tool to help us finding typos and correcting the grammar mistaken, which will be mentioned in the summary message. As we dont have the external help from the previous service provider.  

i dont use it very often in this repo, as my main activity is update the project version.. 
if you feel this is not necessary, we can remove it.. 
but i guess it doesn't hurt to have it.. if it doesn't cause troubles.

FYI @manolo

PR #4889, opened 2025-12-03
docs: progress-bar Aura style variants and props

DiegoCardoso:
Move style variants to styling and add props listing.

Part of #4801


articles/components/progress-bar/styling.adoc
```diff
@@ -7,6 +7,72 @@ order: 50
 ---
 = Progress Bar Styling
 
+== Style Variants
+
+Progress Bar comes with three style variants: contrast, success, and error.
+
+[cols="1,3,1"]
+|===
+|Variant | Usage recommendations | Supported by
```
sissbruecker: Let's make use a "Description" column for consistency and rephrase the bullet points as text. "Visual preference" could be removed, IMO it's kind of unclear without more context.

PR #4905, opened 2025-12-04
docs: upload Aura style variants and props

web-padawan:
Added Upload style variants (with a single Aura variant) and style props table.


articles/components/upload/styling.adoc
```diff
@@ -7,6 +7,152 @@ order: 50
...
+|`--vaadin-upload-border-color`
+|Aura
+
+|`--vaadin-upload-border-radius`
+|Aura
+
+|`--vaadin-upload-border-width`
+|Aura
+
+|`--vaadin-upload-padding`
```
vursen: There is also `--vaadin-upload-gap` which seems to be supported by both Lumo and Aura. Should we include it?

PR #4906, opened 2025-12-04
refactor: convert input field style variants to table

sissbruecker:
- Converts shared input field style variants to a table, similar to what other components are using
- Removes the border section for now
- Removes border width style from input field style variants examples, which could potentially be confused with being a side effect of applying a style variant

The table looks like this:

Bildschirmfoto 2025-12-04 um 16 47 07


articles/components/_input-field-common-features.adoc
```diff
@@ -29,7 +29,7 @@ Visible labels are strongly recommended for all input fields. In cases where the
 ====
 Helpers are used to provide additional information that the user may need to enter in the field, such as format requirements or explanations of the field's purpose below the field.
 
-A style variant is available for rendering the helper <<#helper-above-field,above the field>>.
+A <<styling#style-variants,style variant>> is available for rendering the helper above the field.
```
sissbruecker: This will be broken for most component pages until the style variants section is moved over. Currently only works for Checkbox.

PR #4941, opened 2025-12-09
React support guides for Vaadin 25

peholmst:
This PR contains all the old Hilla guides from Building Apps. They need to be edited so that they fit into a separate *React support* section.


peholmst:
Handled in https://github.com/vaadin/docs/pull/5750

PR #4944, opened 2025-12-09
refactor: simplify scroller basic example

sissbruecker:
Simplify the basic Scroller example to just show a card with some Markdown content. This makes the example work out of the box with both Lumo and Aura.


vursen (approved):
The overflow indicators look a bit strange in this example, do we actually need them? Otherwise LGTM.

Image

PR #4947, opened 2025-12-10
docs: Select Aura style variants and props

sissbruecker:
List style variants and props specific to Lumo / Aura.


sissbruecker:
No, it just listed all item prefixed props for now. It's maybe lacking a bit of context, but I still added it.

PR #4949, opened 2025-12-10
docs: Text Field Aura style variants and props

sissbruecker:
List style variants and props specific to Lumo / Aura.

### DIFF
--- a/articles/components/text-area/index.adoc
+++ b/articles/components/text-area/index.adoc
@@ -16,7 +16,7 @@ page-links:
 Text Area is an input field component that allows entry of multiple lines of text.
 // end::description[]
 
-[.example]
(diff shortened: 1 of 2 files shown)


articles/components/text-field/styling.adoc
```diff
@@ -7,6 +7,35 @@ order: 50
 ---
 = Text Field Styling
 
+// Style Variants
+
+include::{articles}/components/_input-field-common-features.adoc[tags=styles-start;text-alignment;small-variant;helper-above-field;styles-end]
+
+[.example]
```
vursen: **nit:** It would be helpful to have a short paragraph explaining which theme variants are shown in this example.

PR #4952, opened 2025-12-10
refactor: prevent FOUC when switching from Lumo to Aura in production

vursen:
This PR adds [`root.__lumoInjector.forceUpdate()`](https://github.com/vaadin/web-components/pull/10588) to `applyTheme` to force LumoInjector to re-evaluate and update injected styles synchronously, without waiting for `transitionstart`. This prevents a flash of unstyled content when switching from Lumo to Aura using the theme switcher in production.

### DIFF
--- a/articles/components/text-area/index.adoc
+++ b/articles/components/text-area/index.adoc
@@ -16,7 +16,7 @@ page-links:
 Text Area is an input field component that allows entry of multiple lines of text.
 // end::description[]
 
-[.example]
+[.example,themes="lumo,aura"]
 --
 ifdef::lit[]
 [source,html]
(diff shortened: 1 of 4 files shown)


frontend/demo/theme.ts
```diff
@@ -44,6 +44,10 @@ export function applyTheme(root: DocumentFragment | DocumentOrShadowRoot | HTMLE
       }
 
       root.adoptedStyleSheets = [...adoptedStyleSheets, docsStyleSheet];
+
+      // @ts-expect-error Force LumoInjector to update injected styles
+      // synchronously to avoid FOUC when switching from Lumo to Aura.
+      root.__lumoInjector?.forceUpdate();
```
tomivirkki: Should this be `.forceUpdate?.();` since the function isn't there yet?

PR #4971, opened 2025-12-11
docs: mention Grid#scrollToItem, link hierarchy formats with their article

vursen:
Upgrading Guide changes:

- Mentioned that Grid#scrollToItem no longer scrolls if the item is already visible.
- Linked the hierarchy formats with their corresponding documentation


articles/upgrading/index.adoc
```diff
@@ -561,7 +563,7 @@ The Time Picker no longer uses `vaadin-time-picker-combo-box` internally. This m
...
+Tree Grid's client-side approach to data loading has been refactored. Instead of requesting data for each hierarchy level separately, the web component now sends a single request for the visible range, and the server always returns the corresponding items as a flat list. On the server-side, on the other hand, the data provider can now choose to provide hierarchical data in one of two formats: the existing <<../components/tree-grid/data-providers#hierarchyformat-nested-default,`HierarchyFormat.NESTED`>> (default) or the new <<../components/tree-grid/data-providers#hierarchyformat-flattened,`HierarchyFormat.FLATTENED`>>. These updates collectively introduce breaking changes, which are described below.
```
vursen:
```suggestion
Tree Grid's client-side approach to data loading has been refactored. Instead of requesting data for each hierarchy level separately, the web component now sends a single request for the visible range, and the server always returns the corresponding items as a flat list. On the server side, on the other hand, the data provider can now choose to provide hierarchical data in one of two formats: the existing <<../components/tree-grid/data-providers#hierarchyformat-nested-default,`HierarchyFormat.NESTED`>> (default) or the new <<../components/tree-grid/data-providers#hierarchyformat-flattened,`HierarchyFormat.FLATTENED`>>. These updates collectively introduce breaking changes, which are described below.
```

PR #4985, opened 2025-12-14
Document ColorScheme usage for dark and light scheme selection

bramvonk:
See https://github.com/vaadin/flow/issues/22988#issuecomment-3651934862

### DIFF
--- a/articles/upgrading/index.adoc
+++ b/articles/upgrading/index.adoc
@@ -260,7 +260,8 @@ The `themeFor` parameter of the [classname]`@CssImport` annotation (for shadow-D
 
 The special [filename]`document.css` file (for loading styles into the document root in embedded components) is removed as no longer necessary.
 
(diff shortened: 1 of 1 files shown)


articles/upgrading/index.adoc
```diff
@@ -260,6 +260,8 @@ The `themeFor` parameter of the [classname]`@CssImport` annotation (for shadow-D
 
 The special [filename]`document.css` file (for loading styles into the document root in embedded components) is removed as no longer necessary.
 
+You can use [classname]`@ColorScheme` for choosing between light or dark color scheme.
```
caalador: This could also have a link to the documentation here so one can read further on the feature and how it works.
```suggestion
You can use [classname]`@ColorScheme` for choosing between light or dark color scheme. 
See <<{articles}/styling/themes#color-schemes, Color Schemes>> for more details on using color schemes.
```

PR #5005, opened 2025-12-18
docs: add note about context protocol

mcollovati:
When VaadinServlet is configured with a custom mapping, stylesheet must be loaded from the application context path, by adding the context:// protocol to the path used with  annotation.


mshabarov (commented):
Looks good, but we also need this in the styling article for a long-term https://vaadin.com/docs/latest/styling/stylesheets

mcollovati:
> Looks good, but we also need this in the styling article for a long-term https://vaadin.com/docs/latest/styling/stylesheets

Update also the stylesheets guide

PR #5091, opened 2026-01-28
Update docs with correct information about detailed report

szolo:
Add more information about full details report.


dpizzagalli (approved):
This is the same sentence used in the UI in MTK for Eclipse - Approved for consinstency.

Issue #5098, opened 2026-01-28
Make Vaadin pre-release documentation available


paulroemer:
For our testers we want to provide the documentation for the upcoming Vaadin version.

We decided to introduce /pre-release that will always point to the next Vaadin minor.

jouni:
Would be nice to avoid dashes in URLs, so I’d suggest `/next` instead of `/pre-release`, like we’ve talked about before.

PR #5099, opened 2026-01-28
#5098: Introduce /pre-release

paulroemer:
Introduce /pre-release that will always point to the next minor version of Vaadin


Artur-:
Why should this go into 25.0?

peholmst:
@Artur- Because you want to be able to select the pre-release from the menu in 25.0 as well?

PR #5101, opened 2026-01-29
chore: add note on css imports in add-ons

caalador:
Add a note on using implicit relative
imports in add-on css as non relative
breaks may fail resoving
(for instance with taiwindcss)


mcollovati (approved):
Looks good, but this page is a bit hard to find IMO.
I wonder if adding a note also to https://vaadin.com/docs/latest/flow/component-internals/web-components would help.

caalador:
Added note and link to debugging page

PR #5115, opened 2026-01-30
Add update_latest_stable_tag.yml

peholmst:
A GitHub action that always tags the latest commit in the latest `vX.Y` branch as `latest_stable` (because `latest` has already been used in the past for a branch).

Purpose: to avoid having to reconfigure the deployment job every time a new minor version is released. It should be enough that the CI server deploys the `latest_stable` tag to `/docs/latest`.


Artur-:
Is this still relevant?

peholmst:
@paulroemer Is this something we still need?

PR #5140, opened 2026-02-05
fix: Add a note about caching in computed signals

mshabarov:


articles/flow/ui-state/building-ui.adoc
```diff
@@ -85,6 +85,8 @@ fullName.bindText(Signal.computed(() ->
     firstName.value() + " " + lastName.value()));
 ----
 
+The lambda variant executes the callback every time the signal value is changed and doesn't cache the computed result. In contrast, an explicit computed signal (using [methodname]`Signal.computed()`) caches the result until any dependency signal value changes. For simple operations like [methodname]`String::length`, a lambda is enough. However, for complex or expensive computations, use an explicit computed signal to avoid unnecessary recalculations.
```
Legioth: Both variants run the callback when the value is changed. The difference is rather in what happens when someone reads the signal value.

mshabarov: pushed a new wording, is it accurate enough now?

PR #5193, opened 2026-02-18
Fix style issues across building-apps documentation section

Artur-:
Expand contractions to full forms, fix grammar errors, and apply
style consistency improvements across 75 files in the building-apps
documentation section.


jouni:
Vale warnings are never black and white. Errors are hard rules that should be fixed.

It’s okay to use future tense when needed. The rule is there to avoid using it unnecessarily/excessively. See https://developers.google.com/style/tense for more guidance.

PR #5216, opened 2026-02-24
docs: update Badge docs to use new <vaadin-badge> component

web-padawan:
- Updated Badge to use new `vaadin-badge` component.
- Added Styling page for the badge component


jouni:
Is the "dot" variant something we agreed to add? At least the description in the docs is different from the example itself.  The description refers to this experiment I did in the dev pages, but didn't actually think we would add to the component: 
Screenshot 2026-03-03 at 13 01 08

The example itself looks like the previous implementation of empty badges. But that’s also not something I would implement like that anymore (not with the `:empty` selector).

web-padawan:
> At least the description in the docs is different from the example itself.

My bad 🤖  Fixed it to describe what the "dot" variant actually does.

PR #5228, opened 2026-02-25
docs: update notification examples to support both Lumo and Aura

web-padawan:
- Replaced `LUMO_` variants with new unprefixed variants that are supported in Aura
- Removed "contrast" variant mention and replaced usage of "contrast" in other examples
- Replaced "primary" variant example with "info" variant supported by Lumo and Aura
- Replaced `margin` with `slot="end"` to make buttons in notifications look better
- Added `min-width` on layouts to add space between text and buttons in Lumo
- Removed usage of `tertiary-inline` variant from buttons in notification examples
- Updated `notifications-popup` example - replaced ContextMenu with Popover
- Simplified the `notification-rich` example (last notification didn't look good in Aura)
  - IMO, 3 notifications with icons and avatar is enough to showcase rich content
- Added theme switcher, except for `notifications-popup` example using badge
  - Will be added in https://github.com/vaadin/docs/pull/5216


articles/components/notification/styling.adoc
```diff
@@ -23,15 +23,19 @@ Notification supports the following style variants.
...
 |Used to display warnings
-|Lumo
+|Lumo, Aura
 
 |`error`
 |Used to display errors
-|Lumo
+|Lumo, Aura
+
+|`info`
```
sissbruecker: Is there any benefit to still listing `primary`? I assume `info` should be used instead from now on?

PR #5234, opened 2026-02-25
refactor: update HorizontalLayout examples to support both Lumo and Aura

vursen:
- Added theme switcher to all HorizontalLayout examples on main and styling pages
- Simplified spacing documentation: removed the spacing variants table and render-only demo, replaced with concise inline code showing custom gap values and CSS custom properties
- Deleted `horizontal-layout-spacing-variants.ts` demo that showcased Lumo-specific spacing variants
- Added reference links to base gap properties and Lumo space properties for spacing customization
- Increased the padding for `.example-item` to make the "Wrapping Horizontally Aligned Items" example render correctly with Aura

Depends on 
- https://github.com/vaadin/docs/pull/5255

### DIFF
--- a/articles/components/login/index.adoc
+++ b/articles/components/login/index.adoc
@@ -65,7 +65,7 @@ The basic Login component consists of a title (i.e., "Log In"), two input fields
 
 You can customize the form's title and labels using internationalization.
 
-[.example]
+[.example,themes="lumo,aura"]
 --
 
 ifdef::lit[]
@@ -92,7 +92,7 @@ endif::[]
 
 The basic Login component can be used to create log-in pages featuring rich content.
 
-[.example]
+[.example,themes="lumo,aura"]
 --
 
 ifdef::lit[]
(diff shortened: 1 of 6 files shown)


frontend/themes/docs/basic-layouts.css
```diff
@@ -31,6 +31,21 @@
   background: var(--vaadin-background-container-strong);
   color: var(--vaadin-text-color);
   border-radius: 0.5em;
-  padding: var(--vaadin-padding-s) var(--vaadin-padding-l);
   white-space: nowrap;
 }
+
+.basic-layouts-example[theme~='lumo'] {
```
rolfsmeds: This selector doesn't work on the Flow samples: the theme switcher applies the attrib on the exported Flow component WC's root, not to the `.basic-layouts-example` element, so the padding is not getting applied at all.

rolfsmeds (approved):

PR #5239, opened 2026-02-26
docs: update Dialog examples to support both Lumo and Aura

web-padawan:
- Replaced `LUMO_` variants with unprefixed variants supported by Aura
- Replaced `--lumo-space-m` custom CSS property with `--vaadin-gap-m`


src/main/java/com/vaadin/demo/component/dialog/DialogHeader.java
sissbruecker: This one could be improved to remove the border radius of the grid. Though I'm not sure how, `--vaadin-grid-border-radius` removes the round borders at the bottom, but the top ones remain.

Image

PR #5258, opened 2026-02-27
docs: update Login examples to support both Lumo and Aura

web-padawan:
- Updated Login examples to support Lumo and Aura, modified rich content example to set `color-scheme`
- Removed statement about not being able to add components while opened (this is possible in Vaadin 25)

### DIFF
--- a/frontend/themes/docs/login-rich-content.css
+++ b/frontend/themes/docs/login-rich-content.css
@@ -9,7 +9,9 @@
   align-items: center;
   display: flex;
   max-width: 300px;
+  background: var(--vaadin-background-color);
 }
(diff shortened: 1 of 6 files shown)


frontend/themes/docs/login-rich-content.css
```diff
@@ -9,7 +9,9 @@
   align-items: center;
   display: flex;
   max-width: 300px;
+  background: var(--vaadin-background-color);
 }
-.login-rich-content vaadin-login-form-wrapper {
```
web-padawan: Removed this as not needed, `vaadin-login-form-wrapper` is placed in shadow DOM in V25.

PR #5261, opened 2026-02-27
docs: update Select examples to support both Lumo and Aura

web-padawan:
Replaced Lumo custom CSS properties, removed `small` variant usage, added theme switchers.


vursen:
Is this difference in the select dropdown appearance intentional?

**Section: Custom Item Label**

Screenshot 2026-03-02 at 10 35 36

**Section: Custom Item Presentation**

Screenshot 2026-03-02 at 10 35 45

web-padawan:
> Is this difference in the select dropdown appearance intentional?

One example is using `img` with fixed width `2.25rem` and another one is using `vaadin-avatar`.
Updated to use `person-item` class for both examples, otherwise IMO we can keep `img`.

PR #5279, opened 2026-03-02
docs: document MDL Aura style variant, add theme switcher

web-padawan:
Added theme switcher to the MDL vertical example.


articles/components/master-detail-layout/index.adoc
sissbruecker: There's actually a theme variant for Aura that's not documented, so maybe that could be added.

PR #5323, opened 2026-03-05
docs: update examples to use end slot instead of inline styles

web-padawan:
Updated to use `slot="end"` in HorizontalLayout to remove inline styles setting `margin` to `auto`.


src/main/java/com/vaadin/demo/component/button/ButtonGrid.java
```diff
@@ -36,7 +36,6 @@ public ButtonGrid() {
         Button delete = new Button("Delete");
         delete.setEnabled(false);
         delete.addThemeVariants(ButtonVariant.ERROR);
-        delete.getStyle().set("margin-inline-start", "auto");
```
web-padawan: This is no longer needed as we use `footer.addToEnd(delete)` below.

PR #5381, opened 2026-03-24
docs: document security annotations required for custom error views

### LINKED ISSUE #5378: Improve Spring Security, VaadinSecurityConfigurer and NotFoundException documentation for Vaadin 25
The existing documentation regarding Spring Security and Vaadin is generally sufficient, but it is not yet clear regarding NotFoundException. Since the default for any unknown request is now "deny all" it should be clarified in relation to "...CustomNotFoundTarget extends RouteNotFoundError..." that this configuration alone is NOT sufficient to trigger such a message. Additionally, the VaadinSecurityConfigurer has to be customized too. Please update the documentation.

Artur-:
Custom error handler views like RouteNotFoundError subclasses need security annotations (e.g., @AnonymousAllowed) when using VaadinSecurityConfigurer, which denies all unannotated views by default. Without this, users see a generic error page instead of their custom not-found page.

Adds cross-references between the routing exceptions, protect views, security configurer, and enabling security pages so users can discover this requirement from any entry point.

Fixes vaadin/docs#5378


mcollovati (changes requested):
https://github.com/vaadin/docs/pull/5381#issuecomment-4124248797

mcollovati:
@Artur- can you take a look at https://github.com/vaadin/docs/pull/5381#issuecomment-4124248797?

mcollovati (approved):

PR #5456, opened 2026-04-07
docs: add example with nested MDLs

vursen:
Part of https://github.com/vaadin/web-components/issues/11348


web-padawan (approved):
LGTM. There is no "Drag to resize" in the new example, but that applies to existing examples as well and in production, the React example will be actually rendered.

PR #5474, opened 2026-04-14
docs: Document how to compose MessageInput with modular upload for file attachments

Artur-:
Add a File Attachments section to the MessageInput docs with a live example showing how to compose UploadManager, UploadButton, and UploadFileList with MessageInput for a chat-with-attachments experience.

Add a Chat Input with Attachments section to the Modular Upload docs with the same example. Add cross-links between MessageInput, MessageList, Modular Upload, and AI file-attachments pages so users can discover the integration regardless of which page they start from.

### DIFF
--- a/src/main/java/com/vaadin/demo/component/messages/MessageInputUpload.java
+++ b/src/main/java/com/vaadin/demo/component/messages/MessageInputUpload.java
@@ -1,0 +1,90 @@
+package com.vaadin.demo.component.messages;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.vaadin.demo.DemoExporter; // hidden-source-line
+import com.vaadin.flow.component.html.Div;
+import com.vaadin.flow.component.messages.MessageInput;
+import com.vaadin.flow.component.messages.MessageList;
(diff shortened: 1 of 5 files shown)


src/main/java/com/vaadin/demo/component/messages/MessageInputUpload.java
sissbruecker, line 1: The example doesn't work, upload button and the file list are not rendered / registered as custom components.

src/main/java/com/vaadin/demo/component/messages/MessageInputUpload.java
```diff
@@ -0,0 +1,90 @@
...
+
+        messageInput.addSubmitListener(event -> {
+            var message = new MessageListItem(event.getValue(),
+                    Instant.now(), "You");
+            message.setUserColorIndex(1);
+
+            // Add pending files as attachments
+            for (var entry : pendingFiles.entrySet()) {
+                message.addAttachment(new MessageListItem.Attachment(
+                        entry.getKey(), "#", "application/octet-stream"));
```
sissbruecker, line 61: This part doesn't look helpful. It's not clear how to convert something received from the upload handler into a URL that you can provide here. Also the mime type doesn't match what was uploaded.

PR #5475, opened 2026-04-14
docs: Add modular upload to the AI chat example

Artur-:
Update MessageListAiChat to include UploadManager, UploadButton, and UploadFileList so the AI chat demo shows file attachment support. Pending files are added as attachments to the user message on submit.


HerbertsVaadin (changes requested):
The upload button in the flow example is very off. 
There should maybe be an equivalent for lit and react examples (only with fixed designs of course). 

Image

PR #5478, opened 2026-04-15
feat: document playwright usage with load test

caalador:


tltv (commented):
Many chapters are duplicating with main Load Testing article. More specifically: Combined Scenarios, What Gets Generated, Think Time, Understanding k6 Output. These can be just linked to the main article.

Also `loadtest:run` goal section is identical and that part can be also linked to the main article.

Is it necessary to instruct here to create another module with pom.xml or is it enough to just show example plugin setup without suggesting anything regarding the project structure?

caalador:
Updated to not have common and only playwright info.

articles/flow/testing/load-testing/playwright.adoc
```diff
@@ -0,0 +1,406 @@
...
+
+<dependency>
+    <groupId>com.microsoft.playwright</groupId>
+    <artifactId>playwright</artifactId>
+    <version>1.58.0</version>
+    <scope>test</scope>
+</dependency>
+----
+
+NOTE: The Playwright path uses `testbench-loadtest-support` (which provides [classname]`PlaywrightHelper`) instead of the `loadtest-helper` dependency used by the TestBench path.
```
tltv: Main article is not yet updated for the removed loadtest-helper module, this article shouldn't have this note here.

caalador: removed

tltv (approved):

PR #5480, opened 2026-04-16
fix: remove loadtest-helper

caalador:
loadtest-helper dependency
no longer exists, remove
any mentions of it from the
base documentation

### DIFF
--- a/articles/flow/testing/load-testing/index.adoc
+++ b/articles/flow/testing/load-testing/index.adoc
@@ -62,7 +62,6 @@ Add the load test support library as a dependency:
 <dependency>
     <groupId>com.vaadin</groupId>
     <artifactId>testbench-loadtest-support</artifactId>
-    <optional>true</optional>
 </dependency>
 ----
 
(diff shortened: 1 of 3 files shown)


articles/flow/testing/load-testing/index.adoc
```diff
@@ -500,21 +493,9 @@ When using <<running,combined scenarios>>, each scenario's CSV data is loaded wi
...
-<dependency>
-    <groupId>com.vaadin</groupId>
-    <artifactId>loadtest-helper</artifactId>
-    <scope>runtime</scope>
-</dependency>
-----
-
-[NOTE]
-The <<setup,project setup>> section shows `loadtest-helper` with `test` scope for using [classname]`LoadTestItHelper` in test code. If you also want error propagation and `vaadin.view.count` metrics at runtime, add the dependency a second time with `runtime` scope, or use the default (compile) scope to cover both cases. Only include the `runtime` dependency in builds used for load testing, as it changes error handling behavior.
+The <<setup,project setup>> section configures `testbench-loadtest-support` as `optional`, which makes it available both in test code and at runtime. Only include this dependency in builds used for load testing, as it changes error handling behavior.
```
tltv: `optional` leaves the dependency out from the production package, unless explicitly added in via profile or some other way. Maybe this should be clarified here to avoid mistakes.

caalador: Removed optional all together as there is a mention that the build with this should only be used for loadtest.

caalador:
updated the goalPrefixes

tltv (approved):

PR #5484, opened 2026-04-17
docs: update Upgrading Guide for Binder

### LINKED ISSUE #24109: Highlight changed Binder behavior for hidden fields in Vaadin 25 Upgrade guide
### Describe your motivation

We're currently upgrading from Vaadin 24 to Vaadin 25 and run into a couple of issues due to bindings not being processed properly anymore when fields are not visible. 

I understand this was a conscious change that was introduced with `isApplied` and `isAppliedPredicate` in the `Binding` class. However, we personally consider this a rather breaking change and were surprised to not find any mention about this in the [upgrade guide](https://vaadin.com/docs/latest/upgrading).

### Describe the solution you'd like

Please add this to the upgrade guide, plus a recommendation how to restore the old behavior. We have so many forms and going through all of them would require a lot of time.

I guess the code below is a possible workaround, but it would be great to have some kind of property on the binder that can be easily set.

```java
getBindings().forEach(binding -> binding.setIsAppliedPredicate(_ -> true));
```

tltv:
Document breaking change in Binder regarding hidden fields binding being excluded by default. Document new method Binder#setAcceptHiddenFields(boolean) to change the default behavior.

Fixes: vaadin/flow#24109


tltv:
This change depends on https://github.com/vaadin/flow/pull/24139.

articles/flow/binding-data/components-binder-validation.adoc
```diff
@@ -163,7 +163,7 @@ binding.setValidatorsDisabled(true);
 
 Sometimes you need to exclude certain bindings from both validation and bean writing based on runtime conditions. For example, when a form field is hidden based on user selections, you may want to exclude its binding entirely. The [methodname]`isApplied()` method on [classname]`Binding` determines whether a binding participates in these operations.
 
-By default, bindings for non-visible components are automatically excluded. You can customize this behavior by providing a predicate:
+By default, bindings for non-visible components are automatically excluded. To include all hidden fields at the binder level, call [methodname]`binder.setAcceptHiddenFields(true)` (since V25.2). For finer-grained control, you can customize individual bindings by providing a predicate, which takes precedence over the binder-level setting:
```
mshabarov:
```suggestion
By default, bindings for non-visible components are automatically excluded. To include all hidden fields at the binder level, call [methodname]`binder.setAcceptHiddenFields(true)`. For finer-grained control, you can customize individual bindings by providing a predicate, which takes precedence over the binder-level setting:
```

mshabarov: I'd instead backport this method to 25.1 and 25.0.

articles/upgrading/index.adoc
```diff
@@ -1041,6 +1041,8 @@ The [methodname]`getPropertyString`, [methodname]`getPropertyBoolean`, [methodna
 == Binder
 [methodname]`Binder.validate()` implementation has been changed to behave as its Javadoc states. In other words, [methodname]`Binder.validate()` no longer fails when bean level validators have been configured but no bean is currently set (i.e. [classname]`Binder` is used in buffered mode).
 
+[methodname]`Binder.writeBean()` and [methodname]`Binder.writeBeanIfValid()` no longer process hidden fields. Previously, all bound fields were validated and written to the bean regardless of visibility. Now, bindings for non-visible components are skipped by default. To restore the previous behavior, call [methodname]`binder.setAcceptHiddenFields(true)` (since V25.2), or set a per-binding predicate with [methodname]`binding.setIsAppliedPredicate(b -> true)`. See <<{articles}/flow/binding-data/components-binder-validation#conditionally-applying-bindings,Conditionally Applying Bindings>> for details.
```
mshabarov:
```suggestion
[methodname]`Binder.writeBean()` and [methodname]`Binder.writeBeanIfValid()` no longer process hidden fields. Previously, all bound fields were validated and written to the bean regardless of visibility. Now, bindings for non-visible components are skipped by default. To restore the previous behavior, call [methodname]`binder.setAcceptHiddenFields(true)`, or set a per-binding predicate with [methodname]`binding.setIsAppliedPredicate(b -> true)`. See <<{articles}/flow/binding-data/components-binder-validation#conditionally-applying-bindings,Conditionally Applying Bindings>> for details.
```

articles/flow/binding-data/components-binder-validation.adoc
tltv: Makes sense. Method name need to be updated.

tltv: It will introduce two new public methods in the Binder API however.

mshabarov (approved):

PR #5485, opened 2026-04-17
Docs/badge number and dot examples

rolfsmeds:
- Added a Number-Only section with a sample under Numbers.
- Added a button-with-dot to the Dot variant example to showcase usage. (Yeah, it's a bit awkward that you need to add some css to position it reasonably, but we'll address that later.)


jouni (approved):
Tested locally 👍

web-padawan (approved):

PR #5495, opened 2026-04-21
docs: add note on naming in load testing

tltv:
Adds note to use `IT` suffix, not `Test`. Classes ending in `Test` are picked up during the `test` phase, before the server starts.

### DIFF
--- a/articles/flow/testing/load-testing/index.adoc
+++ b/articles/flow/testing/load-testing/index.adoc
@@ -653,6 +653,8 @@ Below is a complete example that starts the application server, records two Test
 
 Running `mvn verify` with this configuration executes the full workflow: start server, record scenarios, run load tests, and stop server.
 
+[NOTE]
+Name your test classes with the `IT` suffix, not `Test`. Classes ending in `Test` are picked up during the `test` phase, before the server starts.
 
 [[remote-testing]]
 == Remote Load Testing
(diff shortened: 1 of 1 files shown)


articles/flow/testing/load-testing/index.adoc
```diff
@@ -653,6 +653,8 @@ Below is a complete example that starts the application server, records two Test
 
 Running `mvn verify` with this configuration executes the full workflow: start server, record scenarios, run load tests, and stop server.
 
+[NOTE]
+Name your test classes with the `IT` suffix, not `Test`. Classes ending in `Test` are picked up during the `test` phase, before the server starts.
```
caalador: As we are talking maven execution and most likely failsafe-plugin for running integration tests in the integration-test phase we should note:
`the failsafe plugin executes test methods in classes whose name starts with IT, or ends with IT or ITCase.`

tltv: Maven context is already included in this chapter since it's about running `mvn verify`.

Suggesting to name tests with IT-suffix is mainly to avoid having the problematic Test-suffix, which are run test too early in the `test` phase. If I remove the IT-suffix, it works too because the `record` goal runs in `integration-test` phase and tests are listed in plugin configuration.

Should it be more verbose anyway? It leaves out maven and failsafe-plugin details, but those can be added if you still think it would be better.

caalador: Perhaps just add there to check failsafe-plugin and surefire-plugin documentation for more details.

caalador: https://maven.apache.org/surefire/maven-failsafe-plugin/examples/inclusion-exclusion.html

https://maven.apache.org/surefire/maven-surefire-plugin/examples/inclusion-exclusion.html

tltv: Added links.

caalador (approved):

PR #5497, opened 2026-04-21
docs: clarify HierarchicalQuery fields for nested and flattened formats

vursen:
Follow-up to https://github.com/vaadin/flow-components/issues/9148#issuecomment-4286462039

### DIFF
--- a/articles/components/tree-grid/data-binding.adoc
+++ b/articles/components/tree-grid/data-binding.adoc
@@ -254,10 +254,17 @@ Both formats have their advantages and disadvantages. When implementing a custom
 
 The nested hierarchy format refers to a data provider implementation that, in each request, returns only the direct children of the requested parent item.
 
-Tree Grid fills the viewport by sending requests to the data provider based on the current scroll position and viewport size. Each request includes a [classname]`HierarchicalQuery` object, where:
+To fill the viewport, Tree Grid sends separate requests to the data provider for each expanded item within the visible range. Each request comes with a [classname]`HierarchicalQuery` object that contains the following fields:
 
-1. [methodname]`HierarchicalQuery#getParent()` specifies the parent item whose direct children are to be fetched.
-2. [methodname]`HierarchicalQuery#getOffset()` and [methodname]`HierarchicalQuery#getLimit()` define the range of direct children to return.
+|===
+|Method |Description
+
+|`HierarchicalQuery#getParent()`
+|The parent item whose direct children should be returned.
+
+|`HierarchicalQuery#getOffset()` and `getLimit()`
+|The offset and limit defining the range of direct children to return.
+|===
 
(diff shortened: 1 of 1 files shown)


articles/components/tree-grid/data-binding.adoc
```diff
@@ -357,13 +364,25 @@ public record Folder(String name) {};
...
+|`HierarchicalQuery#getOffset()` and `getLimit()`
+|The offset and limit defining the slice of the resulting flat list to return.
 
-1. [methodname]`HierarchicalQuery#getParent()` specifies the parent item whose sub-tree to be fetched as a flat list. The `null` value means the entire tree starting from the root items.
-2. [methodname]`HierarchicalQuery#getExpandedItemIds()` provides a set of item IDs that are currently expanded.
-3. [methodname]`HierarchicalQuery#getOffset()` and [methodname]`HierarchicalQuery#getLimit()` define the slice of the flat list to return.
+|===
 
-Additionally, the data provider must implement [methodname]`HierarchicalDataProvider#getDepth(T item)` to return the depth for each item in the hierarchy. Tree Grid uses this information to indent items visually.
+The data provider must also implement [methodname]`HierarchicalDataProvider#getDepth(T item)` to return the depth for each item in the hierarchy. Tree Grid uses this information to indent items visually.
```
vursen: **open question:** should this be added? Or is it already clear?

```suggestion
[WARNING]
====
[methodname]`HierarchicalQuery#getExpandedItemIds()` must not be used to determine the full expansion state of the Tree Grid in the data provider. It's only guaranteed to contain IDs of items that are relevant to the current request.
====

The data provider must also implement [methodname]`HierarchicalDataProvider#getDepth(T item)` to return the depth for each item in the hierarchy. Tree Grid uses this information to indent items visually.
```

sissbruecker: I think it's worth adding. It could use simpler language by stating that it does not necessarily contain all expanded items of the tree, but only those relevant to the current request. "Full expansion state" is something that you may have to think about what is meant by that.

It's not clear for me from this context, but the current wording also sounds like the use-case that triggered this change was doing _something_ with `getExpandedItemIds` that required all expanded items? If that something is a valid use case then it could mention the alternative, as in, how do I get all expanded items. Otherwise disregard.

vursen:
```suggestion
[NOTE]
====
[methodname]`HierarchicalQuery#getExpandedItemIds()` does not necessarily contain all expanded items in the Tree Grid. It's only guaranteed to contain IDs of items relevant to the current request and may also be empty.
====

The data provider must also implement [methodname]`HierarchicalDataProvider#getDepth(T item)` to return the depth for each item in the hierarchy. Tree Grid uses this information to indent items visually.
```

vursen: I've rephrased the description of `getExpandedItemIds` instead, which should make this note unnecessary. Let me know what you think

sissbruecker (approved):
Looks good. Consider cherry-picking it to the current version.

PR #5529, opened 2026-05-03
Add Page Visibility API reference documentation

Artur-:
Document the server-side pageVisibilitySignal() API on Page, covering the four PageVisibility states, signal subscription patterns, and four common use cases: pausing work while hidden, routing notifications, refreshing stale data on return, and broadcasting presence.


mcollovati (approved):
Made some minor changes and added the discussion-id to the page

PR #5572, opened 2026-05-08
docs: cleanup tree grid data binding docs

ugur-vaadin:
Updates tree grid data binding section to make small corrections:

- The `TreeData#addRootItems` reference in the Adding Items section now matches the varargs overload that the example actually calls.
- The Filtering Items paragraph in the Tree Data Provider section is rewritten to describe the contracted behaviour end-to-end: the filter is evaluated against the entire hierarchy (including collapsed branches), the parent chain of any matching descendant is preserved, and a nested match only becomes visible once its ancestors are expanded.
- The two custom-data-provider examples now use `extends` instead of `implements` for `AbstractHierarchicalDataProvider`, which is an abstract class.

Part of https://github.com/vaadin/flow-components/issues/1886


articles/components/tree-grid/data-binding.adoc
```diff
@@ -201,7 +201,7 @@ searchInput.addValueChangeListener(event -> {
 
 The `setFilter` method re-renders the Tree Grid automatically, so calling `refreshAll` isn't necessary.
 
-When filtering, Tree Data Provider evaluates the entire expanded hierarchy. If a nested item matches the filter, its full chain of parent items is also included. This ensures that the matching item appears in the filtered view even if some of its parents don't match the filter themselves.
+When a filter is set, Tree Data Provider evaluates it against every item in the hierarchy, including items in collapsed branches. A parent item is kept in the result whenever any of its descendants match, so the chain of ancestors leading to a match is never filtered out — even when those ancestors don't match the filter themselves. The match itself only appears in Tree Grid once all of its ancestors are expanded, since setting a filter doesn't change Tree Grid's expansion state.
```
vursen:
```suggestion
When a filter is set, Tree Data Provider evaluates it against every item in the tree, including items in collapsed branches. If any item matches, all of its ancestors are included in the result, so the whole path from the root to the matching item is preserved — even if those ancestors don't match themselves. However, setting a filter doesn't expand any collapsed items automatically, so only the expanded part of the path is displayed in Tree Grid.
```

articles/components/tree-grid/data-binding.adoc
```diff
@@ -201,7 +201,7 @@ searchInput.addValueChangeListener(event -> {
 
 The `setFilter` method re-renders the Tree Grid automatically, so calling `refreshAll` isn't necessary.
 
-When filtering, Tree Data Provider evaluates the entire expanded hierarchy. If a nested item matches the filter, its full chain of parent items is also included. This ensures that the matching item appears in the filtered view even if some of its parents don't match the filter themselves.
+When a filter is set, Tree Data Provider evaluates it against every item in the tree, including items in collapsed branches. If any item matches, all of its ancestors are included in the result, so the whole path from the root to the matching item is preserved — even if those ancestors don't match themselves. However, setting a filter doesn't expand any collapsed items automatically, so only the expanded part of the path is displayed in Tree Grid.
```
vursen:
```suggestion
When a filter is set, Tree Data Provider evaluates it against every item in the tree, including items in collapsed branches. If any item matches, all of its ancestors are included in the result, so the whole path from the root to the matching item is preserved, even if those ancestors don't match themselves. However, setting a filter doesn't expand collapsed items automatically, so only the currently expanded part of the path is displayed in Tree Grid.
```

vursen (approved):

PR #5581, opened 2026-05-11
docs: document browserless test improvements

mcollovati:
Documents introduction of JUnit 6 extension and API for multi-user / multi-window testing.


articles/flow/testing/browserless/locators.adoc
```diff
@@ -0,0 +1,180 @@
...
+
+
+= Component Locators
+
+[since:com.vaadin:vaadin@V25.2]
+
+A *locator* is a fluent object that combines a [classname]`ComponentQuery` filter chain with the actions of a [classname]`ComponentTester`. A single expression replaces the common [methodname]`test(find(...).single())` pattern and the locator is reusable across UI changes -- its resolution cache rewinds automatically when filters are reapplied, or explicitly via [methodname]`invalidate()`.
+
+[NOTE]
+Locator entry points are currently available on [classname]`BrowserlessUIContext` (the multi-user/multi-window API). For tests that extend [classname]`BrowserlessTest` or use the JUnit 6 extensions, continue to use <<component-query#, component queries>> with [methodname]`test()`.
```
mshabarov: Should we limit old base classes this way? Can we add locators to these old classes?

articles/flow/testing/browserless/multi-user.adoc
```diff
@@ -0,0 +1,346 @@
...
+
+| [classname]`BrowserlessUserContext`
+| One [classname]`VaadinSession` and the security state of a single logical user. A user can own many windows.
+
+| [classname]`BrowserlessUIContext`
+| One [classname]`UI` instance -- a single browser window. Provides the testing DSL ([methodname]`navigate`, [methodname]`find`, [methodname]`test`, ...).
+|===
+
+[NOTE]
+All three contexts are *thread-affine*: they must be created, used, and closed on the same thread. Driving the same context from parallel test threads is not supported.
```
mshabarov: Maybe this is better to reword so as to say "These three contexts are not thread-safe and tests using them shall run sequentially" ?

mcollovati: "tests using them shall run sequentially" would be misleading. You can run tests in parallel and they will work fine. The problem is that in the same test you cannot spawn multiple threads working with the same contexts.

mcollovati: But let's find a more clear wording.

articles/flow/testing/browserless/locators.adoc
mcollovati: I'm not sure we should. At least I have not tested it, even if it would just be a matter of making the base class implement `Locators` interface.

articles/flow/testing/browserless/multi-user.adoc
mcollovati: Updated. Let me know if now it's more clear

articles/flow/testing/browserless/locators.adoc
mcollovati: We can change the docs to suggest implementing `Locators` in app classes for now, and if we get positive feedback, make it as a default in the next release.

mcollovati: I did a test applying the interface to `BaseBrowserlessTest` subclasses and it seems to work fine.
My only concern is that if we do this, we could potentially break existing tests that define `findButton` or similar helper methods. Probably a rare or nonexisting case, but there would be no workaround if we apply `Locators` on the base class, and the developer would be forced to change their code.
But we can do it for the JUnit extension API, that is new in 1.1.
In addition, I'll update the docs to show how to manually wire `Locators` to base classes.

mcollovati: https://github.com/vaadin/browserless-test/pull/86

articles/flow/testing/browserless/multi-user.adoc
mshabarov: Yes, looks better now!

articles/flow/testing/browserless/locators.adoc
mshabarov: Good idea and compromise to apply `Locators` to the new extension classes. Let's review that. I think the current docs can be merged right away.

mshabarov (approved):

PR #5585, opened 2026-05-14
docs: document license-check Java prefs workaround

mcollovati:
The Vaadin license checker uses the Java `Preferences` API during license validation. In hardened environments where the JVM user has no writable home directory -- for example, a locked-down Tomcat service user on Linux -- the JVM repeatedly logs "Could not lock User prefs" warnings.

Add a troubleshooting entry under the License documentation that describes the symptom and the standard JVM workaround: setting the `-Djava.util.prefs.userRoot` system property to a writable directory. The entry also notes that Windows and macOS use different backing stores and aren't affected.


Artur-:
This is not needed anymore, right?

PR #5596, opened 2026-05-19
docs: document Element.addJsInitializer

Artur-:
Adds a section to the calling-javascript reference covering the new
scoped client-side lifecycle API: when to use it over executeJs, how
re-attach is handled, and the synchronous cleanup-return constraint.


articles/flow/component-internals/element-api/calling-javascript.adoc
```diff
@@ -241,4 +241,39 @@ getElement().executeJs(
...
+.Installing a listener with cleanup
+[example]
+====
+[source,java]
+----
+Registration registration = getElement().addJsInitializer("""
+        const handler = (event) => console.log('clicked', event.target);
+        this.addEventListener('click', handler);
+        return () => this.removeEventListener('click', handler);
+        """);
```
Legioth: Could show how to pass a parameter from the server, e.g. the `clicked` or `click` strings.

articles/flow/component-internals/element-api/calling-javascript.adoc
```diff
@@ -241,4 +241,39 @@ getElement().executeJs(
...
+
+
+=== Re-Attach Semantics
+
+The initializer is re-run after a real re-attach &ndash; when the element is removed from the DOM in one round trip and re-added in a later one, and so the browser receives a fresh DOM node. It is *not* re-run when the element is detached and re-attached on the server inside a single round trip, because the client never discarded its DOM in that case.
+
+
+=== Cleanup Constraints
+
+The return value of the expression is read synchronously. A function is treated as the cleanup callback; returning nothing (no `return`, or `undefined`) means there's no cleanup. Returning any other value &ndash; including a `Promise` resolving to a function &ndash; is logged as an error on the client. If you need asynchronous setup, do the async work inside the body and store the teardown in a synchronously returned function.
```
Legioth: The last sentence is quite confusing without an example. Not sure if it's important enough to add an example, or if it would be just as good to remove the sentence?

Legioth (changes requested):

articles/flow/component-internals/element-api/calling-javascript.adoc
```diff
@@ -273,7 +275,20 @@ The initializer is re-run after a real re-attach &ndash; when the element is rem
...
+If setup is asynchronous, don't return the `Promise`. Do the async work inside the body, keep a reference to whatever needs to be torn down, and return a synchronous cleanup function that uses it:
+
+[source,java]
+----
+getElement().addJsInitializer("""
+        let widget;
+        import('./my-widget.js').then((module) => {
+            widget = module.create(this);
+        });
+        return () => widget?.destroy();
```
Legioth: With this approach, a slow import leads to creating the widget but never destroying it.

I would recommend `const widgetPromise = ...` and then use something like  `async () => (await widgetPromise).destroy()` as the returned cleanup callback.

articles/flow/component-internals/element-api/calling-javascript.adoc
```diff
@@ -256,13 +256,15 @@ The expression syntax is the same as [methodname]`executeJs()`: `this` is the ho
...
-        return () => this.removeEventListener('click', handler);
-        """);
+        const handler = (event) => console.log($1, event.target);
+        this.addEventListener($0, handler);
+        return () => this.removeEventListener($0, handler);
+        """, "click", "clicked");
 ----
 ====
 
+As with [methodname]`executeJs()`, values passed after the expression are available as `$0`, `$1`, and so on. Here the event type (`"click"`) and the log label (`"clicked"`) come from the server rather than being hard-coded in the expression.
```
Legioth: This paragraph mostly just repeats exactly the same content as the paragraph before the code example. What might be meaningful here is to explicitly clarify that `$0` will be `click` and `$1` will be `clicked`.

Legioth (changes requested):

articles/flow/component-internals/element-api/calling-javascript.adoc
Artur-: Updated

articles/flow/component-internals/element-api/calling-javascript.adoc
Artur-: Updated

articles/flow/component-internals/element-api/calling-javascript.adoc
Legioth: I don't see any changes to this part

Legioth (approved):

PR #5602, opened 2026-05-22
docs: document MultiSelectComboBox collapse chips API

web-padawan:
Added examples for `collapseChips` API added in 25.2, with the checkbox to toggle it so see the difference.


rolfsmeds (dismissed):
Otherwise LGTM, but the sample is a bit confusing in that toggling collapseChips seems to reset the selection, which might make it hard to understand the feature.

web-padawan:
> toggling collapseChips seems to reset the selection

Fixed for both React and Lit examples (added selection change listeners to store selected countries).

vursen (approved):

PR #5611, opened 2026-05-27
Document vaadin-button-white-space variable

knoobie:
Add documentation for button white-space CSS variable.

### DIFF
--- a/articles/components/button/styling.adoc
+++ b/articles/components/button/styling.adoc
@@ -332,7 +332,7 @@ include::../_styling-section-theming-props.adoc[tag=style-properties]
 |`--vaadin-button-text-color`
 |Aura, Lumo
 
-|`--vaadin-button-white-space`
+|`--vaadin-button-label-wrap`
 |Aura (wrap), Lumo (nowrap) - since 25.2
 
 |===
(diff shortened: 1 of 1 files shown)


knoobie:
@jouni Thanks for your update :) "drafted" the docs update for it.. as reminder.. not sure if you wanna tackle it like this

articles/components/button/styling.adoc
```diff
@@ -332,6 +332,9 @@ include::../_styling-section-theming-props.adoc[tag=style-properties]
 |`--vaadin-button-text-color`
 |Aura, Lumo
 
+|`--vaadin-button-white-space`
```
jouni: I’m going to change the property name, to align with the existing dashboard property: https://github.com/vaadin/web-components/issues/11779#issuecomment-4554208581

```suggestion
|`--vaadin-button-label-wrap`
```

jouni (changes requested):

articles/components/button/styling.adoc
jouni: Renamed in #11864

jouni (approved):
I removed the "since" annotation. It's unnecessary as the main branch is for 25.2. 

I also removed the default values. No other property lists their default values, so I don't think we should do it here either. Overall, it probably would be good to show the default values for all properties, but let's do that separately.

PR #5614, opened 2026-05-27
docs: add Lumo enabling example to Lumo page (CP: 25.0)

vursen:
## Summary

Backport of the **Lumo** portion of #5612 to v25.0.

The `@StyleSheet` example for enabling Lumo previously existed only on the Themes overview page (`styling/themes/index.adoc`), not on the individual Lumo page — where users most often land from search.

This adds:
- An **Enabling Lumo** section with a `@StyleSheet(Lumo.STYLESHEET)` example, including `import com.vaadin.flow.theme.lumo.Lumo;`
- The same import to the existing Compact Preset example, for consistency

### Notes
- **Aura is not backported.** The Aura page is a stub (front matter only) in v25.0, so an enabling section there would float on an otherwise empty page.
- The "themes should be loaded before other styles" load-order note is **omitted**, matching the latest version.

### DIFF
--- a/articles/styling/themes/lumo/index.adoc
+++ b/articles/styling/themes/lumo/index.adoc
@@ -21,6 +21,21 @@ The built-in dark and compact variants are predefined sets of style property val
 The Lumo theme also includes a comprehensive set of CSS utility classes that can be applied to HTML elements and layout components through Java, to change their styling without writing any CSS of your own.
 
 
+== Enabling Lumo
+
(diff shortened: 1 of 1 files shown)


jouni:
25.0 is no longer supported. I don't think we need to update documentation for it either. The docs site doesn't include that version in the version selector.

web-padawan (approved):

vursen:
Closing as 25.0 is no longer supported.

PR #5621, opened 2026-05-28
docs: add Clipboard API article

mshabarov:
Related-to https://github.com/vaadin/flow/issues/17703

### DIFF
--- a/articles/flow/advanced/clipboard-api.adoc
+++ b/articles/flow/advanced/clipboard-api.adoc
@@ -1,0 +1,159 @@
+---
+title: Clipboard API
+page-title: How to use the Clipboard API in Vaadin
+description: Using the Clipboard API to copy text and HTML to the user's clipboard from the server.
+meta-description: Learn how to copy text and HTML to the user's clipboard in Vaadin applications using the server-side Clipboard API.
+order: 113
+---
+
+
+= Clipboard API
+:toc:
+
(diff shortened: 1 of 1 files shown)


articles/flow/advanced/clipboard-api.adoc
```diff
@@ -0,0 +1,159 @@
...
+The Clipboard API does not expose a permission state the same way other browser APIs do, so there is no separate availability signal -- handle errors at the point of the write.
+
+
+== User Gesture Requirement
+
+The Clipboard API requires every write to happen inside a fresh user gesture, which is why [methodname]`Clipboard.onClick(component)` is the only entry point. A few consequences:
+
+- Writes cannot be initiated from a background thread, a scheduled task, or a server push. The gesture must originate in the browser.
+- Each click is one gesture. Chaining multiple [methodname]`write*` calls on the same binding attaches multiple actions to the same click, but each one needs to succeed against the browser's gesture budget.
+- If the application performs a long server round-trip before the write call, the browser may consider the gesture stale and reject the write. Keep the action attached to the click rather than scheduling it through [methodname]`ui.access()`.
```
platosha: The last bit is a bit confusing: with the API structure given in this article, `ui.access()` isn't applicable at all, because `Clipboard.onClick()` does not allow arbitrary click handlers. I'd suggest to remove this, as there's hopefully enough context for understanding this limitation. Alternatively, could be rewritten with this in mind:

```suggestion
- If the application performs a long server round-trip before the write call, the browser may consider the gesture stale and reject the write. Therefore, the [methodname]`Clipboard.onClick(component)` entrypoint does not allow click handlers other than [methodname]`write*` calls.
```

platosha (changes requested):

articles/flow/advanced/clipboard-api.adoc
mshabarov: Agreed. Remove the last sentence .

platosha (approved):

PR #5627, opened 2026-06-01
docs: document Fullscreen API feature

tltv:
RelatedTo: vaadin/flow#24326

### DIFF
--- a/articles/flow/advanced/fullscreen-api.adoc
+++ b/articles/flow/advanced/fullscreen-api.adoc
@@ -1,0 +1,97 @@
+---
+title: Fullscreen API
+page-title: How to use the Fullscreen API in Vaadin Flow
+description: Using the Fullscreen API to enter, exit, and observe fullscreen mode from the server side.
+meta-description: Use the server-side Fullscreen API in Vaadin Flow to enter and exit browser fullscreen mode for the entire page or a single component from Java code.
+order: 125
+---
+
+
+= [since:com.vaadin:vaadin@V25.2]#Fullscreen API#
+:toc:
+
+The [classname]`Fullscreen` API gives server-side Java code access to the https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API[browser Fullscreen API]. It can fullscreen the entire page or a single component, exit fullscreen on demand, and expose the current fullscreen state as a reactive signal.
+
+[NOTE]
+The browser only allows a fullscreen request during a user gesture -- a click, a key press, or another transient activation. The API enforces this by attaching requests to a trigger component, so the request runs inside the DOM event that fires it. Calls from a constructor, a background thread, or a server push are rejected by the browser.
+
+
+== Entering Fullscreen
+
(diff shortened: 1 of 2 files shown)


articles/flow/advanced/fullscreen-api.adoc
mcollovati, line 1: Documentation on developer.mozilla.org also uses `Fullscreen` term, so I guess we can add `[fF]ullscreen` in .github/styles/config/vocabularies/Docs/accept.txt to get rid of these Vale warnings.

articles/flow/advanced/fullscreen-api.adoc
```diff
@@ -0,0 +1,97 @@
...
+---
+
+
+= [since:com.vaadin:vaadin@V25.2]#Fullscreen API#
+:toc:
+
+The [classname]`Fullscreen` API gives server-side Java code access to the https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API[browser Fullscreen API]. It can fullscreen the entire page or a single component, exit fullscreen on demand, and expose the current fullscreen state as a reactive signal.
+
+[NOTE]
+The browser only allows a fullscreen request during a user gesture -- a click, a key press, or another transient activation. The API enforces this by attaching requests to a trigger component, so the request runs inside the DOM event that fires it. Calls from a constructor, a background thread, or a server push are rejected by the browser.
```
mcollovati: What's `another transient activation`? I think we should either provide an example or remove this part

mcollovati: > Calls from a constructor, ...

Does this refer to directly calling JavaScript Fullscreen API, e.g. with `executeJS`? If so, I would make it more explicit, otherwise it could seem that calling the Java API is not rejected.

articles/flow/advanced/fullscreen-api.adoc
```diff
@@ -0,0 +1,97 @@
...
+
+
+== Entering Fullscreen
+
+Bind a fullscreen request to a component the user can click. [methodname]`Fullscreen.onClick(Component)` returns a [classname]`FullscreenBinding` whose [methodname]`enter()` method registers the request with that component's click. The request then runs inside the browser's click handler and inherits the user gesture:
+
+[source,java]
+----
+Button fullscreenButton = new Button("Fullscreen");
```
mcollovati, line 25: not strictly related to the docs, but what happens if the button also has a server side click handler? Is it just compatible with the client side trigger?

mcollovati (changes requested):

articles/flow/advanced/fullscreen-api.adoc
tltv: Yes it's compatible, no `preventDefault` or `stopPropagation` in play stopping it from working.

articles/flow/advanced/fullscreen-api.adoc
tltv: Added

articles/flow/advanced/fullscreen-api.adoc
tltv: Changed to "touch tap".

Updated to more explicit.

mcollovati (approved):

PR #5635, opened 2026-06-05
shared-signals.adoc: Replace ValueSignal with SharedValueSignal

DarioViva42:
This documentation describes the usage of Shared Signals. So it is really confusing and probably a mistake, that in some of the examples Local Signals are used.

### DIFF
--- a/articles/flow/ui-state/shared-signals.adoc
+++ b/articles/flow/ui-state/shared-signals.adoc
@@ -272,17 +272,17 @@ Use `@VaadinSessionScope` to create one signal instance per user session, i.e. s
 @Component
 @VaadinSessionScope
 public class CurrentUserSignal {
-    private final ValueSignal<String> usernameSignal;
+    private final SharedValueSignal<String> usernameSignal;
 
     public CurrentUserSignal(AuthenticationContext authContext) {
-        this.usernameSignal = new ValueSignal<>(
+        this.usernameSignal = new SharedValueSignal<>(
             authContext.getAuthenticatedUser(UserDetails.class)
                 .map(UserDetails::getUsername)
                 .orElse("anonymous")
         );
     }
 
-    public ValueSignal<String> getUsernameSignal() {
+    public SharedValueSignal<String> getUsernameSignal() {
         return usernameSignal;
     }
 }
@@ -301,13 +301,13 @@ Use `@Component` (singleton) or static fields to create signals shared by all us
 ----
 @Component
 public class SystemStatusSignal {
-    private final ValueSignal<String> statusSignal;
+    private final SharedValueSignal<String> statusSignal;
 
     public SystemStatusSignal() {
-        this.statusSignal = new ValueSignal<>("ONLINE");
+        this.statusSignal = new SharedValueSignal<>("ONLINE");
     }
 
-    public ValueSignal<String> getStatusSignal() {
+    public SharedValueSignal<String> getStatusSignal() {
         return statusSignal;
     }
 }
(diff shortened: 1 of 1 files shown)


DarioViva42:
Also what I also find pretty Strange that there is a Shared Signal `SharedNumberSignal` but not a Local Signal `NumberSignal`. There are also similar discrepancies in this regard. It's especially strange because in an earlier edition of Signals the class `NumberSignal` existed. Is there a migration guide for people that used the Signals as they were in the early days?

Also we get the following error:

> An asynchronous signal can only share transaction with computed signals and other asynchronous signals that belong to the same tree.

Beforehand this code worked.

We had two Signals that we created with the `IN_MEMORY_SHARED` SignalFactory. and then we updated the two signals with `Signal.runInTransaction`. This worked but now it does no longer.

Where should such questions be asked. On some documentation pages you can comment. Like https://vaadin.com/docs/latest/components/avatar. But on the documentation https://vaadin.com/docs/latest/flow/ui-state/shared-signals this does not work, which is kind of strange.

Legioth:
Thank you for pointing out the inconsistency in the documentation! At the same time, the correct fix isn't as easy as just replacing `ValueSignal` with `SharedValueSignal`. That's because `SharedValueSignal` is intended for sharing state between multiple users whereas two of the three examples that you suggest changing are sharing state between different parts seen by the same user. `SharedValueSignal` can be used in that case but it's overkill and will lead to some other trade-offs. We should instead revise the documentation to more clearly point out the conceptual difference between single-user sharing and multi-user sharing.

> Also what I also find pretty Strange that there is a Shared Signal SharedNumberSignal but not a Local Signal NumberSignal.

There's no need for a dedicated local signal type for numbers. You can just as well use `ValueSignal` instead. In the shared case, there's a need for an atomic `incrementBy` operation whereas the local case works just as well with `signal.update(x -> x + 1)`. Additionally, `SharedNumberSignal` is backed by JSON serialization and thus using a 64-bit floating point number which makes it convenient to have helper methods for accessing the value as an `int`. This is not necessary when using `ValueSignal<Integer>`.

> Also we get the following error:
>> An asynchronous signal can only share transaction with computed signals and other asynchronous signals that belong to the same tree.

This was one of the main reasons we decided to introduce the split between "local" and "shared" signals. The shared signals are designed to eventually support synchronizing state across a cluster. One of the trade-offs in that case is that different signals might be "owned" by different nodes in that cluster which means that it's not possible to have an atomic transaction that covers both of them. You can force both signals to be "owned" by the same node by creating them as child signals of the same root signal:

```java
SharedNodeSignal root  = new SharedNodeSignal();
ValueSignal<String> first = root.insertChildWithValue("first", ListPosition.last()).signal().asValue(String.class);
ValueSignal<String> second = root.insertChildWithValue("second", ListPosition.last()).signal().asValue(String.class);
```

It's now possible to use both `first` and `second` in the same transaction.

> On some documentation pages you can comment.

That's a configuration issue on our side. Thanks for pointing out.

In general, the kinds of questions you present here are still more appropriate for our discussion forum at https://vaadin.com/forum/.

Legioth:
I created https://github.com/vaadin/docs/issues/5636 for revising the whole structure of the documentation related to shared signal.

DarioViva42:
@Legioth Small correction in your code snippet:

```java
SharedNodeSignal root  = new SharedNodeSignal();
SharedValueSignal<String> first = root.insertChildWithValue("first", ListPosition.last()).signal().asValue(String.class);
SharedValueSignal<String> second = root.insertChildWithValue("second", ListPosition.last()).signal().asValue(String.class);
```

This might be confusing for others reading this in the future. But asValue returns `SharedValueSignal` and not `ValueSignal`

DarioViva42:
@Legioth 

I have also added an new forum entry. The code that used to work still does not, but I get another exception now.

https://vaadin.com/forum/t/sharednumbersignal/179566

PR #5713, opened 2026-06-23
docs: fix invalid since-badge placement and clarify guidance

Artur-:
Fixes `since` badges that were authored with invalid AsciiDoc and never
rendered. A `since` badge is an inline role that must be immediately followed by
a formatted text span (e.g. `#...#`); a span-less badge renders as literal
`[since:...]` brackets or silently produces nothing.

- Correct 7 broken usages across 7 articles: a badge trailing a heading
  (`== Title [since:...]`) and badges alone on their own line, rewritten to the
  span form (`== [since:...]#Title#`, or wrapping a phrase in body text).
- Update the CLAUDE.md "Version References" guidance to document the correct
  form so future docs use it.

Note: section-heading badges only become visible together with the dspublisher
fix in vaadin/docs-app#537 -- the build attached the version to the wrong
element (a preceding `[classname]` macro or the heading anchor) and had no CSS
rule for badges in `h2`-`h6` headings. This PR makes the markup correct; #537
makes it render.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -184,11 +184,12 @@ The repository uses Vale for automated style checking. Configuration is in `.val
 
 ### Version References
 
-Avoid explicit Vaadin version numbers in text. Instead, use version badges:
+Avoid explicit Vaadin version numbers in text. Instead, use a `since` badge.
 
-```asciidoc
-[since:com.vaadin:vaadin@V25.1]
-```
+The badge is an inline AsciiDoc role: `[since:<coordinates>@<version>]` **must** be immediately
+followed (no space) by a formatted text span that holds the text the badge applies to. The span is
+usually `#...#`, but any inline formatting works (e.g. `*...*`). Without a following span the badge
+does not render — it either shows up as literal `[since:...]` brackets or is silently dropped.
 
 This keeps documentation version-agnostic and maintainable.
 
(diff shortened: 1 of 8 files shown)


CLAUDE.md
```diff
@@ -184,10 +184,28 @@ The repository uses Vale for automated style checking. Configuration is in `.val
...
+
+The badge is an inline AsciiDoc role. It **must be immediately followed (no space) by a formatted
+text span** that holds the text the badge applies to — usually `#...#`, but any inline formatting
+works (e.g. `*...*`). Without a following span the badge shows up as literal `[since:...]` brackets
+or is dropped entirely. Place it on a heading or wrapping a phrase in body text.
+
+```asciidoc
+// Correct — on a page title or section heading:
+= [since:com.vaadin:vaadin@V25.1]#Slider#
+== [since:com.vaadin:vaadin@V25.2]#Hierarchical Data#
```
jouni, line 197: The recommended way to annotate a section heading is to place the role/class on the heading instead of a span inside it. For example:

```asciidoc
[role="since:com.vaadin:vaadin@V25.2"]
== Focus Selected Item [badge-flow]#Flow#
```

The only exception is the main/top-level heading, which doesn't allow it, so there you need to wrap it in a span.

I won't stop us from unifying everything one way or the other. But it’s not great to keep using both styles for 2nd and 3rd-level sections.

Artur-: #5787

PR #5828, opened 2026-07-21
docs: document UploadValidator for rejecting uploads

### LINKED ISSUE #24923: UploadHandler: built-in handlers ignore event.isRejected() when subclassed to reject uploads
### Description

The `UploadHandler` API exposes `UploadEvent.reject()` / `reject(String)` / `isRejected()` for refusing an upload (e.g. after inspecting file metadata or a header). When a plain `UploadHandler` lambda calls `event.reject(...)` and returns, the framework correctly routes the file into the rejected set (`TransferUtil.handleUpload` checks `isRejected()` and produces a 422/207 response).

**In their default usage the built-in handlers are not affected**, because they never expose the `UploadEvent` to user code before completing — there is no supported way to reject an upload during their lifecycle, so the flag is always `false`.

The gap only appears when a built-in handler is **subclassed and the `UploadEvent` is captured and rejected elsewhere** (for example, stashing the event and calling `reject()` from a progress/start listener). In that case the built-in handlers still invoke their success callback unconditionally after the transfer, ignoring `isRejected()`:

- `InMemoryUploadHandler.handleUploadRequest` → calls `successCallback.complete(...)`
- `AbstractFileUploadHandler.handleUploadRequest` → calls `successCallback.complete(metadata, file)`

Neither has an `isRejected()` check, so the success/complete callback fires for a rejected upload. For the file-based handlers the written file is additionally left on disk, which contradicts the `reject()` Javadoc that states the file "will be cleaned up if already processed."

The underlying reason this workaround is needed at all is that there is currently **no supported way to reject from within the built-in handlers' lifecycle**: progress/start listeners receive only a `TransferContext`, which exposes neither the `UploadEvent` nor a reject mechanism.

### Expected behavior

- Built-in handlers should not invoke their success callback when the upload has been rejected.
- File-based handlers should clean up the partially/fully written file on rejection, consistent with the `reject()` Javadoc.
- There should be a supported way to reject an upload from within the built-in handlers (e.g. from a start/progress listener), removing the need to subclass and capture the `UploadEvent`.

### Actual behavior

When a built-in handler is subclassed and the captured `UploadEvent` is rejected:

- The success/complete callback still runs in `InMemoryUploadHandler` and `AbstractFileUploadHandler`.
- Rejected file uploads leave an orphan file on disk.

Rejection is not reachable through the supported listener API (`TransferContext` has no reject support).

### Proposed changes

1. Add an `isRejected()` guard before the success callback in `InMemoryUploadHandler` and `AbstractFileUploadHandler`; for file handlers, delete the created file on rejection.
2. Introduce a reject mechanism on `TransferContext` (e.g. `reject(String reason)`) so uploads can be refused from within the handler lifecycle without subclassing. This should also be honored by `TransferUtil.transfer` so byte consumption can be aborted rather than run to completion. This likely generalizes to `DownloadHandler` as well.

Item 1 is a small consistency/robustness fix; item 2 is an API addition and may warrant a separate design discussion.

### Affected files

- `flow-server/src/main/java/com/vaadin/flow/server/streams/InMemoryUploadHandler.java`
- `flow-server/src/main/java/com/vaadin/flow/server/streams/AbstractFileUploadHandler.java`
- `flow-server/src/main/java/com/vaadin/flow/server/streams/TransferContext.java`
- `flow-server/src/main/java/com/vaadin/flow/server/communication/TransferUtil.java`

mcollovati:
Documents the new `UploadValidator` API from flow#24925, which lets the
built-in upload handlers refuse an upload while it's being received.

Previously only a fully custom `UploadHandler` could call
`UploadEvent.reject(...)`; the pre-made handlers (`InMemoryUploadHandler`,
`FileUploadHandler`, `TemporaryFileUploadHandler`) had no supported way to
reject an upload mid-transfer.

## Related

- Flow PR: vaadin/flow#24925
- Fixes: vaadin/flow#24923

Please review the wording yourself before marking this PR ready.

### DIFF
--- a/articles/components/upload/file-handling.adoc
+++ b/articles/components/upload/file-handling.adoc
@@ -198,7 +198,7 @@ Upload upload = new Upload(temporaryFileHandler);
 [role="since:com.vaadin:vaadin@V25.3"]
 === Validating Uploads with the Built-in Handlers
 
-The built-in handlers can refuse an upload while it's being received, without writing a custom [classname]`UploadHandler`. Add a validator to a handler with the fluent [methodname]`withValidator()` method, or one of the phase-specific shortcuts, and call [methodname]`UploadEvent.reject()` from it to abort the upload. A rejected upload is never delivered to the success callback, and any partially written file is deleted.
+The built-in handlers can refuse an upload while it's being received, without writing a custom [classname]`UploadHandler`. Add a validator to a handler with the fluent [methodname]`withValidator()` method, or one of the phase-specific shortcuts, and call [methodname]`UploadEvent.reject()` from it to cancel the upload. A rejected upload is never delivered to the success callback, and any partially written file is deleted.
 
 A validator can act in three phases, each running synchronously on the request thread as the upload is received:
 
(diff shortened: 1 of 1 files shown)


articles/components/upload/file-handling.adoc
```diff
@@ -192,6 +194,57 @@ TemporaryFileUploadHandler temporaryFileHandler = UploadHandler.toTempFile(succe
...
 
+[[validating-uploads]]
+[role="since:com.vaadin:vaadin@V25.3"]
+=== Validating Uploads with the Built-in Handlers
+
+The built-in handlers can refuse an upload while it's being received, without writing a custom [classname]`UploadHandler`. Add a validator to a handler with the fluent [methodname]`withValidator()` method, or one of the phase-specific shortcuts, and call [methodname]`UploadEvent.reject()` from it to cancel the upload. A rejected upload is never delivered to the success callback, and any partially written file is deleted.
+
+A validator can act in three phases, each running synchronously on the request thread as the upload is received:
+
+- *Metadata* -- before any data is read. Use it to reject by file name, content type, or the client-declared size.
```
knoobie:
```suggestion
- *Metadata* -- before any data is read. Use it to reject by file name, client-declared content type, or the client-declared size.
```

articles/components/upload/file-handling.adoc
```diff
@@ -192,6 +194,57 @@ TemporaryFileUploadHandler temporaryFileHandler = UploadHandler.toTempFile(succe
...
+
+The built-in handlers can refuse an upload while it's being received, without writing a custom [classname]`UploadHandler`. Add a validator to a handler with the fluent [methodname]`withValidator()` method, or one of the phase-specific shortcuts, and call [methodname]`UploadEvent.reject()` from it to cancel the upload. A rejected upload is never delivered to the success callback, and any partially written file is deleted.
+
+A validator can act in three phases, each running synchronously on the request thread as the upload is received:
+
+- *Metadata* -- before any data is read. Use it to reject by file name, client-declared content type, or the client-declared size.
+- *Header* -- over the first bytes of the content. Use it for file-signature ("magic byte") or MIME-sniffing checks that need only the leading bytes.
+- *Complete* -- after the whole upload has been received. Use it for whole-content checks, such as antivirus scanning.
+
+Each phase has a fluent shortcut that takes a lambda:
```
caalador, line 209: Should we have a short sample part here about `UploadValidator` also in addition to the shorthand lambdas?

mcollovati: Totally makes sense. It can be a validator class that implements the same logic as in the lambdas.

caalador (approved):

PR #5847, opened 2026-07-30
docs: add documentation for the experimental Switch component

web-padawan:
Added `Switch` component documentation and updated feature flag to link to the new page.

Note: screenshot for the components page is currently missing, I will create one later.

### DIFF
--- a/articles/components/switch/index.adoc
+++ b/articles/components/switch/index.adoc
@@ -124,7 +124,7 @@ Switch supports a single constraint: it can be marked as required, in which case
 
 include::{articles}/components/_input-field-common-features.adoc[tags=required]
 
-A required Switch isn't shown as invalid until the user has toggled it, or the application triggers validation, so no error appears before the user has acted. Toggling the switch back on clears the invalid state immediately.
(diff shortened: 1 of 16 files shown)


jouni:
The distinction between a checkbox and a switch should perhaps be made clearer, when to use one over the other.

The validation example is problematic for this reason exactly, that a checkbox would be the more appropriate component in that case, where you need to submit the value before it takes effect. Switch is commonly thought to be independent of other components, that its state change takes effect immediately without submitting/saving a form.

web-padawan:
> The distinction between a checkbox and a switch should perhaps be made clearer, when to use one over the other

Updated and added dedicated "When to Use" section clarifying this distinction. Also updated validation example.

jouni:
Thanks! I might change the validation example slightly still, so that the switch is checked by default, and when you uncheck it, the validation error is shown and says "Required by your workplace security policy" (i.e., change it from a helper text to error-message). And add one sentence to the description before the example that says "Turn off the switch to see the validation message."

web-padawan:
> I might change the validation example slightly still, so that the switch is checked by default, and when you uncheck it, the validation error is shown and says "Required by your workplace security policy" (i.e., change it from a helper text to error-message). And add one sentence to the description before the example that says "Turn off the switch to see the validation message."

Thanks, done.

jouni (approved):

Issue #78, opened 2020-12-23
Add more detailed documentation/demo about default behavior after login


DiegoCardoso:
When user submits the login, the submit button is disabled to prevent another attempt before the login is processed.

Add:
- **Documentation** describing the behavior
- **Demo** with details with what the developer should do in order to re-enable login

jouni:
There is now textual documentation about this: https://github.com/vaadin/docs/blob/master/articles/ds/components/login/index.asciidoc#login-event

Should probably still implement a proper code example.

jouni:
I think the first example should already tell about this behavior, and the example should show how to re-enable the button once the login attempt has been handled (either valid or invalid credentials).

web-padawan:
Fixed by #5860.

PR #5860, opened 2026-08-10
docs: add example of handling login and re-enabling login form

HerbertsVaadin:
Solves #78

Adds the missing piece - include it as a part of the first example.

### DIFF
--- a/articles/components/login/index.adoc
+++ b/articles/components/login/index.adoc
@@ -127,6 +127,38 @@ endif::[]
 [NOTE]
 Login is incompatible with password managers if placed inside another component's shadow root. footnote:[When added to a web component that uses shadow root, password managers are unable to find the input fields and therefore won't work.]
 
+=== Handle the Login Event
+
+When the user submits the login form, the Login component fires a login event with the entered username and password. Listen for this event to authenticate the credentials and handle the result.
+
+The log-in button is disabled after submission to prevent multiple submissions while the login attempt is being processed. If authentication fails, make the form available for another attempt.
+
+[.example,themes="lumo,aura"]
+--
+
+ifdef::lit[]
+[source,html]
+----
+include::{root}/frontend/demo/component/login/login-event.ts[render,tags=snippet,indent=0,group=Lit]
+----
+endif::[]
+
+ifdef::flow[]
+[source,java]
+----
+include::{root}/src/main/java/com/vaadin/demo/component/login/LoginEventExample.java[render,tags=snippet,indent=0,group=Flow]
+----
+endif::[]
+
+ifdef::react[]
+[source,tsx]
+----
+include::{root}/frontend/demo/component/login/react/login-event.tsx[render,tags=snippet,indent=0,group=React]
+----
+endif::[]
+--
+
+
 == Modal Overlay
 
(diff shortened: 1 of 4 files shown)


frontend/demo/component/login/login-basic.ts
```diff
@@ -19,11 +19,19 @@ export class Example extends LitElement {
     return root;
   }
 
+  private async onLogin(event: Event) {
+    const loginForm = event.currentTarget as LoginForm;
+
+    // Re-enable login button
+    await loginForm.updateComplete;
+    loginForm.disabled = false;
```
web-padawan: Let's change this example to store actual disabled state:

```ts
@state()
private disabled = false;
```

And sync it from the component in `@disabled-changed="${this._onDisabledChanged}"`:

```ts
_onDisabledChanged(event: LoginFormDisabledChangedEvent) {
  this.disabled = event.target.value;
}
```

Then you can update `onLogin` as follows:

```ts
private onLogin(event: LoginFormLoginEvent) {
  this.disabled = false;
}
```

frontend/demo/component/login/login-basic.ts
```diff
@@ -1,7 +1,7 @@
 import 'Frontend/demo/init'; // hidden-source-line
-import '@vaadin/login/vaadin-login-form.js';
```
web-padawan, line 2: Please keep the import, it is necessary for the component to work. Type-only import must be separate.

frontend/demo/component/login/react/login-basic.tsx
```diff
@@ -9,7 +9,13 @@ function Example() {
       {/* tag::snippet[] */}
       {/* no-autofocus is used to prevent the example from stealing focus when browsing the
       documentation */}
-      <LoginForm no-autofocus />
+      <LoginForm
+        no-autofocus
+        onLogin={(event) => {
```
web-padawan: Please use the same approach as for the TS example with `onDisabledChanged` event and `const disabled = useSignal<boolean>(false);` to store the state from the event, passed to the component.

web-padawan (changes requested):

frontend/demo/component/login/react/login-validation.tsx
```diff
@@ -1,10 +1,42 @@
...
+import type {
+  LoginFormDisabledChangedEvent,
+  LoginFormErrorChangedEvent,
+} from '@vaadin/react-components/LoginForm';
 import { LoginOverlay } from '@vaadin/react-components/LoginOverlay.js';
 
 // tag::snippet[]
 function Example() {
-  return <LoginOverlay opened error no-autofocus />;
+  const [disabled, setDisabled] = useState(false);
```
web-padawan: Let's use signals instead, as in other React examples e.g.

```ts
const disabled = useSignal<boolean>(false);
```

web-padawan (changes requested):
I just realised that this resetting disabled should not be not needed for `LoginOverlay` as it was fixed in https://github.com/vaadin/web-components/pull/12274.

So, let's remove this from the validation example and create a new example using `LoginForm` where the problem exists.

web-padawan (approved):

PR #5875, opened 2026-08-14
Rephrase `@EnableVaadin` clarification to be Spring Boot centric

Legioth:
The note about the `@EnableVaadin` annotation being optional originated from a time when Spring Boot wasn't the defacto way of using Spring. That phrasing is confusing nowadays when Spring Boot is the most common way.


HerbertsVaadin (approved):
Pretty simple and clear change. Approved.

PR #5882, opened 2026-08-18
docs: document combo-box partialMatchMode

vursen:
Adds a `Partial Match Behavior` section to the Combo Box reference, with Lit, Flow, and React examples, documenting the `partialMatchMode` property (`none` | `first-match` | `only-match`)

Fixes https://github.com/vaadin/web-components/issues/1280


articles/components/combo-box/index.adoc
```diff
@@ -343,6 +343,60 @@ endif::[]
 --
 
 
+[role="since:com.vaadin:vaadin@V25.3"]
+== Partial Match Selection
```
vursen: Partial Filter Matches?

rolfsmeds: Partial Match Behavior?

articles/components/combo-box/index.adoc
```diff
@@ -343,6 +343,60 @@ endif::[]
...
+
+Pressing kbd:[Enter] while filtering commits an item only when the typed filter matches its full label. If the filter matches the label only partially, the item has to be focused in the overlay first, for example with the kbd:[Down] arrow key.
```
rolfsmeds: Pressing Tab or blurring the field in any other way also commits the focused value.

rolfsmeds (changes requested):

articles/components/combo-box/index.adoc
vursen: I added "blurring the field" to the sentence. Is that what you meant?

articles/components/combo-box/index.adoc
vursen: Done.

articles/components/combo-box/index.adoc
rolfsmeds: That's probably clear enough.

rolfsmeds (approved):

articles/components/combo-box/index.adoc
```diff
@@ -343,6 +343,60 @@ endif::[]
 --
 
 
+[role="since:com.vaadin:vaadin@V25.3"]
+== Partial Match Behavior
+
+Pressing kbd:[Enter] or blurring the field (for example, by pressing kbd:[Tab]) while filtering commits an item only when the typed filter matches its full label. If the filter matches the label only partially, the item has to be focused in the overlay first, for example with the kbd:[Down] arrow key.
```
vursen:
```suggestion
Pressing kbd:[Enter] or blurring the field (for example, by using kbd:[Tab]) while filtering commits an item only when the typed filter matches its full label. If the filter matches the label only partially, the item has to be focused in the overlay first, for example with the kbd:[Down] arrow key.
```

PR #5891, opened 2026-08-24
docs: add article for threading and ui updates

ugur-vaadin:
Adds documentation for threading and UI updates.

Related feature PR: https://github.com/vaadin/flow-components/pull/9914


articles/flow/ai-support/background-execution.adoc
```diff
@@ -0,0 +1,85 @@
+---
```
tomivirkki: Consider rescoping this page as the threading reference rather than a feature page. The `setBackgroundExecution()` API itself is just one boolean flag, and only matters when streaming is off, but ~80% of this page documents the threading/delivery contract of a turn: what runs on which thread, push vs. polling, detached-UI semantics, concurrency.

Suggestion: title this something like "Threads & UI Updates", cover all three modes (streaming default / synchronous foreground / synchronous background), and demote the flag to one section.

That would also let the threading blurbs added to tool-calling, controllers, and conversation-history in this PR collapse into a single link instead of four independently-worded restatements that can drift.

articles/flow/ai-support/llm-providers.adoc
```diff
@@ -67,6 +69,36 @@ LangChain4JLLMProvider provider = new LangChain4JLLMProvider(chatModel);
...
+A synchronous provider produces the response on the thread that asks for it. For a prompt sent from the browser, that is the request thread: the request does not return until the model has produced the complete response, including any tool calls along the way. The interface freezes for the whole wait, and since the request holds the session lock, other views in the same session wait too.
+
+Background execution moves the exchange to a background thread instead. Enable it on either built-in provider:
+
+[source,java]
+----
+provider.setBackgroundExecution(true);
+----
+
+The user's message and an empty assistant message then appear immediately, the UI stays responsive, and the response is filled in when the model finishes. The setting is off by default, because tool implementations written for synchronous execution may rely on running in the request thread. It's read for each prompt, so it can be changed at any time; the next prompt uses the new mode. It has no effect with a streaming model, whose response already arrives on the LLM client's own threads.
```
tomivirkki: "...because tool implementations written for synchronous execution may rely on running in the request thread"

Maybe drop this...I don't think this is the primary reason why it's off by default.

articles/flow/ai-support/llm-providers.adoc
```diff
@@ -67,6 +69,36 @@ LangChain4JLLMProvider provider = new LangChain4JLLMProvider(chatModel);
...
+
+Wrap component access in `ui.access()`, or capture what a tool needs in [methodname]`AIController.onRequest()` while the request thread is still current -- see <<tool-calling#,Tool Calling & Programmatic Prompts>> and <<controllers#,Controllers>>. The built-in controllers already handle this. [methodname]`AIController.onResponse()` is the exception: the orchestrator calls it through `ui.access()`, so it can update components directly.
+
+.One Prompt at a Time
+[NOTE]
+The orchestrator processes one prompt at a time. In the default synchronous mode, a message submitted while a turn is running waits for the session lock and is processed when the turn ends. With background execution the lock is free, so the same message is rejected and dropped with a server-side warning -- and the Message Input has already cleared its text. Disable the input while a turn is running, for example from [methodname]`AIController.onRequest()` and [methodname]`AIController.onResponse()`.
```
tomivirkki: Should we drop the recommendation to disable the input? We haven't recommended that for streaming mode before either. It may be bad practice if the focus gets lost.

tomivirkki: "waits for the session lock". Can we confirm that's the mechanism? Just "…is processed when the turn ends" is safer?

ugur-vaadin: Dropped the recommendation.

articles/flow/ai-support/llm-providers.adoc
ugur-vaadin: Yep, dropped it.

tomivirkki (approved):

PR #5894, opened 2026-08-27
docs: document CRUD customization API and editor behavior

peholmst:
Addresses [vaadin/agentic-dx-improvement#103](https://github.com/vaadin/agentic-dx-improvement/issues/103), which reported that the Flow CRUD documentation presents the component as a mostly turn-key "grid + generated editor" and leaves several decision-critical facts discoverable only by inspecting the jar.

## Changes

`articles/components/crud/index.adoc`:

- **Editor Actions** promoted from `====` to `===`, and expanded with two new subsections:
  - **Save Button State** — the exact enablement rules (dirty, not validity), the role of `setDirty()`, and how to keep Save enabled at all times.
  - **Editor Button Access** — a table of `getSaveButton()` / `getCancelButton()` / `getDeleteButton()` / `getNewButton()` / `setNewButton()`, with an example of hiding Delete for datasets that archive rather than remove records, plus a note that `setText()` on these buttons is overwritten by localization.
- **Custom Editor** — documents `CrudEditor<E>` as a public extension point, with a table of when CRUD calls each method and a live non-`BinderCrudEditor` example that validates itself and renders a top-of-form `role="alert"` error summary.
- **Controlling the Editor Programmatically** — collects `edit()`/`EditMode`, `setOpened()`, `setEditorPosition()`, and `setEditOnClick()` into one table.
- **No Manual Reordering** note under Sorting & Filtering, cross-linked to Grid Replacement.
- **Lazy Backend Loading** — how `CrudFilter`'s constraint and sort-order maps translate into `WHERE`/`ORDER BY`, the `DataProvider<E, CrudFilter>` requirement enforced by `CrudGrid`, and the existing `PersonDataProvider` as the reference implementation.
- **Styling** — surfaces the `no-border` variant and `CrudVariant` on the main page and links to the styling page, which the main page didn't link at all.

New examples: `PersonCrudEditor.java`, `CrudCustomEditor.java`, `CrudEditorButtons.java`, and `crud-imports.ts` (preimport for the Flow-only demos).

## Two corrections to the ticket

Everything was checked against the sources for `vaadin-crud-flow` 25.2.1 / 25.3.0-alpha8, `@vaadin/crud`, and Flow's element binding. Two of the ticket's claims don't hold:

- **Gap 2 claimed there's no public API for an always-enabled Save.** There is: `crud.getSaveButton().setEnabled(true)`. `Crud`'s internal `SaveButton` overrides `onEnabledStateChanged` to permanently override the web component's `__isSaveBtnDisabled`, and re-applies it on attach. Flow also overrides the client-side `__validate` to always pass, so the server-side `CrudEditor.validate()` is the only gate on a save. Documented as supported, which makes the ADR-0020 pattern from the ticket achievable.
- **Gap 3 proposed `getDeleteButton().setVisible(false)`.** That isn't reliable: `setVisible(false)` sets Flow's `hidden` attribute, but `__deleteButtonPropsChanged` runs `toggleAttribute('hidden', isNew)` each time the editor opens, so Delete reappears when editing an existing item. Vaadin's own integration tests only use `setEnabled` on these buttons. The docs recommend `getStyle().set("display", "none")` instead, with a short note on why.

`CrudVariant.AURA_NO_BORDER` is deprecated in favour of `NO_BORDER` as of 25.2.1, so the docs point at `NO_BORDER`.

## Verification

- `mvn compile` passes.
- `spotless:apply` produces no changes to the new Java files.
- Prettier and ESLint are clean on `crud-imports.ts`.
- Vale couldn't run locally (no Ruby/asciidoctor on this machine), so the page was linted as a `.txt` copy against `.vale-pr.ini` and diffed against the pre-change baseline. The only new hits are artifacts of that workaround — `[methodname]`/`[classname]` roles, `_italics_`, and source-block attributes, all in ignored classes and scopes for the real run. The two genuine ones ("deletable" spelling, a Chicago title-case heading) are fixed.
- The rendered demos have **not** been visually verified — the docs server wasn't started. Worth a look in the preview build.

https://claude.ai/code/session_01CvBbsmxH7A3F8WyMz2hfZZ

### DIFF
--- a/articles/components/crud/index.adoc
+++ b/articles/components/crud/index.adoc
@@ -290,13 +290,13 @@ CRUD's Buttons are ordinary [classname]`Button` instances that you can access fr
 
 Set the Button labels through <<#localization,Localization>>, not [methodname]`setText()`: CRUD writes the localized labels onto its default Buttons, overwriting anything set that way.
 
-Some datasets shouldn't allow deletion at all -- for example, records that are archived or deactivated instead of removed, so that history is preserved. Hide the Delete Button with CSS in that case.
+Some datasets shouldn't allow deletion at all -- for example, records that are archived or deactivated instead of removed, so that history is preserved. There's no API for removing the Delete Button, so the only way to do this at present is to hide it with CSS.
 
 [.example,themes="lumo,aura"]
 --
 [source,typescript]
 ----
-include::{root}/frontend/demo/component/crud/crud-imports.ts[preimport,hidden]
+include::{root}/frontend/demo/component/crud/crud-editor-content.ts[preimport,hidden]
 ----
 
 [source,java]
@@ -304,6 +304,10 @@ include::{root}/frontend/demo/component/crud/crud-imports.ts[preimport,hidden]
 include::{root}/src/main/java/com/vaadin/demo/component/crud/CrudEditorButtons.java[render,tags=snippet,indent=0]
 ----
 --
+
+.Hiding Delete Is a Workaround
+[NOTE]
+Setting `display: none` isn't advocated by the CRUD API; it's the only thing that works today. Don't reach for [methodname]`setVisible(false)`: CRUD manages the Delete Button's `hidden` attribute itself, and it clears the attribute each time the editor opens for an existing item, which makes the Button reappear.
 
 
 [#custom-editor]
@@ -342,7 +346,7 @@ The editor in the example below validates the item itself and renders all proble
 --
 [source,typescript]
 ----
-include::{root}/frontend/demo/component/crud/crud-imports.ts[preimport,hidden]
+include::{root}/frontend/demo/component/crud/crud-editor-content.ts[preimport,hidden]
 ----
 
 [source,java]
@@ -623,13 +627,6 @@ include::{root}/frontend/demo/component/crud/react/crud-localization.tsx[render,
 ----
 endif::[]
 --
-
-
-== Styling
-
-CRUD has a `no-border` style variant, which removes the border around the component. In Flow, apply it with [methodname]`addThemeVariants(CrudVariant.NO_BORDER)`. In Lit and React, set `theme="no-border"` on the component.
-
-Beyond that variant, CRUD exposes CSS custom properties for its background, borders, toolbar, and editor. See <<styling#,Styling>> for the full list.
 
 
 == Related Components
--- a/articles/components/crud/index.adoc
+++ b/articles/components/crud/index.adoc
@@ -249,6 +249,7 @@ endif::[]
 The editor contains three Buttons: _Delete_, _Cancel_, and _Save_. The Delete shows a confirmation dialog asking the user to verify whether they wish to delete the item. Whereas the Cancel closes the editor unless there are unsaved changes. If so, a confirmation dialog is shown and the user can either discard the changes or go back to editing. The Save button, when clicked, saves the changes and closes the editor. This is disabled until a change is made.
 
 
+[#save-button-state]
 ==== Save Button State [badge-flow]#Flow#
 
 Save is enabled as soon as the editor becomes dirty -- that is, as soon as the user changes the value of a field in it. Validity doesn't factor into this: Save is enabled for invalid input, too. Clicking it runs the editor's [methodname]`validate()` method, and when validation fails, the editor stays open and nothing is saved.
@@ -265,6 +266,7 @@ crud.getSaveButton().setEnabled(true);
 See <<#custom-editor,Custom Editor>> for an editor that pairs this with an error summary.
 
 
+[#editor-button-access]
 ==== Editor Button Access [badge-flow]#Flow#
 
 CRUD's Buttons are ordinary [classname]`Button` instances that you can access from the server to change their state, appearance, or behavior:
@@ -350,7 +352,7 @@ include::{root}/src/main/java/com/vaadin/demo/component/crud/CrudCustomEditor.ja
 
 [source,java]
 ----
-include::{root}/src/main/java/com/vaadin/demo/component/crud/PersonCrudEditor.java[render,indent=0]
+include::{root}/src/main/java/com/vaadin/demo/component/crud/PersonCrudEditor.java[indent=0]
 ----
 --
 
--- a/articles/components/crud/index.adoc
+++ b/articles/components/crud/index.adoc
@@ -244,10 +244,140 @@ endif::[]
 --
 
 
-==== Editor Actions
+=== Editor Actions
 
 The editor contains three Buttons: _Delete_, _Cancel_, and _Save_. The Delete shows a confirmation dialog asking the user to verify whether they wish to delete the item. Whereas the Cancel closes the editor unless there are unsaved changes. If so, a confirmation dialog is shown and the user can either discard the changes or go back to editing. The Save button, when clicked, saves the changes and closes the editor. This is disabled until a change is made.
 
+
+==== Save Button State [badge-flow]#Flow#
+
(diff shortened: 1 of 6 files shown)


articles/components/crud/index.adoc
```diff
@@ -469,6 +623,13 @@ endif::[]
 --
 
 
+== Styling
```
peholmst: This is duplicating info that already exists in `styling.adoc`. Remove.

articles/components/crud/index.adoc
```diff
@@ -244,11 +244,141 @@ endif::[]
...
+
+Set the Button labels through <<#localization,Localization>>, not [methodname]`setText()`: CRUD writes the localized labels onto its default Buttons, overwriting anything set that way.
+
+Some datasets shouldn't allow deletion at all -- for example, records that are archived or deactivated instead of removed, so that history is preserved. Hide the Delete Button with CSS in that case.
+
+[.example,themes="lumo,aura"]
+--
+[source,typescript]
+----
+include::{root}/frontend/demo/component/crud/crud-imports.ts[preimport,hidden]
```
peholmst: There is no need for a TypeScript example in a Flow-only section. Remove.

frontend/demo/component/crud/crud-imports.ts
```diff
@@ -0,0 +1,7 @@
+import 'Frontend/demo/init';
+import '@vaadin/crud';
+import '@vaadin/email-field';
+import '@vaadin/form-layout';
+import '@vaadin/text-field';
+
+// This file only has the required imports for the Java-only CRUD examples
```
peholmst: This file should not be needed at all.

articles/components/crud/index.adoc
```diff
@@ -244,11 +244,141 @@ endif::[]
...
 
 
+==== Save Button State [badge-flow]#Flow#
+
+Save is enabled as soon as the editor becomes dirty -- that is, as soon as the user changes the value of a field in it. Validity doesn't factor into this: Save is enabled for invalid input, too. Clicking it runs the editor's [methodname]`validate()` method, and when validation fails, the editor stays open and nothing is saved.
+
+An editor that doesn't propagate its field changes -- one built from a composite component that wraps its fields, for example -- can leave Save permanently disabled. Call [methodname]`setDirty(true)` on the CRUD to enable it explicitly in such a case.
+
+Some applications need Save to be enabled at all times. A common accessibility pattern is to let the user submit at any point and then show a summary of what needs fixing. Set the enabled state of the Save Button directly to get this: CRUD stops managing that Button's state from then on, leaving it enabled regardless of whether the editor is dirty. The editor's [methodname]`validate()` method still decides whether a save goes through.
```
peholmst, line 259: Is this behavior documented in the CRUD Javadoc? If not, this might actually be a bug in the `CRUD` component.

articles/components/crud/index.adoc
```diff
@@ -244,11 +244,143 @@ endif::[]
...
+|[methodname]`getDeleteButton()`
+|The editor's Delete Button. CRUD hides it automatically while a new item is being created.
+
+|[methodname]`getNewButton()` and [methodname]`setNewButton()`
+|The toolbar's _New item_ Button. See <<#toolbar,Toolbar>> for an example of replacing it.
+|===
+
+Set the Button labels through <<#localization,Localization>>, not [methodname]`setText()`: CRUD writes the localized labels onto its default Buttons, overwriting anything set that way.
+
+Some datasets shouldn't allow deletion at all -- for example, records that are archived or deactivated instead of removed, so that history is preserved. Hide the Delete Button with CSS in that case.
```
peholmst: This looks like a hack or a workaround. Unless it is advocated by the CRUD Javadoc, it should be stated as such, pending a better Java API for achieving the same thing.

articles/components/crud/index.adoc
peholmst: Removed — the section is gone. The Flow-side detail it added (`addThemeVariants(CrudVariant.NO_BORDER)`) now lives only in `styling.adoc`, next to the variant table.

frontend/demo/component/crud/crud-imports.ts
peholmst: Deleted. Both examples now use the existing `crud-editor-content.ts` as their counterpart — it already imports `@vaadin/crud`, `@vaadin/email-field`, `@vaadin/form-layout`, and `@vaadin/text-field`, which is a superset of what the file listed. Same approach as the Combo Box "Focus Selected Item" section, which reuses `combo-box-basic.ts`.

articles/components/crud/index.adoc
peholmst: I removed the bespoke file, but I had to keep a hidden TypeScript counterpart here — it is load-bearing rather than illustrative, and it is not shown to readers.

I tested removing it on a local dspublisher run. Both demos break: the elements render at height 0 with no grid, and the console fills with `Cannot read properties of undefined (reading 'initLazy')`, `$0._resetColumnOrder is not a function`, and `window.Vaadin.setLitRenderer is not a function`. Putting a counterpart back makes both render correctly again.

This matches the contributing guide (`contributing/docs/dspublisher/editing.adoc`, "Client-Side Dependencies of Java Examples"): "Styles for components that are used in an example are only loaded for TypeScript examples. You therefore always need a TypeScript counterpart for any Java example." Every other Flow-only example does the same — Spreadsheet, Map, Grid inline editing, Combo Box.

So the include now points at `crud-editor-content.ts` with `preimport,hidden`. Happy to drop it entirely if `DOCS_IMPORT_EXAMPLE_RESOURCES=true` is set for the real builds — the guide mentions that as the alternative, and I could not tell from the repo whether it is.

articles/components/crud/index.adoc
peholmst: Yes, this part is documented. `Crud.getSaveButton()`:

> NOTE: State of the button set with `HasEnabled#setEnabled(boolean)` will remain even if dirty state of the crud changes

and `Crud.setDirty(boolean)`:

> A dirty Crud has its editor Save button enabled. […] NOTE: editor Save button will not be automatically enabled in case its enabled state was changed with `Crud#getSaveButton()`

The two `@see` each other, so the override is a deliberate, documented escape hatch. `SaveButton.onEnabledStateChanged` implements it by overriding the web component's `__isSaveBtnDisabled`, and `Crud.onAttach` re-applies it, so it survives detach/attach.

One caveat on the paragraph above this one, which is **not** in the Javadoc: "Validity doesn't factor into this: Save is enabled for invalid input, too." That comes from the implementation — `vaadin-crud-mixin.js` has `__isSaveBtnDisabled(isDirty) { return !isDirty; }`, and Flow's `Crud.onAttach` sets `this.__validate = function () { return true; }` so client-side validation never gates the save. Server-side `CrudEditor.validate()` is the only gate. Happy to drop that sentence if you would rather not pin down undocumented behavior.

articles/components/crud/index.adoc
peholmst: Agreed — it is a workaround, and the Javadoc does not advocate it. Reworded to say so, and added a note:

> There's no API for removing the Delete Button, so the only way to do this at present is to hide it with CSS.

> **Hiding Delete Is a Workaround** — Setting `display: none` isn't advocated by the CRUD API; it's the only thing that works today. Don't reach for `setVisible(false)`: CRUD manages the Delete Button's `hidden` attribute itself, and it clears the attribute each time the editor opens for an existing item, which makes the Button reappear.

That last part is arguably a component bug worth its own ticket: Flow's `setVisible(false)` sets the `hidden` attribute (`SimpleElementBindingStrategy.setElementInvisible`), while `__deleteButtonPropsChanged` in `vaadin-crud-mixin.js` runs `deleteButton.toggleAttribute('hidden', isNew)` on every `__isNew` change, clearing it. So the one obvious API for this silently does not work. A `setDeleteButtonVisible(boolean)`, or having CRUD respect the server-side visibility, would remove the need for the CSS. Want me to file that against `flow-components`?

PR #5904, opened 2026-08-31
chore: update to Vaadin 25.3.0-alpha9

tomivirkki:
## Summary
- Bumps the platform to 25.3.0-alpha9 (web components 25.3.0-alpha14, Hilla 25.3.0-alpha10); package.json and package-lock.json are regenerated by the Vaadin build, with the lockfile recreated from scratch using npm 11.17 so the component packages hoist to the top level again.
- Updates the Flow connector imports to the new per-component paths introduced in flow-components (the upload connector move has no deprecated entrypoint, so the old path no longer resolves).
- Keeps the jar-resources connectors out of the tsc program: their generated global `Vaadin.Flow` declarations conflict across component jars when type-checked together (to be fixed in flow-components). Also pins a TypeScript 6 copy for the eslint toolchain, since typescript-eslint cannot load the TypeScript 7 that the platform now pins (no JS compiler API).


frontend/demo/component/badge/badge-interactive.ts
```diff
@@ -1,5 +1,5 @@
 import 'Frontend/demo/init'; // hidden-source-line
-import '@vaadin/flow-frontend/comboBoxConnector'; // hidden-source-line
+import '@vaadin/flow-frontend/vaadin-combo-box/comboBoxConnector'; // hidden-source-line
```
web-padawan: suggestion: let's remove the connector import from here - it is covered by `Frontend/demo/init` above.

```suggestion
```

web-padawan (approved):

PR #5906, opened 2026-08-31
chore: work around Flow i18n startup regression

tomivirkki:
## Summary
- The packaged production app fails to start with the current Flow prerelease because the build emits a `vaadin-i18n` folder into the jar, which trips a Flow regression at servlet init.
- This app has no translation bundles, so dropping the empty folder at packaging time is safe and restores previous behavior.
- Temporary workaround — remove once a Flow release containing the fix ([flow#25282](https://github.com/vaadin/flow/pull/25282)) is picked up.
- This fixes the latest `main` revision not getting deployed to https://vaadin.com/docs/next/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -269,6 +269,35 @@
                             </execution>
                         </executions>
                     </plugin>
+                    <!--
+                      Workaround for a Flow regression: the production build
+                      emits a vaadin-i18n folder into the packaged jar, whose
+                      presence makes the application fail to start. This app has
+                      no translation bundles, so the folder is safe to drop.
+                      Remove this once the regression is fixed in Flow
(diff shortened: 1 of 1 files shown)


mcollovati:
We are going to release a new Flow alpha with the fix. I'll close this one once the docs repo gets updated and the problem is resolved.

mcollovati:
@tomivirkki #5937 should fix the issue. Shall we merge and verify?

tomivirkki:
@mcollovati Let's see if it helps, merged the PR.

tomivirkki:
Seems to work again, https://vaadin.com/docs/next/ is now beta1

PR #5910, opened 2026-08-31
docs: add overlay animation properties to Styling tables

web-padawan:
Added documentation for base styles / Aura overlay animation custom CSS properties.

### DIFF
--- a/articles/components/_styling-section-theming-props.adoc
+++ b/articles/components/_styling-section-theming-props.adoc
@@ -254,6 +254,48 @@ endif::include-item-overlay-props[]
 // end::overlay[]
 
 
+// tag::overlay-animation[]
+[role="since:com.vaadin:vaadin@V25.3"]
+=== Animation
+
+The overlay has a built-in opening and closing animation, which is disabled until an animation
+duration is set. The `closed` properties define the state that the overlay animates from when it
+opens, and to when it closes. The Aura theme sets these properties for some components. Lumo has its
+own overlay animations, which these properties don't affect.
+
+While an animation runs, the component has the `opening` or `closing` state attribute, so separate
+values can be applied to each direction. The backdrop always fades, and only the opacity is animated
+when the user prefers reduced motion.
+
+[cols="2,1"]
+|===
+| Property | Supported by
+
+|`--vaadin-overlay-animation-delay`
+|Aura
+
+|`--vaadin-overlay-animation-duration`
+|Aura
+
+|`--vaadin-overlay-animation-timing-function`
+|Aura
+
+|`--vaadin-overlay-opacity-closed`
+|Aura
+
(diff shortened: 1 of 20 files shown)


articles/components/date-time-picker/styling.adoc
```diff
@@ -47,6 +47,12 @@ include::../_styling-section-theming-props.adoc[tag=style-properties]
 |Aura
 |===
 
+include::../_styling-section-theming-props.adoc[tag=overlay-animation]
+
+[NOTE]
+These properties are set on the sub-fields, `vaadin-date-time-picker > vaadin-date-picker` and
+`vaadin-date-time-picker > vaadin-time-picker`, not on the `vaadin-date-time-picker` root element.
```
sissbruecker, line 54: Is there a technical reason for that? Can we not inherit the props to descendants?

articles/components/_styling-section-theming-props.adoc
```diff
@@ -254,6 +254,48 @@ endif::include-item-overlay-props[]
...
+duration is set. The `closed` properties define the state that the overlay animates from when it
+opens, and to when it closes. The Aura theme sets these properties for some components. Lumo has its
+own overlay animations, which these properties don't affect.
+
+While an animation runs, the component has the `opening` or `closing` state attribute, so separate
+values can be applied to each direction. The backdrop always fades, and only the opacity is animated
+when the user prefers reduced motion.
+
+[cols="2,1"]
+|===
```
sissbruecker, line 271: Is it just me, or is it harder to figure out which values you are supposed to use here compared to the other prop sections?

articles/components/date-time-picker/styling.adoc
web-padawan: @jouni should we update animation styles to make this clear? The comment added as part of https://github.com/vaadin/web-components/pull/12120 doesn't explain why it's done this way:

> The properties are registered as non-inheriting, so components with more than one shadow root boundary between the host and the overlay need to forward them explicitly.

articles/components/_styling-section-theming-props.adoc
web-padawan: I agree. Will add a column with some value examples.

articles/components/date-time-picker/styling.adoc
jouni: They don't inherit so that, for example, dialog overlay animations aren't inherited by any overlays within them. If the properties would inherit, we would need to unset them on each host that contains an overlay.

jouni: But we can fix this for date-time-picker if we want to. Just explicitly inherit them in the contained date-picker and time-picker elements from the date-time-picker host.

sissbruecker (approved):

PR #5912, opened 2026-08-31
docs: unify date formatting in TS examples to use date-fns

web-padawan:
Unifies date formatting across the TypeScript examples on `date-fns`, which 34 of the 36 date-formatting examples already used.

- **badge-highlight**: `format(date, 'MMM d, yyyy')` instead of a module-level `Intl.DateTimeFormat` const. These were the last two `Intl.DateTimeFormat` usages in `frontend/demo/`. Output is identical ("Apr 29, 2026"), and the formatting is now visible inside the documented snippet rather than hidden above it.
- **Explicit date patterns everywhere**: replaced `date-fns` tokens `P` / `PP`with `MM/dd/yyyy` / `MMM d, yyyy`. Under the default en-US locale these are exact expansions of the tokens, so nothing renders differently — but a reader of a documentation example shouldn't have to know `date-fns`-specific shorthand. Affects `grid-column-alignment` and `grid-rich-content-sorting`, which already used `P` before this branch.
- **Import style**: `date-fns` is imported from the package root everywhere, instead of mixing root and per-function deep imports (`date-fns/format`, `date-fns/getDaysInMonth`, …).
- **Naming**: dropped the `dateFnsFormat` / `dateFnsParse` aliases in the DatePicker and DateTimePicker custom-format examples. Neither file has a name collision, and every other example imports `format` / `parse` directly.
- **message-list-attachments**: reuses the `isoMinutes` const, as the five sibling MessageList examples do.

`Intl.NumberFormat` usages are deliberately left alone — `date-fns` has no number or currency formatting.

`date-picker-custom-functions` also keeps its literal `'yyyy-MM-dd'` rather than switching to `formatISO`: that example's whole subject is configuring the picker for ISO 8601 via `formatDate` / `parseDate`, so the literal pattern deliberately mirrors what the snippet configures.

## How to test

No rendered output changes anywhere, so this is a read-through plus a check that the demos still build:

1. `npm run dspublisher:start`
2. Open the Badge, Grid, Grid Pro, DatePicker, DateTimePicker and MessageList pages and confirm the date columns/labels render as before — Badge "Due date" shows `Apr 29, 2026`, Grid and Grid Pro date columns show `04/29/2026`.
3. Check the Lit/React snippets shown on those pages read clearly on their own.

### DIFF
--- a/frontend/demo/component/badge/react/badge-highlight.tsx
+++ b/frontend/demo/component/badge/react/badge-highlight.tsx
@@ -8,7 +8,7 @@ import { getReports, type Report, ReportStatus } from 'Frontend/demo/domain/Data
 
 // tag::snippet[]
 function renderDueDate({ item: report }: { item: Report }) {
-  return <span>{format(new Date(report.due), 'PP')}</span>;
+  return <span>{format(new Date(report.due), 'MMM d, yyyy')}</span>;
 }
 
 function renderStatus({ item: report }: { item: Report }) {
--- a/frontend/demo/component/badge/react/badge-highlight.tsx
+++ b/frontend/demo/component/badge/react/badge-highlight.tsx
@@ -1,19 +1,14 @@
 import { reactExample } from 'Frontend/demo/react-example'; // hidden-source-line
 import React, { useEffect } from 'react';
 import { useSignals } from '@preact/signals-react/runtime'; // hidden-source-line
+import { format } from 'date-fns';
 import { useSignal } from '@vaadin/hilla-react-signals';
 import { Badge, Grid, GridColumn } from '@vaadin/react-components';
 import { getReports, type Report, ReportStatus } from 'Frontend/demo/domain/DataService';
 
-const dateFormatter = new Intl.DateTimeFormat('en-US', {
-  year: 'numeric',
-  month: 'short',
-  day: 'numeric',
-});
-
 // tag::snippet[]
 function renderDueDate({ item: report }: { item: Report }) {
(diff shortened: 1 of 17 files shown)


frontend/demo/component/badge/react/badge-highlight.tsx
```diff
@@ -1,19 +1,14 @@
...
-const dateFormatter = new Intl.DateTimeFormat('en-US', {
-  year: 'numeric',
-  month: 'short',
-  day: 'numeric',
-});
-
 // tag::snippet[]
 function renderDueDate({ item: report }: { item: Report }) {
-  return <span>{dateFormatter.format(new Date(report.due))}</span>;
+  return <span>{format(new Date(report.due), 'PP')}</span>;
```
sissbruecker: IMO a format string like `PP` is problematic for a documentation example as it's a date-fns specific pattern that you're likely not familiar with.

web-padawan: I was also not sure about this. Changed to use `MMM d, yyyy` and `MM/dd/yyyy` instead.

DiegoCardoso: nit: the PR description can be updated as it mentions the `P`/`PP` usage.

sissbruecker (approved):

frontend/demo/component/badge/react/badge-highlight.tsx
web-padawan: Updated to mention that `P` / `PP` have been replaced instead of "spelled out".

PR #5919, opened 2026-09-01
docs: mention tools with no parameters

ugur-vaadin:
Clarifies what to use as the schema for parameterless tools and why.

See the related [fix PR](https://github.com/vaadin/flow-components/pull/9971)

### DIFF
--- a/articles/flow/ai-support/controllers.adoc
+++ b/articles/flow/ai-support/controllers.adoc
@@ -132,7 +132,7 @@ Each tool is an implementation of [classname]`LLMProvider.ToolSpec`, which has f
 
 * [methodname]`getName()` -- the unique name the LLM uses to invoke the tool.
 * [methodname]`getDescription()` -- a human-readable description shown to the LLM.
-* [methodname]`getParametersSchema()` -- a JSON Schema string describing the tool's parameters, or `null` for tools that take no parameters.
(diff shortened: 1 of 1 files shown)


tomivirkki:
I think this can be closed without merging; the line it replaces was already correct.

Both providers substitute the constant for `null` or blank, so `return null;` and `return NO_PARAMETERS_SCHEMA;` produce identical requests. Nothing changes for a developer.

The underlying problem also looks Bedrock-specific: in the [forum thread](https://vaadin.com/forum/t/ai-grid-preview-feedback/179716) all five models were Bedrock-hosted and they disagreed with each other, and I haven't been able to reproduce the failure on another provider locally. And an empty `properties` object is the canonical way to declare a no-arg tool — our own `get_form_state` used it before https://github.com/vaadin/flow-components/pull/9971 — so I'd rather not warn people off it in the general `ToolSpec` reference.

Related suggestion for flow-components, while 25.3 is still in alpha: make `NO_PARAMETERS_SCHEMA` package-private rather than public API. It's behaviourally identical to `null`, so it buys users nothing, but as public API it freezes the `reason` property name — the example here has callers reading `arguments` and ignoring `"reason"` — and we lose the option of changing the placeholder later. Every built-in usage is a bare `return NO_PARAMETERS_SCHEMA;` and can become `return null;` with identical behaviour, leaving only the two providers needing it — both in `com.vaadin.flow.component.ai.provider`, so it could live package-private there.

While we're there: whenever a provider substitutes the schema itself, it could also hand `execute()` an empty object. A tool that declared no parameters would then receive no parameters, rather than a property it never asked for, and the placeholder — name and all — stays entirely inside the provider.

ugur-vaadin:
> I think this can be closed without merging; the line it replaces was already correct.

I agree with the assessment. Closing the PR. I will also open a PR that includes the mentioned changes.

PR #5769, opened 2026-06-30
feat: update observability documentation

caalador:
Add migration documentation
and update references, configuration
and customization documents.

PR #5922, opened 2026-09-01
docs: fix inaccuracies and fill gaps in Observability Kit V5 docs

heruan:
Follow-up to #5769, cross-checking the V5 documentation against the kit source at `vaadin/observability-kit` HEAD (`4f98221`) and the `vaadin/use-cases/observability` demo application.

## Accuracy fixes

- **Histogram buckets**: the Reference and Prometheus pages implied timers export `_bucket` suffixes by default. They export count/sum/max only; buckets are opt-in per meter, and a `histogram_quantile()` query over a bucket-less timer silently returns no data.
- **Prometheus renames**: `_created` is a reserved OpenMetrics suffix, so `vaadin.sessions.created` and `vaadin.ui.created` export as `vaadin_sessions_total` and `vaadin_ui_total`. Noted in the naming conventions, the migration rename step, and the Grafana examples.
- `vaadin.navigation` also carries the `error` tag.
- The root request span is named `vaadin.request.<type>` (e.g. `vaadin.request.rpc`), per the kit's observation contract test — not bare `vaadin.request`.
- Grammar: "Three ways exists", "depends on that its conventions".

## New content

- **Client connection meters** (observability-kit#385, merged after #5769): `vaadin.client.connection` and `vaadin.client.connection.downtime` with the `state` tag values, plus the `_unknown` error kind.
- **Percentiles and Histogram Buckets** section on the Integrations page, with the `management.metrics.distribution.*` properties and the series-count caveat.
- **Example Dashboard Queries** on the Grafana page: PromQL recipes derived from the use-case app's provisioned dashboard, plus a link to that dashboard and its compose stack.
- **Troubleshooting** section in Getting Started: dev-mode license check silently skipping instrumentation, lazily registered `vaadin.errors`, Prometheus renames, empty percentile queries.
- **Migration guide**: v4 histogram queries break structurally (no buckets), parallel long-task `_active` series, logs no longer exported by the kit.
- **Insights**: the 1000 ms UX budget is fixed; endpoint access on restricted setups (`management.endpoint.vaadin.access=unrestricted`); in-process reads via the `VaadinObservabilityEndpoint` bean; and a worked **"Fixing Insights with an AI Agent"** example piping the endpoint payload into a coding agent.
- Error propagation to the enclosing Spring HTTP observation (observability-kit#381) and a GraalVM native-image support note.

Deliberately not documented: `traces-session-id` (currently a no-op) and HTTP route templating (observability-kit#384, unmerged).

Vale is clean for the changed lines; remaining warnings in these files predate this branch.

### DIFF
--- a/articles/tools/observability/configuration.adoc
+++ b/articles/tools/observability/configuration.adoc
@@ -222,7 +222,7 @@ The feature toggles map directly to the built-in meters and spans:
 Off by default.
 
 |`navigation`
-|Server-side navigation timing, tagged by route and outcome.
+|Server-side navigation timing, tagged by route, outcome, and error.
 
 |`requests`
 |Server-side request and RPC timing.
--- a/articles/tools/observability/configuration.adoc
+++ b/articles/tools/observability/configuration.adoc
@@ -91,7 +91,7 @@ See <<reference#ui-state-size,UI State Size>>.
 
 |`vaadin.observability.client`
 |`true`
-|Browser-side timing collected from the client.
+|Browser-side timing, errors, and connection state collected from the client.
 
 |`vaadin.observability.resync`
 |`true`
@@ -236,7 +236,7 @@ Also gates the data query insights.
 Also decorates the session error handler, which is what makes the failures Flow routes there countable and attributable to a component.
 
 |`client`
-|Browser-observed timing -- bootstrap, navigation, Web Vitals, and client errors.
+|Browser-observed signals -- bootstrap, navigation, and Web Vitals timing, client errors, and connection-state transitions with their downtime.
 
 |`resync`
 |A counter of UIDL message resends and client-requested resynchronizations, tagged by type.
--- a/articles/tools/observability/integrations/grafana.adoc
+++ b/articles/tools/observability/integrations/grafana.adoc
@@ -128,3 +128,6 @@ Remember that Prometheus renders the dotted meter names in its own convention --
 
 The percentile queries read `_bucket` series, which the kit's timers don't publish by default -- enable them per meter first; see <<index#percentiles,Percentiles and Histogram Buckets>>.
 The UI state panel needs the opt-in `vaadin.observability.ui-state` feature, and a byte figure appears only when [propertyname]`vaadin.observability.ui-state-bytes-per-node` is set; see <<../reference#ui-state-size,UI State Size>>.
+
+A complete dashboard built from these queries ships with the https://github.com/vaadin/use-cases/tree/main/observability[Vaadin observability use-case application]: a provisioned Grafana dashboard ([filename]`grafana/dashboards/vaadin.json`) together with a Docker Compose stack that runs Prometheus and Grafana against the application.
+Import the JSON into your own Grafana as a starting point, or run the stack to see the panels over live data.
--- a/articles/tools/observability/integrations/grafana.adoc
+++ b/articles/tools/observability/integrations/grafana.adoc
@@ -104,8 +104,8 @@ Remember that Prometheus renders the dotted meter names in its own convention --
 |Active UIs (browser tabs)
 |`sum(vaadin_ui_active)`
 
-|Requests handled
-|`sum(vaadin_request_duration_seconds_count)`
+|Request rate
+|`sum(rate(vaadin_request_duration_seconds_count[$__rate_interval]))`
 
 |Server errors
 |`sum(rate(vaadin_errors_total[$__rate_interval]))`
--- a/articles/tools/observability/integrations/grafana.adoc
+++ b/articles/tools/observability/integrations/grafana.adoc
@@ -87,3 +87,44 @@ Select a metric from the [guilabel]*Metric* drop-down -- for example `vaadin_ses
 
 Open the Explore view and select [guilabel]*Loki* as the data source.
 Under [guilabel]*Labels*, configure a label for `service_name` with your application's service name, then click the refresh button in the top-right to search for logs.
+
+
+== Example Dashboard Queries
+
+The following PromQL expressions cover the most useful panels for a Vaadin dashboard.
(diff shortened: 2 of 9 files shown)


articles/tools/observability/reference.adoc
```diff
@@ -435,7 +452,10 @@ This separates a combo box loading matches for typed text from one loading the w
 |On `vaadin.client.navigation.duration`, what moved the browser: `back` for history navigation, or `programmatic` for a `pushState`/`replaceState` call.
 
 |`kind`
-|On `vaadin.client.errors`, the source of the browser error: `uncaught` or `promise`.
+|On `vaadin.client.errors`, the source of the browser error: `uncaught` or `promise`, with `_unknown` for a report that is neither.
+
+|`state`
+|On the client connection meters, the connection state: `connected`, `reconnecting` (the first request failed, the client is retrying), or `connection-lost` (the client exhausted its retries), with `_unknown` for a reported state outside that set.
```
caalador: The state tag row is wrong for the downtime timer.
`ClientMetricNames.downtimeState()` admits only OFFLINE_STATES, with the javadoc spelling it out: "a browser only spends time being unreachable, so STATE_CONNECTED on this meter  would be a contradiction and is bucketed with the rest." So on vaadin.client.connection.downtime, a reported `connected` becomes `_unknown.` The counter takes all three; the timer takes two.

articles/tools/observability/reference.adoc
caalador, line 453: trigger was left behind. The PR added _unknown to the kind and state rows, but MeterNames.TRIGGER_UNKNOWN = "_unknown" exists too, and reference.adoc:452 still lists only back and programmatic. Same fallback, same table, now inconsistent within itself.

articles/tools/observability/configuration.adoc
caalador: configuration.adoc wasn't updated alongside reference.adoc. The navigation row still reads "tagged by route and outcome" while the PR added error to the same meter at reference.adoc:162.

The errors row right below it does list all three of its tags, so the omission reads as a fact, not a summary.

articles/tools/observability/migrating-to-v5.adoc
```diff
@@ -324,6 +324,19 @@ Actuator registers JVM and process metrics out of the box.
...
+.Queries Break Structurally, Not Only by Name
+[IMPORTANT]
+====
+Three export-level changes affect existing queries and alerts beyond the renames above:
+
+* The version 4 agent exported OpenTelemetry histograms with buckets.
+Version 5 timers publish count, sum, and max only, so a `histogram_quantile()` query or a latency alert finds no `_bucket` series until you enable them per meter -- and it fails silently, returning no data rather than an error.
+See <<integrations#percentiles,Percentiles and Histogram Buckets>>.
+* Prometheus strips the reserved `_created` suffix, so `vaadin.sessions.created` and `vaadin.ui.created` appear as `vaadin_sessions_total` and `vaadin_ui_total`.
+* With tracing on, each observed timer also publishes a parallel long-task timer (`_active` series in Prometheus), which roughly doubles the series count for those meters.
```
caalador, line 336: `_active` is explained only here; reference.adoc:24 and prometheus.adoc:93 enumerate suffixes without it.

articles/tools/observability/integrations/grafana.adoc
```diff
@@ -87,3 +87,47 @@ Select a metric from the [guilabel]*Metric* drop-down -- for example `vaadin_ses
...
+|Panel |Query
+
+|Active sessions
+|`sum(vaadin_sessions_active)`
+
+|Active UIs (browser tabs)
+|`sum(vaadin_ui_active)`
+
+|Requests handled
+|`sum(vaadin_request_duration_seconds_count)`
```
caalador: "Requests handled" uses a raw cumulative counter while the neighbouring row uses rate()

caalador (commented):
`vaadin.db.fetch.rows` already publishes percentiles — DatabaseFetchMetrics.java:58 .publishPercentiles(0.95, 0.99). It's a DistributionSummary, so the new section's wording ("The kit's timers export their count, sum, and max only") is technically safe, but a reader comes away believing no percentile data exists anywhere without opting in.
This is the one meter that ships client-side quantiles by default, and it's documented nowhere. One sentence in the new Percentiles section would close it.

heruan:
Good catch — verified: `vaadin.db.fetch.rows` is indeed the one meter with `publishPercentiles(0.95, 0.99)` (DatabaseFetchMetrics.java:58). Addressed in 6d139cfb4: the Percentiles section now calls it out as the exception (including that pre-computed percentiles are per-instance and can't be re-aggregated like buckets), and the reference table row for `vaadin.db.fetch.rows` notes the alerting in Database Monitoring needs no extra configuration.

articles/tools/observability/reference.adoc
heruan: Fixed in f2ea14083. The state row now separates the two meters: the counter takes all three states, the downtime timer takes only the two unreachable ones, and both bucket anything else as `_unknown`.

articles/tools/observability/reference.adoc
heruan: Fixed in f2ea14083. The trigger row now lists `_unknown` as the fallback, same as kind and state.

articles/tools/observability/configuration.adoc
heruan: Fixed in f2ea14083. The navigation row now lists route, outcome, and error, matching the reference table.

articles/tools/observability/migrating-to-v5.adoc
heruan: Fixed in f2ea14083. The `_active_seconds` family is now also mentioned in the reference naming note and on the Prometheus page, where the suffixes are listed.

articles/tools/observability/integrations/grafana.adoc
heruan: Fixed in f2ea14083. Changed to a rate query and renamed the row to "Request rate", consistent with the neighboring rows.

caalador (approved):

PR #5927, opened 2026-09-01
docs: add Dev Loop CLI documentation for AI agents and hot reload

tltv:
vaadin-dev CLI is experimental feature targeting Vaadin 25.3.

RelatedTo: vaadin/flow#25376

### DIFF
--- a/articles/flow/configuration/live-reload/dev-loop-cli.adoc
+++ b/articles/flow/configuration/live-reload/dev-loop-cli.adoc
@@ -45,38 +45,25 @@ Two design properties make that verdict worth trusting:
 
 *Exactly one gate decides.* While the daemon owns your application, Flow's own save-triggered watchers are suspended. Nothing else pushes a stylesheet or refreshes a component behind your back, so one command decides when a change goes live and one command reports on it.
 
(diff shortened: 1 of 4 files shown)


articles/flow/configuration/live-reload/dev-loop-cli.adoc
```diff
@@ -0,0 +1,724 @@
...
+Two design properties make that verdict worth trusting:
+
+*Exactly one gate decides.* While the daemon owns your application, Flow's own save-triggered watchers are suspended. Nothing else pushes a stylesheet or refreshes a component behind your back, so one command decides when a change goes live and one command reports on it.
+
+*It never over-claims.* A successful class redefinition is not proof that a change is visible. If Flow had no component to refresh, if an entity mapping is involved, if a proxied Spring bean changed shape, or if a frontend import moved, `apply` says so and escalates to a restart instead of reporting `Stable`. Compile errors, and errors the application logged while the change went live, are reported rather than swallowed.
+
+Developers get exactly the same commands. Nothing on this page is agent-only.
+
+
+== How It Works
```
mshabarov: Starting from here it describes very low level details of the tool that are supposed to be used by agents only today. I'd not share all of these in our docs, so drop it or make it significantly shorter.

tltv: Reason is that page was prompted to be written for agents primarily but since related skills have also reference to details, we can leave some details out from this article.

tltv: Updated

mshabarov (approved):

PR #5945, opened 2026-09-04
docs: make vaadin.launch-browser findable by search

### LINKED ISSUE #5944: Add vaadin.launch.browser in the documentation
The property `vaadin.launch.browser` is not listed and documented here:

https://vaadin.com/docs/latest/flow/configuration/properties

jcgueriaud1:
Fixes #5944

`vaadin.launch-browser` is documented, under [Launch Browser in Development Mode](https://vaadin.com/docs/latest/flow/integrations/spring/configuration#launch-browser-in-development-mode) — but it was impossible to find, because the exact name appeared on that page **only inside an `application.properties` code block**. The only occurrences in the page text were the `-delay` variant, so searching for `launch-browser` matched `launch-browser-delay` and never surfaced the property itself.

## Changes

**`articles/flow/integrations/spring/configuration.adoc`** — name the property in the sentence that introduces the code block, which is what the page already does for `vaadin.allowed-packages`, `vaadin.blocked-packages`, and `vaadin.exclude-urls`:

> …by setting the `vaadin.launch-browser` property to `true`:

Two other properties on the same page had the same omission, so they're fixed the same way: `vaadin.devmode-caching` and `vaadin.url-mapping`.

**`articles/flow/configuration/properties.adoc`** — remove the `launch-browser-delay` entry. That page documents properties "set through the system properties, or with the servlet initialization parameters", and its list is scoped to those defined in `DeploymentConfiguration`, `Constants`, and `InitParameters`. Browser launching doesn't work by either means:

- `launch-browser` is not an init parameter at all — it's a field on the Spring `VaadinConfigurationProperties` bean, i.e. the Spring Boot `vaadin.launch-browser` property (default `false`).
- The only code that launches a browser is `DevModeBrowserLauncher` in `vaadin-spring`, a `SpringApplicationRunListener` that calls `DevModeHandlerManager.launchBrowserInDevelopmentMode(url)` from `ready()`, guarded by `properties.isLaunchBrowser()`. It's the sole caller in the codebase.
- `launch-browser-delay` *is* an `InitParameters` constant read via `ApplicationConfiguration`, but since only that Spring listener ever triggers a launch, it has no effect outside a Spring Boot application — and it's meaningless without `launch-browser`.

So the delay entry was both inapplicable on that page and the reason searches for `launch-browser` landed there instead of on the page that actually explains the feature.

https://claude.ai/code/session_01C9V2HYdj4htGjdh3qW7ZXC

### DIFF
--- a/articles/flow/configuration/properties.adoc
+++ b/articles/flow/configuration/properties.adoc
@@ -237,8 +237,13 @@ Sets the fully-qualified name for the internationalization provider class. To tr
 Default: `null` +
 Mode: Runtime
 
+`**launch-browser**`::
+Opens the application in the default browser when the application is started in development mode. A new tab isn't opened on every restart: after a tab has been opened, Vaadin waits for the period defined by `launch-browser-delay` before opening another one. To open a tab on the next start, clean the project (e.g., `mvn clean` or `gradle clean`), or delete the [filename]`tab.launch` file in the build directory. See <<{articles}/flow/integrations/spring/configuration#launch-browser-in-development-mode, Launch Browser in Development Mode>> for Spring-specific instructions. +
+Default: `false` +
+Mode: Development
+
 `**launch-browser-delay**`::
-Defines how long to wait before opening a new browser tab for the staring application. The time is counted always from the latest application start that would try to open a tab. +
+Defines how long to wait before opening a new browser tab for the starting application, when `launch-browser` is enabled. The time is counted always from the latest application start that would try to open a tab. Set to `0` to open a tab on every start. +
 Default: 30 min +
 Mode: Development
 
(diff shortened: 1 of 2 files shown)


articles/flow/configuration/properties.adoc
```diff
@@ -237,11 +237,6 @@ Sets the fully-qualified name for the internationalization provider class. To tr
 Default: `null` +
 Mode: Runtime
 
-`**launch-browser-delay**`::
```
mcollovati, line 240: I notice the points in the PR description about the removal of the property from here, but shouldn't this instead be documented in the Spring configuration page?

jcgueriaud1: It was already documented so you don't see it in the diff: https://github.com/vaadin/docs/blob/54d6f2b32cb031ffc9635ea897e43ced643d5e9c/articles/flow/integrations/spring/configuration.adoc#launch-browser-in-development-mode

mcollovati: Ah, yeah. My bad, sorry.

mcollovati (approved):

PR #5958, opened 2026-09-07
fix: update observability documentation

caalador:
Update the documentation of
observability-kit to match
changes that made statements faulty.

### DIFF
--- a/articles/tools/observability/configuration.adoc
+++ b/articles/tools/observability/configuration.adoc
@@ -22,7 +22,7 @@ When you need to turn features off or tune them, how you configure the kit depen
 |`vaadin.observability.*` properties in [filename]`application.properties` (or any Spring property source).
 
 |Plain Spring
-|Most of the same `vaadin.observability.*` keys, read with `@Value`.
+|Most of the same `vaadin.observability.*` keys, read from the Spring [classname]`Environment`.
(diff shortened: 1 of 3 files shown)


heruan (changes requested):
The edits themselves are accurate: three collectors, tripled capacity, `getRecentClientErrors()`, the grouping keys, the client counter descriptions and the client property all match the kit.

Since this PR was opened, two kit changes made more of `reference.adoc` stale. They are outside the diff, so listing them here rather than inline:

- Lines 498, 522 and 572 still name the span `vaadin.ui.access`. It was renamed to `vaadin.executor.task` in vaadin/observability-kit#397. The scope is now "tasks submitted to the Vaadin service executor" (signal effects, signal result notifications, application tasks), not every `UI.access`.
- Line 540 lists the request types as `uidl`, `heartbeat`, `push`, `static`, `other`. vaadin/observability-kit#399 added `bootstrap` and `stream`.

Optional: the replay example in `insights.adoc` around line 168 shows the production format only. Since vaadin/observability-kit#400, development mode adds view state steps such as `Set the 'Reason' Select to 'Defective'` and `Click the 'Process return' Button`. A sentence on that difference would help, but it can also be a follow-up.

caalador:
Updated documentation to match current state of the kit.

heruan (approved):

PR #5961, opened 2026-09-07
feat: document observability-kit client insights

caalador:
Add documentation for client
error insights.


heruan (approved):
Checked the page against the kit source on main, including the changes merged this week. Counter and tag names, the endpoint path, the properties and defaults, the JSON shape and wording, the stack frame rules, the browser buffer limits and the insight cap all match. Anchors and cross references resolve.

Heads up: #6019 adds a bullet at the same spot in the "See also" list of getting-started.adoc, so whichever merges second will need a small conflict fix.

PR #5969, opened 2026-09-08
docs: align AI docs with the current AI components module

tomivirkki:
## Summary
- The form filler returns rejections in the `fill_form` tool result, so the model corrects them within the same turn, not on the next one as the Grid and Form pages claimed.
- The built-in controllers now let the system prompt take precedence over conflicting workflow steps, which the Grid, Chart, and Form workflow tips did not mention.
- The automatic field marker and programmatic attachments lacked since badges, and the quick start pinned a Spring AI BOM release candidate instead of the version the module is built against.


ugur-vaadin (approved):
LGTM

PR #5972, opened 2026-09-08
docs: SwingBridge config updates and rendering details page

eftunv:
## Description

Documents the configuration surface Swing Bridge 1.3.0 added, and closes three open documentation issues.

**Configuration.** `configuration.adoc` documented nine system properties; 1.3.0 added ten more and removed none. The page is regrouped into Core, Display and Rendering, Input and Logging, and a JVM tuning section records the garbage-collector settings that matter when packing many sessions into one host — G1 with an aggressive free ratio for the tightest footprint, Shenandoah where memory should return promptly as sessions end, and why ZGC is a poor fit at this scale.

**New page.** `rendering.adoc` covers the three properties that decide what the Swing application is told about the screen and how each window's back buffer is allocated: `swingbridge.screenSize`, `swingbridge.hidpi` with its two cost ceilings, and `swingbridge.backBufferType`.

**Also in here.** The annotations JAR now comes from Maven Central rather than the `vaadin-addons` repository. The interop *What's Shipped Today* table has been re-audited against the plugin — it was not listing the `extract-shared-types` goal. "Customer" is reworded where it meant the reader's own code rather than a domain entity. The installation page gained `--add-opens=java.base/java.lang`, which was missing from both its runtime block and its flag table while the starter and playground both pass it. And `faq.adoc` gained the version include and title heading its siblings have, plus the nav order collision between Configuration and Interoperability is resolved.

No new public API is documented: the members 1.3.0 added are not intended for application developers.

Fixes: vaadin/vaadin-swing-bridge#269
Fixes: vaadin/vaadin-swing-bridge#259
Fixes: vaadin/vaadin-swing-bridge#213


mvysny (approved):
LGTM

PR #6000, opened 2026-09-10
docs: one section for theme name bindings, and no dark in the examples

### LINKED ISSUE #5995: Hackathon 25.3: say that Page.setColorScheme is the dark mode API, and that theme="dark" is a Lumo selector rather than a second one
### Description

There is one way to put a Vaadin 25 application into dark mode and it works in both themes. Nothing says so, and the pattern every Lumo era example teaches looks like a second way while only being half of one.

Measured on 25.3.0-beta1, with no theme declared on the app shell so neither theme leaks into the other, reading the **root element** rather than the body:

| | `ui.getElement().getThemeList().add("dark")` | `Page.setColorScheme(DARK)` |
| --- | --- | --- |
| Lumo | `<body>` background `rgb(35, 51, 72)`, `<html>` stays light | `<html>` background `rgb(35, 51, 72)` |
| Aura | nothing changes, on either element | `<html>` background `oklch(0.2 0.01 260)` |

`Page.setColorScheme` sets both `theme="dark"` and `style.color-scheme: dark` on `<html>`, so Lumo's attribute selector and Aura's `color-scheme` are both satisfied by the one call. `getThemeList()` sets only the attribute and only on `<body>`, because that is the element a `UI` is.

So `getThemeList().add("dark")` is not a dark mode API. It is the generic theme name API, which every component has and which is right for `small`, `contrast` or an application's own name, and it happens to match Lumo's own dark selector. Under Aura it matches nothing.

### Why it matters

An application that follows the older examples writes the line that only works in one theme, and nothing tells it. Nothing is logged, nothing throws, and the page is simply light. We shipped a browserless test asserting the attribute; it passed while the feature was dead, and it took clicking a button in a real browser to notice.

There is a second, quieter consequence of the same difference that is worth a sentence. Even under Lumo, where the attribute path does work, it paints `<body>` and leaves `<html>` light, so anything the body does not cover shows the light page behind it. `setColorScheme` does not have that problem because it paints the root.

### Expected

A page, or a section on the theming page, that says:

- `Page.setColorScheme(...)` and `@ColorScheme(...)` are how an application goes dark, and they work under Lumo and Aura;
- the `theme="dark"` attribute is Lumo's own selector, not an API, and setting it through `getThemeList()` reaches `<body>` and does nothing at all under Aura;
- what to assert in a test, which is `getColorScheme()` rather than the attribute.

The Aura theming page is the place a reader arrives with this question, and the Lumo one is where the older habit comes from, so both want the pointer.

### Reproduce

[`38-dark-mode.zip`](https://github.com/vaadin/hackathon-25-3/raw/898c37c36c2d2d66d4f62757498d9604ef1663c1/specs/issues/projects/38-dark-mode.zip), then:

```
mvn spring-boot:run
```

`http://localhost:8138/`: a button per theme and a button per mechanism, with no theme on the app shell. Load a theme, press a dark button, and read `getComputedStyle(document.documentElement).backgroundColor`. Reading `document.body` instead is what produced our own wrong table, so the page is worth measuring on the root.

### History

This was half of https://github.com/vaadin/flow/issues/25575. The other half, a `ThemeList` that cached the attribute and silently discarded another instance's writes, was a real Flow bug and is fixed there.

That issue claimed dark mode has two mechanisms and an application has to set both. A maintainer's triage asked for one cell of the table to be re-measured on the root element, and the re-measurement went their way: there is one mechanism, it works in both themes, and the table above is the corrected one. What is left is the documentation.

Found on 25.3.0-beta1.

manolo:
Fixes #5995

* One section for theme name bindings instead of two.
* No `dark` in the examples: theme names are variants like `small`.
* Theming page: `@ColorScheme` and `Page.setColorScheme()` are the color scheme API, the `dark` theme name is a Lumo selector on `<body>`, and tests assert `getColorScheme()`.

### DIFF
--- a/articles/styling/themes/index.adoc
+++ b/articles/styling/themes/index.adoc
@@ -143,6 +143,18 @@ The color scheme can be changed dynamically at runtime using the [methodname]`Pa
 UI.getCurrentOrThrow().getPage().setColorScheme(ColorScheme.Value.DARK);
 ----
 
+Both the annotation and the method set the `theme` attribute and the `color-scheme` CSS property on the `<html>` element. That's why a single call covers both themes: Lumo selects its dark colors with `[theme~="dark"]`, while Aura selects its own through `color-scheme`.
+
+.The `dark` Theme Name Is Not a Second Mechanism
+[CAUTION]
+====
+Adding a `dark` theme name, as in `ui.getElement().getThemeList().add("dark")`, looks like an alternative but isn't one. [methodname]`getThemeList()` is the generic theme name API that every component has, and it's the right tool for variants such as `small` or `contrast`. It only happens to match Lumo's own `[theme~="dark"]` selector.
+
+It writes the attribute on the `<body>` element, since that's the element a [classname]`UI` is. Under Lumo that leaves the `<html>` element light behind whatever the body doesn't cover, and under Aura it has no effect at all. Neither case logs anything or throws, so the page stays light with no indication of why.
+
+In tests, assert on [methodname]`Page.getColorScheme()` rather than on the `theme` attribute. The attribute is only a plain `dark` for [propertyname]`ColorScheme.Value.DARK`. The system preference values write `light-dark` or `dark-light`, which no `[theme~="dark"]` selector matches.
+====
+
 .Testing Color Schemes with Browser DevTools
 [NOTE]
 ====
--- a/articles/flow/ui-state/building-ui.adoc
+++ b/articles/flow/ui-state/building-ui.adoc
@@ -281,38 +281,6 @@ field.bindRequiredIndicatorVisible(required);
 For more details, see <<element-bindings#required-indicator-binding,Required Indicator Binding>> in Element Bindings.
 
 
-== Binding Theme Variants
-
-Use [methodname]`getThemeList().bind()` to toggle theme variants based on a signal:
-
-[source,java]
-----
-ValueSignal<Boolean> darkMode = new ValueSignal<>(false);
-
-VerticalLayout layout = new VerticalLayout();
-layout.getThemeList().bind("dark", darkMode);
-// Dark theme applied when darkMode is true
-
-Button toggleTheme = new Button("Toggle Dark Mode");
-toggleTheme.addClickListener(e -> darkMode.update(v -> !v));
-----
-
-
-=== Multiple Theme Variants
-
-Bind multiple independent theme variants:
-
-[source,java]
-----
-ValueSignal<Boolean> compact = new ValueSignal<>(false);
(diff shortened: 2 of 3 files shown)


articles/styling/themes/index.adoc
```diff
@@ -143,6 +143,18 @@ The color scheme can be changed dynamically at runtime using the [methodname]`Pa
 UI.getCurrentOrThrow().getPage().setColorScheme(ColorScheme.Value.DARK);
 ----
 
+Both the annotation and the method set the `theme` attribute and the `color-scheme` CSS property on the `<html>` element. That's why a single call covers both themes: Lumo selects its dark colors with `[theme~="dark"]`, while Aura selects its own through `color-scheme`.
```
jouni: The `theme=dark` attribute is only for backwards compatibility. Lumo is also built on `color-scheme` since V25.

```suggestion
```

articles/styling/themes/index.adoc
```diff
@@ -143,6 +143,18 @@ The color scheme can be changed dynamically at runtime using the [methodname]`Pa
...
+
+.The `dark` Theme Name Is Not a Second Mechanism
+[CAUTION]
+====
+Adding a `dark` theme name, as in `ui.getElement().getThemeList().add("dark")`, looks like an alternative but isn't one. [methodname]`getThemeList()` is the generic theme name API that every component has, and it's the right tool for variants such as `small` or `contrast`. It only happens to match Lumo's own `[theme~="dark"]` selector.
+
+It writes the attribute on the `<body>` element, since that's the element a [classname]`UI` is. Under Lumo that leaves the `<html>` element light behind whatever the body doesn't cover, and under Aura it has no effect at all. Neither case logs anything or throws, so the page stays light with no indication of why.
+
+In tests, assert on [methodname]`Page.getColorScheme()` rather than on the `theme` attribute. The attribute is only a plain `dark` for [propertyname]`ColorScheme.Value.DARK`. The system preference values write `light-dark` or `dark-light`, which no `[theme~="dark"]` selector matches.
+====
```
jouni, line 157:
```suggestion
.Avoid Using The `theme="dark"` Attribute Globally
[CAUTION]
====
Avoid using `ui.getElement().getThemeList().add("dark")` to switch to the dark color scheme globally.

[methodname]`getThemeList()` is the generic theme name API that every component has, and it's the right tool for variants such as `small` or `contrast`. 

When using the [propertyname]`ColorScheme.Value.DARK` value specifically, Vaadin adds the `theme="dark"` attribute on the `<html>` element for backwards compatibility. You should not rely on it going forward.

In tests, assert on [methodname]`Page.getColorScheme()` rather than on the `theme` attribute.
```

jouni (approved):

PR #6004, opened 2026-09-11
docs: split Chart Types into one page per chart type

### LINKED ISSUE #6003: Chart Types is a single 95 KB page: split it into one page per chart type
**Tell us about your request**

`articles/components/charts/charttypes.adoc` is by a wide margin the largest article in the docs: **95,161 bytes** on `v25.2` and **99,722** on `main`. It documents 27 chart families on one page.

Please consider splitting it — one page per chart type, or per family, under `components/charts/charttypes/`, with the current page becoming the section index.

**Why**

*For readers:* the page's own structure is already a list of independent topics. Each `==` section is self-contained — a description, its plot options, a data model where relevant, and an example — and they do not build on one another. Someone who wants Waterfall Charts scrolls past 17 chart types to reach it. The sections are evenly sized (median ~700 tokens, largest 1,771), so they would make reasonably uniform pages.

*For AI assistants, which is what prompted this:* the docs are consumed by the Vaadin documentation MCP server, which serves whole pages to coding agents. Converted to Markdown this page is **21,592 tokens** — the largest in a corpus of 1,165 pages whose median page is **1,045**. An assistant asked about one chart type either refuses to load the page (the original report, vaadin/vaadin-mcp#77) or truncates it, and truncation takes the page from the top: at a generous 8,000-token budget, 19 of the 28 sections are unreachable. No other page in the docs has this problem to this degree — the next largest is 16,391 tokens and the one after that 13,727.

`v24` measures the same (21,566 tokens for the converted page). `v14` has no charts articles, so this is a question about `v24` onwards.

**What this is not**

Not a blocker for the docs team. The MCP server is pursuing section-level retrieval independently, so agents will stop needing the whole page either way. Splitting upstream is the option that also helps human readers and every other consumer of the docs, which is why it is worth raising rather than solving purely on our side. If the page is considered right as it is for readers, saying so is a useful answer too.

Filed from vaadin/vaadin-mcp#77, where the measurements above are recorded in full.

peholmst:
Fixes #6003.

`components/charts/charttypes.adoc` documented 27 chart families on a single 95 KB page — by a wide margin the largest article in the docs. Its `==` sections were already independent of one another, so a reader who wanted Waterfall Charts scrolled past 17 chart types to reach it, and consumers that serve whole pages (such as the documentation MCP server) either refused to load the page or truncated it from the top.

This splits it into `components/charts/charttypes/`: one page per chart type, with the former page becoming the section index.

| | before | after |
| --- | --- | --- |
| pages | 1 | 29 |
| largest | 95 KB | 7.7 KB (`timeline.adoc`) |
| median | — | 3.4 KB |

## URLs and anchors

The index keeps the `charts.charttypes` anchor, and `charttypes/index.adoc` serves the same path that `charttypes.adoc` did — the pattern already used by `charts/styling/` — so the page's URL and the links into it are unchanged. Every section kept its own anchor (`charts.charttypes.pie`, and so on) on its new page, and the alphabetical `ChartType` table on the index now links to those pages. Deep links from `data.adoc` and `timeline.adoc` were repointed.

Pages are ordered as the sections were, which keeps related types together (Gauges before Solid Gauges, Error Bars before Box Plot) and leaves the three cross-cutting topics — Data Labels, Data Point Markers, 3D Charts — at the end. The index's alphabetical table covers lookup by `ChartType` name.

## Content changes

Content is carried over verbatim apart from what the split forces:

* Headings promoted one level, and the four lowercase `example` headings put in title case (these were pre-existing `Vaadin.HeadingCase` warnings).
* References that pointed at another section of the same page became cross-page links. Three sentences that read as "described previously" were reworded: Bubble Charts no longer says scatter charts "demonstrated" the same purpose, and the two gauge pages now link to each other.
* Image and sibling-page paths rebased onto `../`.
* A note on Timeline Charts distinguishes the `TIMELINE` chart type from the chart timeline navigator, now that those are two separate pages with the same name.
* `&mult;` in the 3D article was a broken HTML entity — it rendered literally as `2&mult;2&mult;2`. Fixed to `&times;` while rewording the same sentence to drop a first-person plural.

`Sankey` is added to the accepted vocabulary and `OHLC` to the `Vaadin.Abbr` exceptions. Both now appear in a page title, where they can't be spelled out on first use; the Sankey spelling errors predate this PR.

## Verification

* A round-trip check confirms all 2,036 content lines land in exactly one new page, unchanged apart from the transforms above.
* Vale with the CI config (`.vale-pr.ini`): **0 errors, 0 warnings** across all 29 pages. The single page it replaces had 8 errors and 5 warnings.
* Asciidoctor renders all 29 pages with no syntax warnings. The index table now renders as a clean 4×8 grid; the original emitted a stray empty row.
* An xref/anchor checker over `components/charts/` resolves every link in the new pages. Three failures remain in `gantt.adoc` and `styling/styling.adoc` — pre-existing, in files this PR doesn't touch.


DiegoCardoso (approved):
LGTM

PR #6019, opened 2026-09-15
feat: document the copilot panel

### LINKED ISSUE #394: Document the dev tools panel
No page under `articles/tools/observability` in the docs project mentions the dev tools panel. Developers who add the kit to a local project do not know it exists, what it shows, or that it is only injected outside production mode. The panel is what they see first, long before there is production traffic to look at.

Add a development-mode section to the getting-started page covering:

- How the panel appears in Copilot and how to open it.
- What it lists today: `vaadin.*` meters from the active registry, refreshed every few seconds.
- That it is injected only when production mode is off, and that it does not change what is exported.

Reuse the insights vocabulary from the insights and client-errors pages. Update the section again when #391 and the view profiler land.

The change itself goes to the docs project; this issue tracks it here.

caalador:
Add documentation for the
copilot panel for observability
that is available in development
mode.

fixes vaadin/observability-kit#394


articles/tools/observability/copilot-panel.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+The panel keeps watching while its window is closed, and writes a line to the Copilot log for each finding the payload didn't have before.
+This is the point of the panel for most of a working day: you don't have to have it open to learn that something broke.
+
+Announcements are deduplicated on the same grouping key the endpoint uses, so one problem notifies once, however often it recurs.
+Only findings that the current page raised are announced -- anything first seen since the page loaded, including during the load itself, so a slow query on the landing view is reported.
+The retained records outlive a reload, and those older findings are not announced again.
+
+Errors are logged as errors and everything else as a warning, each prefixed with `Observability:`.
+The message is a summary, and the server cuts anything longer than 300 characters.
```
heruan, line 73: The cut moved to the browser script together with the announcement, so the server is no longer involved.

```suggestion
The message is a summary, cut to 300 characters when longer.
```

articles/tools/observability/copilot-panel.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+[NOTE]
+Announcements are best-effort.
+Copilot's plugin API has no notification of its own, so the line is written by asking the server to send Copilot's own log command; a Copilot that drops it costs you a notification, never the panel.
```
heruan, line 77: Since vaadin/observability-kit#398 the script writes the line itself, as a `log` event on Copilot's event bus. There is no server round trip. Relaying through the server made the log panel show each line twice.

```suggestion
Copilot's plugin API has no notification of its own, so the line is written as a log event on Copilot's event bus; a Copilot that drops it costs you a notification, never the panel.
```

heruan (changes requested):
The page matches the panel closely, down to the verbatim UI strings, the refresh interval, ranking, route grouping, meter rendering and navigation order. Two sentences describe the announcement path that vaadin/observability-kit#398 replaced this week, see inline.

Heads up: #5961 adds a bullet at the same spot in the "See also" list of getting-started.adoc, so whichever merges second will need a small conflict fix.

articles/tools/observability/copilot-panel.adoc
caalador: done

articles/tools/observability/copilot-panel.adoc
caalador: done

heruan (approved):

caalador:
Some minor changes due to duplicate information from the other prs

PR #6039, opened 2026-09-15
docs: add an Agentic Development section with a dev loop tutorial

tltv:
Group the AI coding agent material under building-apps/agentic-development: move the MCP Server section in, and add a hands-on Dev Loop tutorial that installs the vaadin-dev CLI and skills, drives Claude Code and Codex through three changes, and measures what the session costs.

Public URLs move from /building-apps/mcp/** to
/building-apps/agentic-development/mcp/**, so redirects are needed.


articles/building-apps/agentic-development/dev-loop-tutorial.adoc
```diff
@@ -0,0 +1,333 @@
...
+* *Network access the first time you install*, to download HotswapAgent once per machine.
+* *https://www.claude.com/product/claude-code[Claude Code] or https://developers.openai.com/codex/cli[Codex CLI]*, installed and signed in.
+
+See <<{articles}/flow/configuration/live-reload/dev-loop-cli#prerequisites,Prerequisites>> for the full list.
+
+
+[[step-1]]
+== 1. Get the Base Application
+
+Open https://start.vaadin.com[Vaadin Start] and keep the sample view. The dev loop ships with Vaadin 25.3, so expand *Configure Project Settings* and select *25.3 (pre)* as the Vaadin version. Download the project, extract it, and open it in your IDE.
```
peholmst: When 25.3 ships, this needs to be updated (including the other references to 25.3 beta).

peholmst (commented):
Looks good to me, except the explicit mentions of 25.3 being a pre-release. That will not be the case when this page goes live with the 25.3 documentation in a week.

articles/building-apps/agentic-development/dev-loop-tutorial.adoc
tltv: Maybe it would be just fine to update it now for the 25.3.0 release, so no need to get back to this later?

peholmst: Yes, do that. And if possible, make it future proof so that we don't have to go back and edit it for every subsequent Vaadin version. There should also be a variable in `articles/_vaadin-version.adoc` that you can use if you only need to refer to the latest version (although it does not appear to be updated either).

tltv: Done. Note that preview will show old version after changed to use the variable, but that's separate issue.

peholmst (approved):

PR #130, opened 2021-01-22
Accessibility improvements

anezthes:

PR #6056, opened 2026-09-16
docs: add a self-hosting guide for the Vaadin MCP server

peholmst:
Adds a self-hosting section under `articles/building-apps/agentic-development/mcp/self-hosting/` for customers who want to run the Vaadin MCP server inside their own perimeter, plus a pointer from the MCP landing page.

Written against vaadin/vaadin-mcp#130, which is the spec for this work.

## Structure

The essentials are on the front page; detail and special cases are on subpages.

| Page | Carries |
|---|---|
| `index` | What self-hosting changes, what you need, the two deployment modes, pull → license → run → verify, connecting tools, and that the server authenticates nobody |
| `licensing` | Which key to ask for, both install routes and their traps, renewal, the expiry-warning channels, refusals |
| `configuration` | Environment-variable reference, reverse-proxy shapes, custom images |
| `air-gapped-operation` | Getting the image in, and the no-outbound-traffic statement |
| `monitoring-and-audit-logging` | What to scrape and alert on, and the audit trail |
| `sizing` | Measured figures, and scaling out |

## Coverage against vaadin-mcp#130

| Deliverable | State |
|---|---|
| Deployment guide — shared service and per-developer, same image | ✅ front page |
| Audit-logging configuration guide, incl. jurisdictional review | ✅ `monitoring-and-audit-logging` |
| Air-gap walk-in instructions (`docker save` / carry / `docker load`) | ✅ `air-gapped-operation` |
| License management — path, renewal, grace, location-agnostic | ✅ `licensing` |
| Operational monitoring — `/health` shape, expiry signals, log prefix | ✅ `monitoring-and-audit-logging` |
| No-outbound-traffic statement for security review | ✅ `air-gapped-operation` |
| Working nginx and oauth2-proxy reverse-proxy auth recipes | ❌ **not included, deliberately** |

On the last row: #130 assigns verification of those recipes to engineering against a running image, and they're security-sensitive — a subtly wrong auth config exposes an internal service. This working copy has no `vaadin-mcp`-entitled key, so nothing written here could be tested. The front page states plainly that the server ships no auth and that network placement is the operator's control point. The recipes belong in a follow-up.

## Facts verified against the server sources

Rather than assumed from the README:

- **One key covers every instance** — no machine binding, and nothing in `com.vaadin.mcp.license` counts seats, instances, or concurrent callers.
- **The 30-day soft-warning window before expiry**, which the first draft missed entirely (it covered only the grace period after). It surfaces on four channels; the one worth knowing in advance is that tool responses gain a `[PLEASE RELAY TO THE USER]` block inside both windows, so developers see renewal reminders in their agent's own output.
- **Hourly license re-probing** (`LicenseStatusProvider.refresh()`), which is why a replaced key *file* is picked up without a restart while an environment variable is not.
- **`/health` status values** are lowercase enum names (`valid`, `expiring_soon`, `in_grace`) via `LicenseStatusLabel`, with two components: `license` and `versionSnapshot`.
- **The `[vaadin-mcp]` log prefix**, from `application.yml`'s console pattern.
- **The image is amd64-only** — confirmed against Docker Hub, not inferred.

## Verification

- **asciidoctor**: all six pages render clean at `--failure-level=WARN`, and every same-page anchor resolves.
- **Vale: 0 errors, 0 warnings** on all six pages at the CI alert level (`.vale-pr.ini`).
- A full `articles/` scan goes from **572 errors to 571**, so the one-line vocabulary change adds nothing. That entry uses the case-insensitive `[pP]odman` form on purpose: an exact-case entry makes the built-in `Vale.Terms` rule demand that capitalization everywhere, trading one spelling error for many new ones.
- Vale caught a real defect early on: the front matter's `meta-description` had an **unquoted colon**, making the YAML invalid. The docs build parses that same block, so it would have failed at build time.

## Notes for review

- **The *There Is No Unlicensed Mode* note is gone.** It argued against a bypass that earlier builds had — history no customer shares, since nobody has run this server before — and read as a warning about something they were never going to try. What a reader needs is that a key is required, which the pages say plainly.
- **The hosted server's posture now lives on `mcp/index.adoc`, not here.** Where `mcp.vaadin.com` runs and what it reports is what a reader asks *before* deciding to point a tool at it, and that reader is on the MCP overview rather than in a self-hosting guide. It sits ahead of the tool cards, outside the `tags=supported-tools` region, so the setup guide's include is unaffected. The self-hosting page keeps a one-line pointer.
- **That posture is prose, not derived from anything.** Analytics-on, audit-off and EU hosting are not recorded in the server repo — consistent with its own note that nothing there deploys — so it needs maintaining alongside the deployment. The "same image" claim *is* backed by the repo, and now sits in *Running Your Own Server*, where it explains why a self-hosted instance answers identically.
- **The sizing figures are explicitly not a spec sheet.** They come from a natively built arm64 image on Apple Silicon, not the amd64 image customers deploy, and the page says so.
- **The custom-image section is framed as a request, not a task**, since `vaadin/vaadin-mcp` is not public. Every external link in the section is public.
- **Pre-existing and left alone:** `mcp/index.adoc:47` trips `Vale.Spelling` twice on "Dev" (from the Dev Loop CLI link text). Identical on `main` and unrelated to my edit there, so it's not folded into this PR.


articles/building-apps/agentic-development/mcp/self-hosting/index.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+*Every tool call can be recorded.* Point the server at a path and it appends one JSON record per call, so you have a log of what your agents asked for. See <<monitoring-and-audit-logging#audit-logging,Audit Logging>>.
+
+*Nothing is reported to Vaadin.* Analytics reporting stays off unless you set an Amplitude key, which a self-hosted deployment has no reason to do. See <<air-gapped-operation#analytics,Analytics>>.
+
+What you give up is corpus freshness. The hosted server picks up new documentation weekly without you doing anything; a self-hosted one is as current as the image you last pulled.
+
+
+[[hosted-server]]
+== What the Hosted Server Runs
```
peholmst: This should go into `mcp/index.adoc`, with a reference to the self-hosting page.

articles/building-apps/agentic-development/mcp/self-hosting/licensing.adoc
```diff
@@ -0,0 +1,104 @@
+---
+title: Licensing
+page-title: License a Self-Hosted Vaadin MCP Server Instance
+description: Which key the self-hosted MCP server needs, how to install it, and how renewal works.
+meta-description: Get and install a license key for a self-hosted Vaadin MCP server, renew it before expiry, and read what each startup refusal means.
+order: 10
+---
+
+
+= Licensing
```
peholmst: Simplify this entire page. It's difficult to understand what some parts mean ("like a key being hostage to the base image"). Explaining why a machine-bound key wouldn't work is just noise. A human system administrator should be able to have a look at this page and easily figure out what license key to ask for and what to do with it. No more, no less.

articles/building-apps/agentic-development/mcp/self-hosting/licensing.adoc
```diff
@@ -0,0 +1,135 @@
...
+docker run --rm -p 8080:8080 \
+    -v "$HOME/.vaadin/serverKey:/home/nonroot/.vaadin/serverKey:ro" \
+    vaadin/mcp-server:latest
+----
+
+Four requirements. If you miss any of them the server reports *no license file found* and stops, so check all four before looking elsewhere:
+
+. *Mount into `/home/nonroot/.vaadin/`.* That is the only directory the server searches. A key under `/root/.vaadin/` is never found.
+. *Keep the filename you downloaded.* The accepted names are `offlineKeyV2`, `offlineKeyV2.txt`, `offlineKey`, `offlineKey.txt`, `serverKey`, and `serverKey.txt`. Renaming the file makes it invalid.
+. *Make the file readable by user ID 65532*, which is the account the container runs as. Run `chown 65532:65532` on your copy of the key; on Kubernetes, use a secret with `defaultMode: 0400` and `runAsUser: 65532`. A root-owned key with `0600` permissions cannot be read. Docker Desktop on macOS remaps ownership for you, so this catches out Linux hosts only.
```
peholmst: **[P1] Kubernetes secret would be unreadable**

`runAsUser: 65532` changes the process UID, not the ownership of Secret files. A Secret with `defaultMode: 0400` is normally root-owned, so this configuration prevents the server from reading the key. Please provide a tested manifest using compatible ownership/group settings and permissions. Kubernetes documents volume ownership separately under `fsGroup`.

articles/building-apps/agentic-development/mcp/self-hosting/index.adoc
```diff
@@ -0,0 +1,151 @@
...
+* *Nothing is sent to Vaadin.* Usage reporting is off unless you turn it on, and there's no reason to. See <<air-gapped-operation#analytics,Analytics>>.
+
+The trade-off: your documentation is only as current as the image you last pulled, where the hosted server updates itself.
+
+To compare against what the hosted server does -- where it runs, and what it reports -- see <<{articles}/building-apps/agentic-development/mcp/#hosted-server,What the Hosted Server Runs>>. It runs the same image you would.
+
+
+== What You Need
+
+* *A container runtime on an x86_64 host.* Docker, Podman, or Kubernetes. The published image is `linux/amd64` only. An arm64 host such as Apple Silicon runs it through emulation, which is slower to start and slower to search -- fine for trying out, not for a deployment.
```
peholmst: **[P2] Advertised deployment modes lack instructions**

The prerequisites promise Docker, Podman, and Kubernetes support, but every runnable deployment example uses `docker run`. A shared production service still requires readers to invent the Deployment, Secret mount, health probes, audit volume, and security context. Please add at least one complete Compose or Kubernetes example and explain the Podman substitutions.

articles/building-apps/agentic-development/mcp/self-hosting/index.adoc
```diff
@@ -0,0 +1,151 @@
...
+Vaadin runs a hosted MCP server at https://mcp.vaadin.com/docs, and for most teams that's the one to use: there's nothing to operate, and its documentation is updated every week.
+
+Self-host when your organization can't send development queries to a service outside its own network. The same server is available as a container image you run on your own host.
+
+This page gets one instance running. The pages under <<#topics,Topics>> cover each part in detail.
+
+
+== What Self-Hosting Changes
+
+* *The server needs no network access.* All the documentation it searches is inside the image, so the container can run with no network route at all. See <<air-gapped-operation#,Air-Gapped Operation>>.
```
peholmst: **[P2] Air-gap wording disables client access**

A shared server cannot run with “no network route at all”: clients still need an inbound route to port 8080. The intended guarantee appears to be no outbound or internet route. Please use that wording consistently so an administrator does not deploy the container with networking disabled.

articles/building-apps/agentic-development/mcp/self-hosting/licensing.adoc
```diff
@@ -0,0 +1,135 @@
...
+
+A `status` of `valid` means the setup is done. If the server didn't start, see <<#license-refusals,If the Server Doesn't Start>>.
+
+
+[[renewal]]
+== Renewing the Key
+
+Put the new key in place the same way you installed the first one, then restart the container.
+
+A replaced key *file* is also picked up on its own within the hour, with no restart. A key passed as an environment variable always needs a restart.
```
peholmst: **[P2] Automatic key reload is not universal**

The claim that a replaced key file is picked up automatically needs qualification. Kubernetes Secret mounts using `subPath` do not receive updates, and atomically replacing a host file may leave a single-file bind mount attached to the old inode. Please state which mount patterns support live renewal; otherwise require a restart, as the preceding instruction already does.

articles/building-apps/agentic-development/mcp/self-hosting/monitoring-and-audit-logging.adoc
```diff
@@ -0,0 +1,112 @@
...
+== Monitoring
+
+Point your monitoring at `/health` -- at the root, not under `/actuator`. It needs no authentication:
+
+[source,bash]
+----
+curl -s http://localhost:8080/health
+----
+
+Use the top-level `status` field for liveness and readiness probes. The response also reports two components: `license`, with the current status and expiry date, and `versionSnapshot`, with the documentation versions the image was built from.
```
peholmst: **[P2] Probe instruction is not directly actionable**

Standard Kubernetes HTTP liveness and readiness probes evaluate the response status code; they do not inspect a JSON field. Please explain the endpoint’s HTTP status behavior and provide an `httpGet` example. Reserve parsing the top-level `status` field for monitoring systems that inspect the response body.

articles/building-apps/agentic-development/mcp/self-hosting/index.adoc
```diff
@@ -0,0 +1,151 @@
...
+
+Every build also gets a permanent tag that never moves:
+
+----
+vaadin/mcp-server:<version>-<YYYYMMDD-HHMMSS>-<commit>
+----
+
+For example, `vaadin/mcp-server:1.0.0-20260916-114956-2b7c708`.
+
+*Use a permanent tag in anything you deploy.* Both `:latest` and a plain version number change over time: the version number tracks the server code, so a documentation update republishes the same version under a new timestamp. Only the full tag identifies a single build.
```
peholmst: **[P3] Explain how to find the immutable tag**

The guide correctly requires a permanent tag but only provides its format and a dated example. An administrator has no stated way to discover the current tag. Please link directly to the image’s tags page or provide a command for selecting and verifying the desired immutable build.

peholmst (commented):
The writing is concise and well structured, but these operational details may prevent an administrator from successfully following the guide as written.

articles/building-apps/agentic-development/mcp/self-hosting/licensing.adoc
peholmst: Fixed in d58ca060d — you were right, and this one would have produced a server that could not start.

`runAsUser` sets the process UID and changes nothing about file ownership, so the Secret stays root-owned and `defaultMode: 0400` denies the read. The fix is `fsGroup: 65532` in the pod security context, which sets group ownership on the mounted volume and adds the group to the process, with `defaultMode: 0440` to grant the group read.

The licensing page now states that `runAsUser` alone is not enough, and there is a full Deployment manifest on the new Deployment Examples page with `fsGroup` and the `items:` list (the latter keeps the mounted directory to your key alone, which is what the surrounding "mount one file, not a directory" rule needs).

One caveat worth your judgement: I have no cluster to apply this against, so the manifest is correct by the documented `fsGroup` semantics but is not empirically tested. Worth one `kubectl apply` before this merges.

articles/building-apps/agentic-development/mcp/self-hosting/index.adoc
peholmst: Fixed in d58ca060d. Agreed, and the wording was actively dangerous — "no network route at all" invites `--network none`, which breaks every client.

All three places now say no *outbound* route and name the inbound one explicitly. The bullet reads: "The server makes no outbound connections. All the documentation it searches is inside the image, so it needs no route to the internet — only an inbound route from your developers' tools."

One use of "no network route" survives on the air-gap page, deliberately: it describes Vaadin's own pre-release test, which really does run with no route because the test client shares the container's network namespace. That is a statement about the test, not an instruction to the reader.

articles/building-apps/agentic-development/mcp/self-hosting/licensing.adoc
peholmst: Fixed in d58ca060d. Dropped the claim — both failure modes you name are real, and I could not enumerate the safe mount patterns with any confidence.

Restarting is now the instruction, with the reason there is no shortcut: "The server does re-read its key every hour, but whether a replaced file reaches it depends on how you mounted it: a Kubernetes Secret mounted with `subPath` never receives updates, and replacing a bind-mounted file on the host can leave the container reading the original file. Restart the container and you do not have to work out which case you are in."

The hourly re-check is still documented on the monitoring page, where the claim is only that `/health` reflects the *passage of time* toward expiry. That holds for any mount, since it re-evaluates the same key rather than needing a new one.

articles/building-apps/agentic-development/mcp/self-hosting/monitoring-and-audit-logging.adoc
peholmst: Fixed in d58ca060d. The page now gives an `httpGet` example and states the status-code behaviour: 200 while serving, 503 when the documentation fails to load.

Checking the sources turned up something better than the comment asked for, and it changes the advice. `LicenseHealthIndicator` returns `up()` for every reachable status — `valid`, `expiring_soon` and `in_grace` alike — so **an expiring or expired key never moves the status code**. A probe cannot warn you about it. The only DOWN branch is `VersionSnapshotHealthIndicator` when the documentation did not load.

So the two are now split deliberately: probes read the status code, and the license needs its own alert on the response body. The page says that outright ("A license problem never changes the status code") rather than leaving a reader to infer it.

articles/building-apps/agentic-development/mcp/self-hosting/index.adoc
peholmst: Fixed in d58ca060d. Added a "Finding a Tag" section with a link to the tags page on Docker Hub and a command I ran against the live registry:

```
curl -s 'https://hub.docker.com/v2/repositories/vaadin/mcp-server/tags?page_size=25' \
    | jq -r '.results[] | select(.name != "latest") | .name' | sort -r | head
```

Sorting by name works because the timestamp is inside the tag, so the newest build sorts first. It also notes how to confirm afterwards which build is running: the `vaadin-mcp://image-info` resource reports the same commit the tag names.

articles/building-apps/agentic-development/mcp/self-hosting/index.adoc
peholmst: Fixed in d58ca060d. Fair — three runtimes were promised and only `docker run` was shown.

There is now a Deployment Examples page with:

- a Compose file, including the host-side `chown` the bind mounts need;
- a Kubernetes Deployment covering the Secret mount with `fsGroup`, readiness and liveness probes, resource requests, and the audit volume;
- the one Podman difference that matters — rootless UID mapping means a host-side `chown 65532` is not user 65532 inside the container, so the mount needs the `:U` flag or the environment-variable route.

Two things the manifests need that the review did not raise, and I would not have found without writing them out:

- **The image is distroless**, with no shell and no HTTP client, so a Docker or Compose `healthcheck:` running a command inside the container cannot work at all. Kubernetes `httpGet` probes are fine because Kubernetes sends them from outside. This now has its own short section.
- **The audit directory has the license mount's ownership problem in the other direction**: it must be *writable* by 65532, and a Docker named volume and a Kubernetes PVC are both created root-owned. Since an unopenable audit path stops startup, that was a guaranteed first-deploy failure.

As on the P1 comment: the Kubernetes manifest has not been applied to a real cluster from here.

PR #6059, opened 2026-09-16
docs: document reverse proxy configuration for the SSE push transport

mcollovati:
Server-Sent Events is a plain HTTP GET that never ends, so none of the WebSocket upgrade configuration applies to it. What it needs instead is that the proxy does not buffer or close the response stream.

Add a Server-Sent Events section covering those requirements, the browser connection limit, the WebSocket-only Hilla reactive endpoints, TLS termination, and how to verify a deployment. Add the required directives to the configuration examples, marked with "Server-sent events only" comments:

- proxy_http_version, proxy_buffering and proxy_read_timeout on the nginx push locations. The push location is no longer marked as WebSockets only, since both transports need it.
- flushpackets=on on the AJP workers, which lets push travel over AJP and removes the need for a second channel to the HTTP port of the backend.

Also switch the path specific Apache HTTPD rules from ws:// to upgrade=websocket, so that a single rule serves both transports, and declare them before the catch-all. Without that ordering, mod_proxy shares the catch-all worker and silently discards the upgrade option.

Add a Compression and Push section. Compressing a streamed response does not stall it, since both mod_deflate and the nginx gzip filter flush after each event, but it is still worth avoiding: a compression context stays allocated for the life of every push connection, and the 2000 bytes of Atmosphere padding shrink to a few dozen bytes, defeating their purpose of forcing an intermediary to flush. PUSH answers with text/event-stream over Server-Sent Events and with text/plain over streaming and long polling, so keeping both out of gzip_types and AddOutputFilterByType is what matters, and it holds wherever the push endpoint is mapped.

Rename the WebSocket Connection Timeout section to Push Connection Timeout, as the heartbeat applies to every transport, and raise the timeout from 90 to 300 seconds. The default of 60 seconds coincides exactly with the heartbeat interval, leaving no margin.


mcollovati:
The documented configurations have been tested [here](https://github.com/mcollovati/vaadin-reverse-proxy-tests), and smoke tests for over 60 different configurations are passing (https://github.com/mcollovati/vaadin-reverse-proxy-tests/actions/runs/35142512190)

PR #6076, opened 2026-09-18
docs: Document how the browserless test environment differs from a running application

mcollovati:
Follow-up to vaadin/browserless-test#201, which reported four ways a browserless test behaves differently from a running application. Two of them were bugs and are fixed in the framework; two are consequences of the test lifecycle and were never documented. This covers all of it on the docs side.

## New page: Environment Differences

`articles/flow/testing/browserless/environment-differences.adoc`, placed right after Getting Started.

- **Test lifecycle** — the order in which a test method sets things up: Spring field injection, then the security context, then the Vaadin environment and the initial navigation, then the test body. Neither of the two items below is derivable without it.
- **Session-scoped beans** — a `@VaadinSessionScope` bean can't be autowired into a test class field, because the field is injected before a session exists. Worth noting that it fails every test in the class, not only the ones using the bean. Resolve it from the test method with `getBean(...)`, `ObjectProvider`, or `@Lazy`.
- **Authentication applied during a test** — a sign-in performed after setup leaves the view rendered for the previous user in place, which reads as a broken security configuration. Navigate again. `reload()` is not a substitute, since it re-renders the location access control already redirected to.

## Getting Started

The Navigating to Views section listed only the path and parameter forms. A location string can carry a query string and a fragment as of vaadin/browserless-test#204, and that was never documented.

## Spring Security

The page states that authentication details are available before the UI is created and the default route is navigated to. Added the boundary: that holds when the authentication is in place before the test method starts.

---

Draft for review of scope and placement. The behavior described was verified against the current `browserless-test` main.


articles/flow/testing/browserless/environment-differences.adoc
```diff
@@ -0,0 +1,116 @@
...
+        Cart cart = applicationContext.getBean(Cart.class);
+
+        test(view.addButton).click();
+
+        Assertions.assertEquals(1, cart.getItems().size());
+    }
+}
+----
+
+Injecting [classname]`ObjectProvider<Cart>` or annotating the field with [annotationname]`@Lazy` works as well, because both defer resolution until the bean is first used.
```
Artur-: Would this be an easier to use method to suggest as the default, instead of injecting ApplicationContext?

articles/flow/testing/browserless/environment-differences.adoc
```diff
@@ -0,0 +1,116 @@
...
+    // Navigating again applies access control to the current authentication.
+    navigate(AdminView.class);
+
+    Assertions.assertTrue(find(Avatar.class).single().isVisible());
+}
+----
+
+.Reloading Isn't Enough
+[NOTE]
+[methodname]`reload()` isn't a substitute for navigating again. It re-renders the current location, and when access control has redirected to the login view, that location is the login view.
```
Artur-: Is this how it should work or a limitation?

articles/flow/testing/browserless/environment-differences.adoc
```diff
@@ -0,0 +1,116 @@
...
+
+Navigate again after signing in:
+
+[source,java]
+----
+@Test
+@WithMockUser(username = "admin", roles = "ADMIN",
+        setupBefore = TestExecutionEvent.TEST_EXECUTION)
+void adminSignsIn_adminViewShown() {
+    // The initial navigation ran while the user was still anonymous.
```
Artur-: By just looking at the code example, it's not clear what this refers to as there is no "initial navigation" anywhere

articles/flow/testing/browserless/environment-differences.adoc
```diff
@@ -0,0 +1,116 @@
...
+}
+----
+
+Injecting [classname]`ObjectProvider<Cart>` or annotating the field with [annotationname]`@Lazy` works as well, because both defer resolution until the bean is first used.
+
+Views and other components aren't affected: they're instantiated during navigation, inside the test method, so their own session-scoped dependencies resolve normally.
+
+
+[#authentication-applied-during-a-test]
+== Authentication Applied During a Test
```
Artur-, line 80: If you only have a quick look at this page, you might very well think this is how you should to authentication in a test. Could it maybe start by showing the correct way and only after that explain why you would want another approach and how you deal with things if you pick that approach?

articles/flow/testing/browserless/getting-started.adoc
```diff
@@ -202,6 +202,17 @@ To navigate to another registered view, use the [methodname]`navigate()` methods
 
 All navigation methods return the instantiated view, so that its fields can be used directly for testing.
 
+A location string can also [since:com.vaadin:vaadin@V25.3]#carry a query string, a fragment, or both#. They are split off from the path the same way the browser address bar does it, so the query parameters reach the view through the navigation event:
```
Artur-: This should just be another bullet point in the list above

articles/flow/testing/browserless/getting-started.adoc
```diff
@@ -202,6 +202,17 @@ To navigate to another registered view, use the [methodname]`navigate()` methods
...
+
+[source,java]
+----
+navigate("orders/1?tab=history&page=2", OrderView.class);
+----
+
+Query parameters can be passed as a separate [classname]`QueryParameters` object through [methodname]`UI.getCurrent().navigate(String, QueryParameters)`. Providing them both ways at once is rejected, because the query string in the location would be dropped.
```
Artur-: Isn't this the wrong way to do it as then it would simulate a server side navigation instead? Could just drop this if you can use `navigate` with parameters

articles/flow/testing/browserless/environment-differences.adoc
```diff
@@ -0,0 +1,116 @@
+---
+title: Environment Differences
+page-title: How the Vaadin browserless test environment differs
+description: What the browserless environment creates, in which order, and which application behavior therefore needs a different approach in a test.
+meta-description: Learn the browserless test lifecycle in Vaadin, and how it affects session-scoped beans and authentication applied during a test.
+order: 12
```
Artur-: This page seems way to high up, it is more of an advanced topic that should be closed to the end

articles/flow/testing/browserless/environment-differences.adoc
mcollovati: Yes, agreed — it is now the default in the example, with `@Lazy` and `ApplicationContext.getBean(...)` mentioned after it as equivalents. `@VaadinSessionScope` has no `proxyMode`, so the only thing that matters is that the lookup happens inside the test method rather than at field injection time, and `ObjectProvider` says that with the least ceremony.

articles/flow/testing/browserless/environment-differences.adoc
mcollovati: How it should work. `MockPage.reload()` closes the UI, recording `internals.activeViewLocation`, and `MockVaadin.createUI` navigates the new UI back to that location — which is exactly what pressing F5 does in a browser. Since access control redirected the earlier navigation to the login view, that recorded location *is* the login view. The note now says it mirrors a browser reload, instead of reading like a shortcoming.

articles/flow/testing/browserless/getting-started.adoc
mcollovati: You are right, and it is dropped. There is no `navigate(String, QueryParameters)` on the test base class, so that line went through `UI` directly, which skips the target view validation the `navigate()` methods do. The location string carries the query parameters since vaadin/browserless-test#204, so nothing is lost by removing it.

articles/flow/testing/browserless/environment-differences.adoc
mcollovati: Restructured that way: the section now opens with a plain `@WithMockUser` and a test that just navigates, and only then covers a sign-in performed during the test, why the view does not follow it, and the re-navigation. The "prefer doing it before the test method" paragraph that used to close the section is gone, since that is now the opening.

articles/flow/testing/browserless/environment-differences.adoc
mcollovati: Reworded to say what happened instead of naming it: "Setup navigated to the root route while the user was still anonymous, and access control redirected that navigation to the login view."

articles/flow/testing/browserless/getting-started.adoc
mcollovati: Done — it is a fourth bullet now, in the same shape as the other three, with the split-up and fragment-only details as its continuation.

articles/flow/testing/browserless/environment-differences.adoc
mcollovati: Moved to `order: 80`, so it sits after the framework-specific pages and before Optimizing Tests and Migration.

Artur- (approved):

PR #6084, opened 2026-09-18
fix: dspublisher start progress bar never reaches the ready state

tomivirkki:
## Summary
- `npm run dspublisher:start` stays at "Starting up DSP" with a full progress bar even after both servers are running, and the spinner stops.
- The start script assumed the ready signals arrive in phase order, but Astro reports ready long before the Spring Boot docs server. The early Astro signal triggered the finished state, and the later Spring Boot signal overwrote the label with the "next" phase text.
- The phase tracking is now order-independent: a phase is marked done when its signal appears, the label shows the first phase still pending, and the finished state is reached only once every phase has reported.


web-padawan (approved):
LGTM, seems to work as expected 👍

PR #6090, opened 2026-09-21
docs: explain what a browserless query can and cannot find

mshabarov:
## Summary

`find()` walks the server-side component tree, and two kinds of component are not in it when a test looks for them: a component that a renderer creates per item, and the content of an overlay that is closed. Both return an empty result rather than an error, so the failure reads as a missing component instead of an unrendered one. The *Querying Components* page did not mention either case, and the repository documented it only in its README (`vaadin/browserless-test#219`).

The same section records the fix in `vaadin/browserless-test#203`: a component set as a Grid column header, footer, or editor is now part of the tree, where a query used to fail with "no such component". It is reported once, even in a header cell that spans several columns.

## What changed

In `articles/flow/testing/browserless/component-query.adoc`, a new *Components a Query Cannot Find* section with two subsections:

- *Components a Renderer Creates* — why `find(Checkbox.class)` finds nothing for a component column, how `GridTester.getCellComponent(...)` renders the cell on request, the warning that every call attaches a new instance, the `LitRenderer` methods for columns that have no server-side component, and the `V25.3` note about header, footer, and editor components.
- *Overlay Content* — why a top-level query does not see a closed overlay's content, with a cross-reference to *Testing Overlay Components*, which already covers the tester-based approach.

## Notes for the reviewer

- The overlay page's *Common Pitfalls* section already says the same thing for context menus. The new section links to it rather than repeating the guidance, so the query page carries the general rule and the overlay page keeps the component-specific detail.
- `vaadin/browserless-test#203` shipped in 1.2.0-rc1, which is what Vaadin 25.3 ships, hence the `V25.3` badge on that one sentence. The rest of the section describes behavior that has always been there.

## How to test

Documentation only, with inline snippets that are not compiled. The behavior matches `GridTester` and `ContextMenuTester` on `vaadin/browserless-test` `main`, where tests pin both cases.


articles/flow/testing/browserless/component-query.adoc
```diff
@@ -240,4 +240,53 @@ TextField textField = findInView(VerticalLayout.class)
 ----
 
 
+== Components a Query Cannot Find
+
+A query walks the server-side component tree. A component that another component renders per item, and the content of an overlay that is closed, are not in that tree, and a query returns an empty result for them instead of an error. The failure therefore reads as though the component was never created.
+
+
+=== Components a Renderer Creates
```
mcollovati, line 248: This paragraph should be revisited to be consistent with changes in vaadin/browserless-test#250 that should prevent renderer components to be created and attached again and again.

mcollovati (changes requested):

mcollovati (approved):

PR #6091, opened 2026-09-21
docs: describe how an add-on pins npm versions

mshabarov:
Documents vaadin/flow#25308

**DX/docs** · building-apps/components · add-on authors who depend on
npm packages

Pinned npm versions used to come from two files that only the platform
ships. Flow now reads every versions file it finds in
`META-INF/VAADIN/versions/` on the classpath, so an add-on can pin the
npm versions its Java code needs. Nothing in the docs described the
folder, the file format, or what happens when two files disagree.

**Risks:**
- ✅ Docs only. No behavior change, no code examples to compile.

- Added a "Pinning npm Versions" section to Package a Component, with
  a since badge for the release that introduced the folder.
- Described the file format: an object with `npmName` plus
  `npmVersion` or `jsVersion` is one pinned package, and the keys
  around it only group entries.
- Described the merge rules: newest version wins when two files
  disagree, a package without a version is left unpinned, and an
  unreadable file is skipped on its own.


Artur-:
This is for internal use, not for public consumption

PR #6092, opened 2026-09-21
docs: show that a browserless navigation location takes a query string

mshabarov:
## Summary

`navigate("order/ORD-1?tab=history", OrderView.class)` used to fail with `Base path can not contain query separator=?`, so a test could not reach a view through a location with query parameters. The mocked UI now parses the location the way Flow's `UI.navigate` does (`vaadin/browserless-test#204`), and the query parameters reach the view. Nothing on the *Getting Started* page said the location could carry them.

## What changed

In `articles/flow/testing/browserless/getting-started.adoc`, under *Navigating to Views*: one badged sentence, a snippet, and a line saying that the map taken by `navigate(Template.class, Map)` holds route template parameters, which is the mistake the old wording invited.

## Notes for the reviewer

- The fix shipped in Browserless Test 1.2.0-rc1, which is what Vaadin 25.3 ships, hence the `V25.3` badge.
- A location that is only a fragment identifies no route, so the current view stays in place, and giving both a query string and a separate `QueryParameters` argument to `UI.navigate` now fails with a message naming the location. Neither case seemed worth a paragraph on a getting-started page; say the word if they belong here.

## How to test

Documentation only, with an inline snippet that is not compiled. The behavior matches `MockedUI.navigate` and the navigation tests on `vaadin/browserless-test` `main`.


mcollovati:
This has also been documented here vaadin/docs#6076 along with other changes.
Maybe better keep this PR and remove the duplication from 6076

articles/flow/testing/browserless/getting-started.adoc
```diff
@@ -200,6 +200,15 @@ To navigate to another registered view, use the [methodname]`navigate()` methods
...
 
+[since:com.vaadin:vaadin@V25.3]#The location is written the way it appears in the browser's address bar, so it can carry a query string and a fragment#:
+
+[source,java]
+----
+OrderView view = navigate("order/ORD-1?tab=history", OrderView.class);
+----
+
+The view reads those query parameters from the navigation event, as it does in a running application. The map that [methodname]`navigate(Template.class, Map)` takes holds route template parameters instead, so a query string belongs in the location string.
+
```
mcollovati, line 211: A potential addition from #6076 

>A location consisting only of a fragment, such as `"#details"`, identifies a place within the current page rather than a route, so it leaves the current view in place.

PR #6096, opened 2026-09-21
docs: index.html is generated into frontend/generated again

mshabarov:
## Summary

The docs still said the default `index.html` is generated into the `frontend` folder (and, in the bootstrap page article, into `target`). Since Vaadin 25.3 it is generated into `frontend/generated/` instead, so these pages now describe the new location and explain what upgrading projects should do.

## What changed

This is a documentation-only change; no code or API is affected.

- **Modifying the Application Shell:** the default `index.html` and `index.tsx` are generated into `frontend/generated/`, and neither file has to exist in the project. A `frontend/index.html` is optional and only overrides the generated default. To customize the shell, copy the generated file into `frontend` rather than write a new one from scratch. The new location is marked as new in 25.3 for `index.html` only — `index.tsx` was already generated there.
- **Source control:** the file tree now shows `index.html` under `frontend/generated/`, which stays out of source control. Only a project's own `frontend/index.html` should be committed.
- **Upgrading guide:** a new section covers the 25.3 change. An upgraded project usually has a committed `frontend/index.html` that the older build generated. It still takes precedence, so an uncustomized copy shadows the default and misses later improvements. The advice: compare it with the default, delete it if it was never customized, and keep it only for real shell customizations.


articles/flow/advanced/modifying-the-bootstrap-page.adoc
```diff
@@ -18,25 +18,25 @@ The _Application Shell_ in Vaadin Flow is also known as the _Bootstrap Page_, or
 
 The developer has full control of the contents of [filename]`index.html`. You can modify it in various ways.
 
-On the client side, you can edit [filename]`frontend/index.html` when the content is static, for instance the `<viewport>` tag. On the server side, you can make changes that require dynamic server content, or when Java syntax is preferred. An example of would be making the application installable by enabling the `@PWA` built-in feature.
+On the client side, you can add an [filename]`index.html` of your own to the [filename]`frontend` folder and edit it when the content is static, for instance the `<viewport>` tag. On the server side, you can make changes that require dynamic server content, or when Java syntax is preferred. An example of would be making the application installable by enabling the `@PWA` built-in feature.
```
Artur-: Could mention that you should copy frontend/generated/index.html and not create from scratch

PR #400, opened 2021-04-23
fix: explain file removal errors and give recommendations

mshabarov:
Related-to https://github.com/vaadin/flow/issues/9948

PR #6098, opened 2026-09-21
fix: update observability documentation (#5958)

caalador:
Update the documentation of
observability-kit to match
changes that made statements faulty.


heruan (approved):
Line for line the same as what merged on main in #5958, including the `vaadin.executor.task` rename and the `bootstrap` and `stream` request types. The anchors it links to exist on v25.3, and the client errors page is already there. Keeping the Copilot panel section inline in insights.adoc is right, since the separate page from #6019 is not on this branch.

One note: kit 5.0.0-rc1 predates the kit changes these pages describe (vaadin/observability-kit#397 to #400). They ship in 5.0.0 final, which is cut from main, so this is fine as long as that is the plan.

PR #6099, opened 2026-09-21
docs: explain the warning about undelivered JavaScript invocations

mshabarov:
Documents vaadin/flow#25252

**memory leak** · flow/component-internals, flow/production ·
applications that schedule `executeJs()` from background threads or for
components that are not on screen

JavaScript scheduled on the server is held in memory until it's written
into a response for the browser, so an application that keeps
scheduling invocations which never reach the browser runs out of
memory. Flow now logs a warning once 1000 of them are waiting for one
UI. Nothing in the docs described the retention, the warning, or the
alternatives, and the symptom is hard to recognize from a heap dump.

**Risks:**
- ✅ Docs only. One new section and one troubleshooting entry, no
  moved anchors.

- Added an "Undelivered Invocations" section to Calling JavaScript from
  the Server, with the anchor `undelivered-invocations`.
- Listed the three cases where invocations pile up: a detached owner,
  an invisible owner, and a UI with no open push connection.
- Listed the alternatives for values where only the latest one matters:
  `cancelExecution()` on the returned `PendingJavaScriptResult`,
  `Element.setProperty()`, and `UI.getLastUpdateSentTimestamp()` to
  stop scheduling while the browser isn't receiving.
- Added an entry to Common Issues on the production troubleshooting
  page, pointing at the new section.


articles/flow/component-internals/element-api/calling-javascript.adoc
```diff
@@ -164,6 +164,25 @@ CompletableFuture<Map<String, Person>> future = getElement()
...
+Invocations pile up whenever nothing carries them to the browser:
+
+- *The owner is detached.* An invocation for a detached element is held until that element is attached again, which never happens for a component the application has discarded.
+- *The owner is invisible.* Invocations for an invisible component are kept until it becomes visible.
+- *The UI has no open push connection.* Nothing is sent until the browser makes a request of its own, which a closed tab whose session hasn't expired yet never does.
+
+Where only the latest value matters -- a progress value, a clock, a live measurement -- avoid scheduling a new invocation for every update:
+
+- Keep the [classname]`PendingJavaScriptResult` that [methodname]`executeJs()` returns, and call [methodname]`cancelExecution()` on it before scheduling the next one.
+- Set an element property with [methodname]`Element.setProperty()` instead. Only the last value of a property is sent.
```
mcollovati, line 182: It seems not clear to me how element properties are related to pending JavaScript invocation. Do we have documentation that suggests setting properties via JavaScript?

mcollovati (changes requested):

articles/flow/component-internals/element-api/calling-javascript.adoc
Artur-: Maybe it should be phrased as "If you set an element property, use Element.setProperty instead of executeJS..."

PR #6094, opened 2026-09-21
docs: describe per-test Vaadin configuration for browserless tests

mshabarov:
## Summary
Documents how a browserless test can set Vaadin application properties, feature flags, and Lookup services for a single test class or test method. This is documentation only — no product code changes.

## What changed
Purely additive docs. No existing page content was removed or rewritten.

- **New page** `articles/flow/testing/browserless/test-configuration.adoc`: what `@BrowserlessTestConfig` sets, how class, superclass, `@Nested`, and method declarations merge, and the programmatic equivalents (`BrowserlessExtension` builder, `BrowserlessApplicationContext.Builder`, `BrowserlessConfiguration.from(...)`, and a `testConfiguration()` override). It also notes that Spring properties from the Spring environment still win, and that `BrowserlessClassExtension` rejects a method-level annotation because it builds one environment per class.
- **Extensions page**: adds the new `withApplicationProperty`, `withFeatureFlags`, and `withConfiguration` builder methods to the table, plus a link to the new page.
- **Multi-user page**: adds the same builder methods for the application context, and notes that the context does not read `@BrowserlessTestConfig` itself.
- **Feature flags page**: points out that a browserless test can enable a flag for one test, overriding the other flag sources.
- **Upgrade guide**: describes a 25.3 behavior change in the product — the Spring and Quarkus services moved from `lookupServices()` to the new `frameworkLookupServices()` and are always registered, so an override of `lookupServices()` now adds to them instead of replacing them. `lookupServices()` is deprecated in favor of the test configuration. This affects users who override `lookupServices()` in a browserless test.

PR #6089, opened 2026-09-21
docs: cover the browserless testers and methods added in 25.3

mshabarov:
## Summary

Browserless Test 1.2 is what Vaadin 25.3 ships. It adds seven testers and a set of tester methods that the *Component Testers* page does not mention, and it changes what a value tester does with a value that breaks a constraint. The page currently tells a test author the opposite of what the code does.

Covered here:

- `vaadin/browserless-test#211` — `NumberFieldTester`, `DatePickerTester`, `TimePickerTester` and `DateTimePickerTester` no longer throw on `min` / `max` / `step` / required violations. They commit the value and leave the field invalid, and `isValid()` asserts the outcome. The page's *Constraint Enforcement* example claimed the opposite.
- `vaadin/browserless-test#189` — `TreeGridTester` with `expand`, `collapse`, `isExpanded` and `hasChildren`.
- `vaadin/browserless-test#197` — `SplitLayoutTester`, `CardTester` and `AvatarGroupTester`.
- `vaadin/browserless-test#187` — `GridTester.deselect(int)` and `deselectAll()`.
- `vaadin/browserless-test#186` — `AccordionTester.closeDetails()` and `toggleDetails(String)`.
- `vaadin/browserless-test#191` — `NumberFieldTester.stepUp()` / `stepDown()` and `isValid()`.
- `vaadin/browserless-test#183` — `DialogTester.pressEscape()` and `clickOutside()`.
- `vaadin/browserless-test#168` — a uniform `clear()` plus the new `clickClearButton()`.
- `vaadin/browserless-test#185` — `CheckboxTester.check()` / `uncheck()` and `SwitchTester.switchOn()` / `switchOff()`.
- `vaadin/browserless-test#198` — `UploadTester` enforces `maxFiles`, `maxFileSize` and accepted types, and adds `getLastUploadStatus()` and `ensureUploaded()`.
- `vaadin/browserless-test#188` — tester interactions reach the application as client-originated events.

Four testers released earlier, in 1.1.2, never reached the docs either, so they are added with a `V25.2` badge: `DashboardTester` (`vaadin/browserless-test#105`), `BreadcrumbsTester.getItemPaths()` (`vaadin/browserless-test#107`), `TextTester` (`vaadin/browserless-test#135`) and `MarkdownTester` (`vaadin/browserless-test#132`).

## What changed

In `articles/flow/testing/browserless/component-testers.adoc`:

[...]

PR #6101, opened 2026-09-21
docs: add the browserless behavior changes to the upgrade guide

mshabarov:
## Summary

Four changes in Browserless Test 1.2, which is what Vaadin 25.3 ships, break tests that pass today. The *TestBench and Browserless Testing* section of the upgrade guide mentions none of them, so a suite that starts failing after the upgrade has nothing to point to.

- `vaadin/browserless-test#211` — `NumberFieldTester`, `DatePickerTester`, `TimePickerTester` and `DateTimePickerTester` commit a value that violates `min`, `max`, `step` or a required constraint, instead of throwing. Every `assertThrows(IllegalArgumentException.class, ...)` around those calls fails.
- `vaadin/browserless-test#139` — `isUsable()` returns `false` for a read-only value component, so setting a value on a read-only field through its tester now throws.
- `vaadin/browserless-test#203` — a component set as a Grid column header, footer or editor is found by a query, where the lookup used to fail.
- `vaadin/browserless-test#190` — `CheckboxGroupTester.updateSelection` is private and no longer appears on the generated locator.

## What changed

Four subsections in `articles/upgrading/index.adoc`, under the existing *TestBench and Browserless Testing* heading, each saying what changed and what to write instead. The first carries a before-and-after snippet, and links to *Component Testers* for the full rule.

## Notes for the reviewer

- The `isUsable()` entry documents a change that `#6034` already described on the *Component Testers* page. It is repeated here because it breaks existing tests silently, which is what this guide is for. Drop the subsection if the duplication is unwanted.
- The cross-reference to `component-testers#constraint-enforcement` points at the section rewritten in vaadin/docs#6089, which explains which constraints still make a tester throw. Merging that first keeps the link meaningful, though the anchor exists either way.
- The `lookupServices()` change from `vaadin/browserless-test#148` is not here: that API is still unreleased, and its upgrade note travels with vaadin/docs#6094.

## How to test

Documentation only, with an inline snippet that is not compiled. The behavior matches the testers on `vaadin/browserless-test` `main`.


mcollovati:
Some more changes that should be mentioned in the upgrade guide:

* vaadin/browserless-test#252 complements vaadin/browserless-test#211, by throwing exceptions for inputs that the browser does not allow at all.
* vaadin/browserless-test#250 prevents `GridTester` creating a new component instance every time`getCellComponent` is called. Old behavior can be kept using `renderCellComponent` method.

Maybe also worth mentioning that `test` methods added to the `TesterWrapper` interface for `TreeGridTester` and `GridContextMenuTester` can break compilation, since previously a different overload with a different return type was taken.

mcollovati (approved):

PR #6104, opened 2026-09-21
docs: align the AI integration docs with the latest changes

tomivirkki:
## Summary
- Several AI integration pages described behavior that has since changed: Spring AI tool call limits now fail the turn like LangChain4j, `ToolCallLimitExceededException` lost its accessors, `restoreFieldSource()` was removed, form fields are found inside composites and addressed by their component id, and the typing indicator replaced the empty assistant message.
- Tool Call Limits now lives on the Tool Calling page, with only a pointer left under LLM Providers, so the loop, its limits, and its failure mode are explained in one place.
- The DatabaseProvider and chart pages now say which query-derived values the chart state tool does expose to the LLM (categories, series names, organization chart nodes), instead of claiming that results never reach it.


ugur-vaadin (approved):
LGTM

PR #6121, opened 2026-09-22
docs: cover in-view access control in the security guides

peholmst:
## Problem

The security guides under Building Apps explain *which views* a user can reach, but not how to limit what they can do once inside one. The only in-view material was a short `AuthenticationContext` example under *Programmatic View Security* with no guidance on how to act on the role check.

That gap leaves CSS looking like a valid way to hide or disable privileged controls — it isn't. CSS only changes how a component looks in the browser; on the server it stays visible and enabled and keeps accepting events, so anyone can restore it from devtools and use it.

## Changes

**`articles/building-apps/security/protect-views.adoc`** — promoted *Controlling Access within Views* from a subsection of *Programmatic View Security* to a top-level section (so it appears in the page TOC) and split it into three parts:

- **Checking the Roles of the Current User** — the existing `AuthenticationContext` content, unchanged.
- **Leaving Out, Hiding, and Disabling Components** — new. Ranks the options: don't create the component at all when possible; otherwise `setVisible(false)` / `setEnabled(false)`. Explains why these are the right tool — both are enforced server-side, so Flow ignores events and property changes from an invisible or disabled component even if someone re-enables the element in devtools. Links to *Component Visibility* and *Component Enabled State* in the Flow reference manual, and notes the `DisabledUpdateMode.ALWAYS` caveat. A `WARNING` names the wrong approaches explicitly: `display: none`, `visibility: hidden`, `pointer-events: none`, and setting the `hidden`/`disabled` attribute directly on the element.
- **Hiding Isn't a Security Boundary** — new. UI-level restriction governs what users can reach, not what they can do to the data; points at *Protect Services* for the real boundary.

**`articles/building-apps/security/protect-services.adoc`** — the intro already used "forget to disable a button" as its motivating example, so it now links to the new section.

**`articles/building-apps/react/security.adoc`** — the conditional-rendering example was correct but unexplained, leaving the same mistake available in a different form. Added a paragraph: leave restricted parts out of the returned markup rather than rendering them and hiding them with CSS, since the content stays in the DOM.

## Verification

`vale` reports 0 errors on all three files. Every remaining warning and suggestion is in pre-existing content.


articles/building-apps/security/protect-views.adoc
```diff
@@ -218,6 +223,53 @@ public class MyView extends Main {
...
+
+When the component has to exist -- because other code refers to it, or because you change its state while the view is open -- hide it with [methodname]`setVisible(false)`, or disable it with [methodname]`setEnabled(false)`:
+
+[source,java]
+----
+var deleteButton = new Button("Delete", event -> deleteSelected(grid));
+deleteButton.setVisible(authenticationContext.hasRole("ADMIN")); // <1>
+
+var salaryField = new TextField("Salary");
+salaryField.setEnabled(authenticationContext.hasRole("ADMIN")); // <2>
```
peholmst: setReadOnly would be better in this case (which means it is read only UNLESS you are an admin). We may need a third example for setEnabled.

peholmst: Agreed — changed to `setReadOnly(!hasRole("ADMIN"))`, and split the guidance into one example per method:

- `setVisible(false)` — the Delete button, for a control that shouldn't appear at all.
- `setEnabled(false)` — a new Approve button example, for an action users should see but can't perform.
- `setReadOnly(true)` — the salary field, where the value stays legible and selectable rather than dimmed.

Also mentioned `Binder.setReadOnly()` and `bindReadOnly()` for whole forms.

One thing I checked while writing it, since the section claims server-side enforcement: read-only holds up the same way. `AbstractFieldSupport.setValue()` rejects a value when `fromClient && component.isReadOnly()` and reapplies the buffered one. But it's narrower than the other two — it guards the field's value only, not everything the component can send — so the text now says that explicitly rather than lumping all three together.

That also affected the CSS warning: `isReadOnly()` reads the `readonly` *property*, not the attribute, so setting the attribute directly leaves the server writable. Added `readonly` to the list of attributes not to set by hand.

PR #6156, opened 2026-09-23
docs: SwingBridge installation fixes (java.version, localhost debug port, launch timeout)

eftunv:
## Description

Three fixes to the SwingBridge pages, found while checking the docs against the skeleton starter and the swing-bridge repository:

- **`installation-from-scratch.adoc`** — the from-scratch `pom.xml` sets `<java.version>21</java.version>` instead of `<maven.compiler.source>21</maven.compiler.source>`. The Spring Boot parent derives `maven.compiler.release` from `java.version`, so the old property was inert and a project built from the page compiled to Java 17 bytecode. A note under the snippet explains the trap. The starter made the same change in vaadin/skeleton-starter-vaadin-swing-bridge#25.
- **`installation-from-scratch.adoc`, `running-and-debugging.adoc`** — the JDWP debug agent is bound to `localhost:5005` instead of `*:5005`, with a sentence on what each accepts. Matches the starter's `pom.xml`.
- **`troubleshooting.adoc`, `configuration.adoc`** — the launch timeout now has a value: 5 seconds after `main` returns, not configurable (`SwingBridgeRunner.WINDOW_LOOKUP_TIMEOUT_MS`). The troubleshooting row also drops "a long-running blocking operation in `main`" as a cause, because the wait only starts once `main` has returned.

All four docs branches carry identical SwingBridge content, so the PR is labelled for `v25.2`, `v25.1` and `v25.0`.


mvysny (approved):
LGTM

